PR #108, opened 2014-07-23
Tab index bug fix

jonathanrocher:
[See line 137 of the split_editor_area_pane and line 245 of the unit test file for the important changes.]

When a tab is created with the splitEditorAreaPane, the index variable in the `add_editor` method was wrong only for the first editor. That led to the tooltip for that editor being lost. This is a hack to fix that. The real fix is probably down inside qt or pyside...

Thanks to @corranwebster for his help tracking this one!

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -8,6 +8,9 @@ Fixes
 
 * Fixed wx 2.9 incompatibility bug in ProgressDialog (PR#106).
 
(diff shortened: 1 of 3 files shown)


dpinte:
@corranwebster @jonathanrocher that looks ready to go no?

jonathanrocher:
I think so.

corranwebster:
@dpinte Yes, this should be merged.

PR #109, opened 2014-08-08
Add keyboard shortcuts for toolbar actions

tonysyu:
[QMenu.addAction](http://qt-project.org/doc/qt-4.8/qmenu.html#addAction) has a key-sequence argument in it's signature, but [QToolBar.addAction](http://qt-project.org/doc/qt-4.8/qtoolbar.html#addAction) doesn't, so the initialization of keyboard shortcuts is a bit different.


dpinte:
I suggest not to push this in 4.5.0 as the idea of 4.5.0 is to get a stable released version for the apps like Canopy that are currently working out of master.

corranwebster:
I find that I often end up with two versions of an `Action` anyway - typically you want an icon for toolbar `Actions`, but don't for menu actions.

PR #114, opened 2014-12-18
Generate a development version at build time if not building a release

jwiggins:
I'm going to keep shaving these yaks... This should be identical to enthought/traits#214


dpinte:
LGTM - merging

PR #118, opened 2015-02-03
Test Assistants

corranwebster:
This PR open-sources some useful test assistants that have been built over the past couple of years for Enthought's products and consulting projects.


corranwebster:
@rkern @jonathanrocher As discussed a couple of weeks ago.  If this could be reviewed reasonably quickly it would help some on-going work for consulting projects.

itziakos:
:+1:

jonathanrocher:
Looks good. This is valuable and for now hard to find. What about moving this into its own `testing_utils` folder? We should add an entry in CHANGES.txt, and another one in the documentation in my opinion.

pyface/ui/qt4/util/modal_dialog_tester.py
```diff
@@ -0,0 +1,296 @@
...
+"""
+import contextlib
+import sys
+import traceback
+
+from pyface.api import GUI
+from pyface.qt import QtCore, QtGui
+from traits.api import Undefined
+
+from canopy_data.util.test_assistants.event_loop_helper import EventLoopHelper
```
jonathanrocher: I don't think you are meaning to make these 2 imports!

pyface/ui/qt4/util/gui_test_assistant.py
```diff
@@ -0,0 +1,242 @@
...
+
+import mock
+
+from pyface.qt.QtGui import QApplication
+from pyface.ui.qt4.gui import GUI
+from traits.testing.unittest_tools import UnittestTools
+from traits.testing.unittest_tools import _TraitsChangeCollector as \
+    TraitsChangeCollector
+
+from .testing import find_qt_widget, print_qt_widget_tree
```
jonathanrocher: `print_qt_widget_tree` not used.

jonathanrocher:
I took the liberty to fix some obvious typos in the imports. We really ought to have pyface on travis CI...

pyface/ui/qt4/util/gui_test_assistant.py
corranwebster: It's there because the internal codebase that this code was pulled from had this function available in this module, and it will make it easier to translate code if we keep it here, even though it is not used.

dpinte:
LGTM, merging

PR #131, opened 2015-04-10
Add MANIFEST.in

corranwebster:
Pyface is missing a whole lot of stuff when you build an sdist.


itziakos:
should this PR be against a maintenance branch?

corranwebster:
I'm prepping another, equivalent one for the maintenance branch, but was interrupted by lunch.

itziakos:
@corranwebster  see similar comments in #132

PR #142, opened 2015-07-22
Add overview and improve API docs

corranwebster:
Some work in providing at least a high-level overview in the docs, plus improve the auto-documented API.  In particular, this now autodocs the major interface and mixin classes, and the classes themselves have been updated to use `#:` comments that can be picked up by Sphinx.  The latter means a lot of files have been touched in a trivial way.


corranwebster:
This PR now depends on #145 - but hopefully we now have enough test coverage to justify the code cleanup that is part of the PR.

corranwebster:
I think that this is ready for review/merge

PR #143, opened 2015-07-23
Fix Qt4 python editor imports and toolkit import mechanism

corranwebster:
There was an import error in the Qt implementation of the `PythonEditor` that was being silently swallowed, which meant that the `PythonEditor` widget was unavailable in Qt4.  This PR:

* fixes that particular issue
* modifies the toolkit import mechanism so that it only swallows import errors for the module that we are trying to import
* adds pygments to the travis.yml as import errors around pygments were also being swallowed silently

The latter means that import errors like the original one will make `import pyface.api` fail very quickly and should be noticed in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ before_install:
   - sudo apt-get update -qq
   - sudo apt-get install python-qt4 python-wxgtk2.8
 install:
+  - pip install pygments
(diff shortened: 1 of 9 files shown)


pankajp:
:+1: This has bitten me more than i care to admit

PR #145, opened 2015-07-25
Basic Interface Tests

### LINKED ISSUE #133: Error: global name '_RESULT_MAP' is not defined
There is an undefined dict at this line:
https://github.com/enthought/pyface/blob/master/pyface/ui/qt4/confirmation_dialog.py#L125

That can lead to errors of the form:
```
File "/Users/jrocher/Library/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/pyface/ui/qt4/confirmation_dialog.py", line 125, in _show_modal
    retval = _RESULT_MAP[retval]
NameError: global name '_RESULT_MAP' is not defined
```

corranwebster:
This aims to build out a basic suite of tests for all the formal interface implementations.  At this point this will be simple life-cycle, smoke-tests and, where possible, tests of the exposed methods of the Interfaces.

Includes some fixes for bugs found as part of the testing process:

* fixed missing import in qt4 `confirmation_dialog` (fixes #133)
* fixed failure to release application window from `AUIManager` in Wx.
* `ActionItem`'s change handler for the `visible` trait was always setting to `True`
* `Group` objects incorrectly set the `text` trait instead of the `name` trait of an `Action` created from a callable.
* python shell on Qt wouldn't execute files correctly on Windows because of how files were being passed to interpreter.


corranwebster:
Note: seem to be getting random segfaults in the qt tests

pyface/ui/qt4/util/modal_dialog_tester.py
```diff
@@ -247,6 +247,22 @@ def click_widget(self, text, type_=QtGui.QPushButton):
         )
         widget.click()
 
+    def click_ok(self):
```
itziakos: How about using a dictionary 
```python
BUTTONS = {'ok': 'OK', 'OK', 'yes': '&Yes'}
```
it would be easier to extend later to container buttons like 'Next', 'Continue', 'Stop'

It would be easier to extent later

itziakos: update my comment with an alternative.

corranwebster: Yes, using pyface constants as arguments would be better.

corranwebster:
I consider this done for now.  There is obviously a lot more that could be done, but this gives fairly good coverage of basic widgets and actions.

@itziakos could you please do a quick sanity check and also make sure that everything runs OK on Windows.

itziakos:
on windows that tests hold for ever

itziakos:
is it ok if I add support for appveyor so that we can collaborate on fixing the problems?

corranwebster:
By all means add appveyor support, but possibly on a separate PR off of master so we can separate getting appveyor working from getting these tests working on windows.

corranwebster:
Have we lost coverage in CI with the appveyor changes?

PR #149, opened 2015-08-03
Deprecate wx toolkit default

corranwebster:
Unfortunately, can't easily test this because `_init_toolkit()` pulls the ladder up after itself, so we can't reset the toolkit to test the code.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -14,6 +14,9 @@ The following GUI backends are supported:
 - PyQt
 - PySide
 
+**Warning:** In Pyface version 5.0 the default GUI backend will change from
+``wx`` to ``qt4``.
+
 Documentation
 -------------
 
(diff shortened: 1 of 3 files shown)


corranwebster:
@itziakos care to do a quick check of this code.

I notice that appveyor seems to be paying attention to this code, but the failure is spurious.

itziakos:
> I notice that appveyor seems to be paying attention to this code, but the failure is spurious.

It was a problem with the appveyor settings, I have updated the configuration to not try to built branches that do not have an appveyor.yml

itziakos:
a tests similar to what was done in https://github.com/enthought/traitsui/pull/229 should be added.

corranwebster:
@itziakos You can't test this in the same way with the code as currently written: once the `toolkit` module is imported for the first time it self-deletes the `_init_toolkit` method we want to test.

You could probably do it with hacky `reload`ing of the module, but I don't want to go there.

itziakos:
> You could probably do it with hacky reloading of the module, but I don't want to go there.

fair enough

jwiggins:
I'm looking forward to the equivalent of nucleic/enaml#165 someday :smile:

PR #150, opened 2015-08-04
Deprecate wx toolkit default maintenance

corranwebster:
4.5.x maintenance branch version of #149.


corranwebster:
Merging even though we don't satisfy code coverage for the change.  See #149 for reasons.

PR #155, opened 2015-08-11
Fix task window layout running after control destroyed

corranwebster:
The layout callback run by the following code https://github.com/enthought/pyface/blob/master/pyface/ui/qt4/tasks/main_window_layout.py#L129 can run after the tasks control has been destroyed.

This PR resolves the issue by flagging that the control is about to be destroyed from the TaskWindow backend implementation by setting the control to `None`.  The layout can then check this and not do any further work.


rkern:
LGTM.

PR #159, opened 2015-08-14
Don't force a conversion to str when reading text from Qt clipboard

### LINKED ISSUE #146: clipboard should support non ascii text
```
/Users/.../Library/Enthought/Canopy_64bit/User/lib/python2.7/site-packages/pyface/ui/qt4/clipboard.pyc in _get_text_data(self)
     80 
     81     def _get_text_data(self):
---> 82         return str(cb.text())
     83 
     84     def _set_text_data(self, data):

UnicodeEncodeError: 'ascii' codec can't encode character u'\xf4' in position 2045: ordinal not in range(128) 
```

The conversion to `str` will fail miserably with non-asci characters: https://github.com/enthought/pyface/blob/master/pyface/ui/qt4/clipboard.py#L82

brett-patterson:
The conversion was failing when the string had non-ASCII characters. Fixes #146

### DIFF
--- a/pyface/ui/qt4/clipboard.py
+++ b/pyface/ui/qt4/clipboard.py
@@ -79,7 +79,7 @@ class Clipboard(BaseClipboard):
     #---------------------------------------------------------------------------
 
     def _get_text_data(self):
-        return str(cb.text())
+        return cb.text()
 
     def _set_text_data(self, data):
         cb.setText(data)


corranwebster:
:+1

dpinte:
@brett-patterson maybe adding a line to the CHANGES.txt file would be useful.

corranwebster:
As it turns out, this has been folded into #160 (it was needed to get Python 3 working correctly).

brett-patterson:
@corranwebster Great! Closing.

PR #105, opened 2014-05-22
Python 3 compatibility and Travis integration

pradyunsg:
This PR adds:
  - Python 3 compatibility sourced from burnpanck's fork.
  - Travis integration for Python 2.7 and 3.3.

PR #160, opened 2015-08-17
Python 3 support

corranwebster:
Updating #105 - see that PR for history & discussion.


corranwebster:
I think appveyor is pulling the wrong traitsui branch (actually, maybe everything is, but we aren't running tests under Python 3 on Travis CI...)

corranwebster:
Assuming appveyor is happy, tests are now passing in all expected cases.  For some reason we are not excluding the Wx & Python 3 combination from the Travis CI test matrix, even though it appears to be in the .travis.yml file to allow it.

Anyway, at this point it's all about the CI infrastructure.

corranwebster:
I think this is good to go.

Merging!

PR #161, opened 2015-08-18
Use TraitsUI master for Travis CI

corranwebster:
Use traitsui master instead of Python 3 branch.


corranwebster:
And we're good.

Merging.

PR #162, opened 2015-08-20
Some miscellaneous Python3 fixes.

prabhuramachandran:
The changes should be compatible with Python 2.

### DIFF
--- a/pyface/ui/qt4/workbench/split_tab_widget.py
+++ b/pyface/ui/qt4/workbench/split_tab_widget.py
@@ -10,6 +10,9 @@
 # Standard library imports.
 import sys
 
+if sys.version_info[0] > 2:
+    unicode = str
+
 # Major library imports.
 from pyface.qt import QtCore, QtGui, qt_api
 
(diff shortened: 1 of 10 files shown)


corranwebster:
Thanks for these: the code should be being run through 2to3 on install if Python 3 is being used, which should automatically fix all of these, but it would be really nice one day to not have to rely on 2to3.

prabhuramachandran:
Thanks, the collections change is not seen by 2to3.  There might be a few more small changes needed and I will push those too.

pyface/ui/qt4/workbench/split_tab_widget.py
```diff
@@ -10,6 +10,9 @@
 # Standard library imports.
 import sys
 
+if sys.version_info[0] > 2:
+    unicode = str
```
corranwebster: `str` here is being used to coerce the output of `QSplitter.saveState()` which is a `QByteArray`.  From fixing up the TraitsUI code, I found that using `QByteArray`'s `data()` method worked well across versions to give an appropriate Python object (`str` in 2.x, `bytes` in 3.x).

corranwebster: Offline discussion determined that I was misunderstanding here.

corranwebster:
Going to merge this now - we can always open another one later.

PR #163, opened 2015-08-20
Ensure examples work in Python3

corranwebster:
Some examples are Wx specific, so we won't port those, but we would like to get everything else working if we can.

Builds on #162.


corranwebster:
This is ready to merge.  Some examples require traitsui fixes if numpy isn't installed.

PR #164, opened 2015-08-21
Fix regression on traitsui branch used in CI

corranwebster:
I think this happened in the move from `dev_requirements.txt` to `travis-ci-requirements.txt`.


corranwebster:
Looks good, but @itziakos Can you double-check this?

itziakos:
:+1:

PR #115, opened 2015-01-15
Fixes and updates for wxPython 3.0.x compatibility

RobinD42:
This is the results of my first look at compatibility testing, there are likely still more changes that need to be made. I just wanted to get this PR created for now in case others would like to contribute to it.

WIP, don't merge yet. More to come as free time allows...

PR #168, opened 2015-09-24
Fixes for wxPython 3.0

corranwebster:
Merged version of #115 

Will see if I can get tests turned on for wxPython 3.0 as well before merging.


corranwebster:
I am willing to merge this with < 80% of the delta tested, since testing the change is presently not feasible, and is in otherwise untested code.

It would also be nice to add CI tests using wxWindows 3, but that has not been whitelisted for containerized builds, so it will have to wait.

PR #171, opened 2015-10-12
Qt5 Support (V2) - DO NOT MERGE!

corranwebster:
Keeping this around for potential future move from Qt4 API to Qt5 API, but needs work based on what we ended up doing for #264.

See https://github.com/enthought/traitsui/pull/242

These will probably need to be worked on in tandem to get the tests working.


corranwebster:
@geggo Looks like there is an issue that `QItemSelection` has moved from `QtGui` to `QtCore` in the 4 to 5 shift.  This is causing test failures in TraitsUI.

I _think_ the fix is conditionally import it into `pyface.qt.QtCore` when using Qt4.

corranwebster:
Additional to-do: check changed `QFileDialog` methods.

pankajp:
I have a different idea about Qt5 support in pyface. With this PR, a lot of code woul need to be changed significantly to support Qt5. I propose that we implement a new `pyface.qt5` subpackage which would mimic the Qt5 api, and keep the existing `pyface.qt` provide the Qt4 api (including whatever Qt4 api needs to be compatible ).

When using Qt5, `pyface.qt` would pull in all relevant imports needed to mimic the Qt4 api (ex import all things from `QtGui` and `QtWidgets` into `pyface.qt.QtGui` module.

When using Qt4, `pyface.qt5` would pull in all relevant imports into corresponding Qt5 api (ex import everything from `QtGui` into both `pyface.qt5.QtGui` and `pyface.qt5.QtWidgets` modules).

That way, applications can decide to incrementally move from qt4 to qt5 api, and components using the different parts could still run together

corranwebster:
@pankajp that suggestion has some merit.  The advantages I see are:
* you can have a "clean" qt5 api for the future,
* you are signalling the intent of your code
The downsides:
* code written to qt4 API will persist, and be copied without thinking,
* trying to make current code "just work" with the qt5 API without modification may lead to subtle bugs
* more effort in upgrading code to the qt5 API properly (you need to touch _every_ `pyside.qt` import)

If we do this, I can see the point of having a mirror `pyside.qt4` API.

However, my immediate concern is just getting the CI tests to pass reliably with a decent set of coverage.

corranwebster:
Note, although these are passing, I'm not sure I trust that the results of the Python 3.5 + pyqt5 test right now.

corranwebster:
This has been superseded by #264.

Keeping this around as we would like to move from Qt4 API to Qt5 API in the internal code, and this has the basics of making that move.

corranwebster:
Closing to avoid confusion - Qt5 support is in master.

PR #176, opened 2015-11-04
A small fix to ProgressDialog and a change to code_widget in order to enabe (auto_set=False) for traitsui Qt CodeEditor

bergtholdt:
The first commit fixes QT and WX ProgressDialog to enable updates to the message which did not work.

The second commit adds a "focus_lost" signal in the QT CodeWidget that is emitted whenever focus goes out of the widget, thus enabling traitsui CodeEditor to behave properly for auto_set=False (see other pull request on traitsui)


corranwebster:
Thanks for this.  It looks good, so I'll merge and write tests for it in a separate PR.

PR #181, opened 2015-11-14
MAINT: Add 'detail' and 'informative' to dialog helpers.

jvkersch:
As the title says.


corranwebster:
Generally LGTM.

It probably makes sense to add these to the methods on the `IWindow` class as well.  Oh and tests would be nice (we test the class but not the helper functions).  But I understand if you don't want to do that :)

jvkersch:
@corranwebster Thanks for the review. I added a test for the helpers, can you check if this is what you had in mind?

corranwebster:
Yep, test cases look good, not sure why tests are failing.

corranwebster:
@jvkersch Looked into this a bit more.  Test failures seem to be that you are returning the dialog object where previously the methods returned `None`.  For message dialogs there doesn't appear to be much point in returning the dialog after the user has finished with it - there is no state to interrogate because the user isn't given any choice in responses.  Was there something you had in mind with this change?

jvkersch:
> Was there something you had in mind with this change?

Even if there was, I can't remember now :(

corranwebster:
I think it was so that you could introspect the dialog in the tests, I suspect.  I think I'm OK with basic smoke tests of the methods for now.

jvkersch:
Thanks Corran!

PR #189, opened 2016-01-12
Clarified the purpose of LayoutContainer construction args.

mdwhite:
LayoutContainer's __init__ accepts arguments *items and **traits. At first glance, the argument names make it seem as if one could initialize LayoutContainer.items by calling LayoutContainer(items=some_items). However, doing this would create a task layout with items=() and traits={'items': some_items}. 

This behavior is misleading. I ran into some code yesterday where a vsplitter had been constructed with VSplitter(items=list_of_items), and it took me a few looks to realize that {items: list_of_items} was being passed to **traits instead of items.

The current commit will not fix the above example, but I'm hoping it will make it clear that the above usage is incorrect. All of my ideas for making LayoutContainer(items=[...]) assign a value to items would break backwards compatibility. @dpinte any ideas?

### DIFF
--- a/pyface/tasks/task_layout.py
+++ b/pyface/tasks/task_layout.py
@@ -73,9 +73,9 @@ class LayoutContainer(LayoutItem):
 
     items = List(pretty_skip=True)
 
-    def __init__(self, *items, **traits):
-        super(LayoutContainer, self).__init__(**traits)
-        self.items = list(items)
+    def __init__(self, *args, **kwargs):
+        super(LayoutContainer, self).__init__(**kwargs)
+        self.items = list(args)
 
     def iterleaves(self):
         for item in self.items:
--- a/pyface/tasks/task_layout.py
+++ b/pyface/tasks/task_layout.py
@@ -73,9 +73,16 @@ class LayoutContainer(LayoutItem):
 
     items = List(pretty_skip=True)
 
-    def __init__(self, *args, **kwargs):
-        super(LayoutContainer, self).__init__(**kwargs)
-        self.items = list(args)
+    def __init__(self, *items, **traits):
+        # Items may either be specified as a positional arg or a kwarg.
+        if items:
+            if 'items' in traits:
+                raise ValueError(
+                    "Received 'items' as positional and keyword argument."
+                )
+            else:
+                traits['items'] = list(items)
+        super(LayoutContainer, self).__init__(**traits)
 
     def iterleaves(self):
         for item in self.items:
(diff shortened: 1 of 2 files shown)


pyface/tasks/task_layout.py
```diff
@@ -73,9 +73,9 @@ class LayoutContainer(LayoutItem):
 
     items = List(pretty_skip=True)
 
-    def __init__(self, *items, **traits):
-        super(LayoutContainer, self).__init__(**traits)
-        self.items = list(items)
+    def __init__(self, *args, **kwargs):
```
corranwebster: Leave `**kwargs` as `**traits` as this is common usage for keyword args of traits `__init__` methods.

corranwebster:
Personally, I prefer the more descriptive name for the positional arguments.  To handle the error you encountered better, it might make sense to instead do:

    if not self.items:
        self.items = list(items)

or, alternatively, doing something like:

    if 'items' not in traits:
        traits['items'] = items
    super(...)

dpinte:
@mdwhite Can you address the comments?

mdwhite:
@dpinte @corranwebster very sorry for dropping the ball here. @corranwebster, thanks for looking into this! I'll get back to you later today.

mdwhite:
@corranwebster good points. there may not be a perfect solution. I'm not sure if this is a case we should consider, but both of the options might not behave as expected if the user passes in items as a positional arg _and_ a kwarg. With either change, the constructor would ignore the positional arg. 

If we want to consider both, we could replace `self.items = items` with `self.items.extend(*items)`. This seems like a safer change to me, but I'm happy to defer to your judgment here.

corranwebster:
I think that the case of passing both a list of positional arguments as items and an items keyword is confusing (ie. `LayoutContainer(foo, bar, items=[baz, wom])` and shouldn't be supported.  How does the following proposal sound:

* if both positional and keyword values are passed for `items` we should throw a `TypeError: LayoutContainer() got multiple 'items' arguments` (or something similar).
* if either is passed, but not both, we should pass it into the superclass constructor
* if none are passed, then we need to ensure it is initialized to an empty list somehow (which should be the default if it's not passed, but it may be cleaner to pass an empty list, depending on how the other cases are coded)

If you can get this cleaned up today, @kitchoi can probably get it into the upcoming release.

mdwhite:
@corranwebster I addressed your comments, but now the build is failing. Any ideas why? I can't reproduce the test failures locally

corranwebster:
Looks like its an issue with the Tasks Enaml components.  Its possible that this change tweaked some issue, or there could just be something that's changed with Enaml.

corranwebster:
I'm guessing that the issue may have been this: https://github.com/enthought/pyface/pull/190.  Certainly merging master in seems to have fixed the failure.

corranwebster:
Test pass, LGTM.  Merging.

PR #192, opened 2016-02-18
Qt4 Workbench should create UI with parent being the dock widget

kitchoi:
The dock widget's ViewContainer instead of the main window should be passed for creating UI.


corranwebster:
LGTM - the test failures appear to be orthogonal to all of this.

kitchoi:
Thanks.  #194 should fix those bugs.  Can I wait until #194 is merged and then run the test suites in order to make sure?

corranwebster:
Yep - you can wait until #194 is merged.

But since I just merged it, you don't have to :)

PR #194, opened 2016-02-19
fix: correct the extened trait change description

itziakos:
On of the change handlers was using the `on_trait_change` decorator providing three separate trait names. However, this signature is not supported since (https://github.com/enthought/traits/pull/207) and actually never properly worked. Older versions of traits would ignore the second trait (not covered in the tests) and the recent version raises an error.

This PR fixes the issue for both old and new versions of traits


corranwebster:
LGTM - merging

PR #198, opened 2016-03-09
Feature: add simple TaskApplication class (simpler than envisage's version)

jonathanrocher:
This PR introduces a simple Task based `Application` object to support creating and running a simple GUI application containing Tasks and panes as defined in [pyface](http://docs.enthought.com/envisage/tasks_user_manual/index.html). Users can simply subclass it and provide an implementation of the `start` method that creates tasks and windows for them. This class provide some of the features of the Envisage [version](https://github.com/enthought/envisage/blob/master/envisage/application.py) of the `Application` object and aims at being used for light weight application that don't need the plugin infrastructure that Envisage offers.

Its introduction allows to simplify the `run.py` script of the task example so that users don't have to mess around with the GUI event loop and can just create tasks and windows. I added a run_fancy_app.py that shows how to use more of the bells and wistles of the TaskApplication class. 

Code vastly written by @corranwebster .


examples/tasks/advanced/images/image_LICENSE.txt
```diff
@@ -3,7 +3,7 @@ licensed accordingly to the original license:
 
 Project                 License                   File
 ----------------------------------------------------------------------------
-Enthought               BSD 3-Clause              LICENSE.txt
+Enthought               Proprietary
```
pankajp: I think we should not put proprietary files in open source projects. Not sure though, @rkern any feedback ? The Enthought name and the logos are trademarked so people cannot use them indiscriminately. What kind of license do these files need? Btw, we should also mention somewhere that the Oxygen icons are LGPL licensed

corranwebster:
I think that there shouldn't be dependencies on Apptools in this basic class since that introduces circular project dependencies.  In other words, for an absolutely basic `TasksApplication` we shouldn't have the preferences and the undo manager.

examples/tasks/advanced/images/image_LICENSE.txt
corranwebster: I concur.  We should only include files which have reasonably free licenses, so we'd either have to open the license file (which is problematic from the point of view of Enthought-as-a-business) or find something else to put there.

jonathanrocher:
Thanks for the feedback. I will remove the apptools stuff from the class, and will probably keep the preference and undo manager out of the example for this PR.

Will follow up with legal about what the license should be for the images...

examples/tasks/advanced/images/image_LICENSE.txt
jonathanrocher: I spoke with @eric-jones and @cgodshall and we don't need any license for our logo files. I have removed the entry altogether.

corranwebster:
Licensing information really isn't fixed by just removing the Enthought line from the image license file - there may be non-Oxygen files that we _are_ releasing as open source; and by leaving the licensing of the enthought logo files as ambiguous it makes the licensing of the whole project ambiguous.

Either we should use some other files for the images (maybe something from the python logos?); or we should release the logos with some open-source but non-commercial licencing (eg. CreativeCommons no derivatives, non-commercial, share-alike or something like that, but that would require Bill's OK).

rkern:
IMO, there is no particularly strong reason to have the Enthought logos in the demo app.

examples/tasks/advanced/run_fancy_app.py
```diff
@@ -0,0 +1,132 @@
...
+        logger.debug("Log file is at '{}'".format(filepath))
+
+    def _on_window_closing(self, window, trait, old, new):
+        """ Ask confirmation when a window is closed. """
+        from pyface.api import confirm, YES
+
+        msg = "Are you sure you want to close the window?"
+        return_code = confirm(None, msg)
+        if return_code != YES:
+            logger.debug("Window closing even was veto-ed")
```
fbender: Nit: "even**t**"

jonathanrocher:
I want to push this base class a little more, and add a couple more bells and whistles to it. I am thinking of:
1. an `AboutDialog` attribute
2. add a confirmation dialog when catching a window closing request event.

corranwebster:
@jonathanrocher How do you want to proceed with this?

I'd be fairly happy to merge as-is (assuming tests are all passing), but if there is something in particular that you think is missing, now would be a good time to discuss.

jonathanrocher:
Thanks for all the work you did in your PR to this @corranwebster . Let me review today what this now looks like and comment if I see anything.

examples/tasks/advanced/run.py
```diff
@@ -1,35 +1,49 @@
...
-    task = ExampleTask()
-    window = TaskWindow(size=(800, 600))
-    window.add_task(task)
+    # hook up listener to application initialized event
+    def app_started(event):
+        create_task_window(event.application, files=argv[1:])
 
-    # Show the window.
-    window.open()
+    app.on_trait_change(app_started, "application_initialized")
```
jonathanrocher: The only thing I am still not happy with in this PR is the complexity of this main, with 2 helper functions and the need for this on_trait_change @corranwebster . In my opinion, this API isn't reasonable because it requires to know so many things:
1. you have to listen to application initialized,
2. you have to know that the event has an attribute that gives you back the application object, 
3. dynamic listeners like this isn't the most commonly known feature of traits (for example not something we/Enthought teach in classes)

The way I implemented the window creation in my version of `TaskApplication`, was by overriding the start method, call super and launch window creation afterwards. It is working well enough, though looking at the chain of events precisely, this approach listening to `application_initialized` may be cleaner. If that's the case, I would much rather add complexity in `TaskApplication` than in every subclass. So I would propose to move the listener to `TaskApplication`, calling back a method that does nothing, and that is supposed to be overridden.

The resulting user code could be something like:
```
class MyTaskApplication(TaskApplication):
    def on_application_initialized(self):
        task = ExampleTask()
        self.create_task_window(task)
        for filename in self.initial_files:
            task._open_file(filename)

def main():
    initial_files = sys.argv[1:]
    app = MyTaskApplication(initial_files=initial_files)
    app.run()
```
Do you like that?

examples/tasks/advanced/run_fancy_app.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def _icon_default(self):
+        return ImageResource("python_icon.png")
+
+    def _splash_screen_default(self):
+        img = ImageResource("python_logo.png")
+        return SplashScreen(image=img)
+
+
+def main(argv):
```
jonathanrocher: `argv` not used.

pyface/base_application.py
```diff
@@ -0,0 +1,395 @@
...
+
+    def setup_logging(self):
+        """ Initialize logger.
+
+        Default behaviour is to log to a rotating file handler whose name is
+        the id of the app.
+        """
+        root_logger = logging.getLogger()
+
+        if DEBUG:
```
jonathanrocher: What about making this an instance attribute (or an argument to the method) instead of a module level variable? Would make it easier to access for end users.

pyface/base_application.py
```diff
@@ -0,0 +1,395 @@
...
+    def _run(self):
+        """ Actual implementation of running the application
+
+        This should be completely overriden by applications which want to
+        actually do something.  Usually this method starts an event loop and
+        blocks, but for command-line applications this could be where the
+        main application logic is called from.
+        """
+        # Fire a notification that the app is running.  If the app has an
+        # event loop (eg. a GUI, Tornado web app, etc.) then the
```
jonathanrocher: nit: then the -> then this

pyface/base_application.py
```diff
@@ -0,0 +1,395 @@
...
+                logger.exception("Error preparing for application exit")
+            finally:
+                logger.info('---- Application exit ----')
+                self._exit()
+        else:
+            logger.info('---- Application exit vetoed ----')
+
+    # Initialization utilities -----------------------------------------------
+
+    def setup_logging(self):
```
jonathanrocher: On a second look, it feels to me like we are trying to do too many things here. It doesn't feel like this class' responsibility to create all this logging infrastructure. I would leave this empty and let users subclass and implement if needed.

pyface/base_application.py
```diff
@@ -0,0 +1,395 @@
...
+
+    def run(self):
+        """ Run the application
+
+        Return
+        ------
+        status : bool
+            Whether or not the application ran normally
+        """
+        started = run = stopped = False
```
jonathanrocher: `started` is never used since reset 2 lines down.

pyface/base_application.py
jonathanrocher: Additionally, I believe that this method is never called. So I would make it empty in this base class, but I would call it at the beginning of the `start` method, so that if someone subclasses it, it is invoked automatically.

examples/tasks/advanced/run_fancy_app.py
```diff
@@ -0,0 +1,109 @@
...
+        -------
+        window : TaskWindow
+            Window that was created, containing the newly created task.
+        """
+        task = ExampleTask()
+        task.extra_actions = self.extra_actions
+        window = self.create_task_window(task)
+        return window
+
+    def _setup_logging(self):
```
jonathanrocher: Typo in method name. It should be `setup_logging` if we want to match the parent class.

pyface/application.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+class Application(BaseApplication):
+    """ A basic Pyface application.
+
+    This handles setting up starting up the GUI, window management, logging,
+    and other common features that we want when creating a GUI application.
+
+    Often you will want to subclass this application, but it can be used as-is
+    by hooking a listener to the `initialized` event::
```
jonathanrocher: Typo: `application_initialized` is the correct event name.

pyface/application.py
```diff
@@ -0,0 +1,241 @@
...
+            def _create_contents(self, parent):
+                self._label = HeadingText(parent, text="Hello World")
+                return self._label.control
+
+        def main():
+            application = Application(id='hello-world', name='Hello World')
+            window = MainWindow()
+            application.on_trait_change(
+                lambda: application.add_window(window),
+                'initialized'
```
jonathanrocher: Typo: `application_initialized` is the correct event name.

pyface/base_application.py
```diff
@@ -0,0 +1,395 @@
...
+        This permits applications to do things like catch and display an error
+        dialog when something unexpectedly goes wrong.
+        """
+        self._old_excepthook = sys.excepthook
+        sys.excepthook = self._excepthook
+        logger.debug('Exception hook installed')
+
+    # Teardown utilities -----------------------------------------------
+
+    def reset_logging(self):
```
jonathanrocher: Same comment as above about `setup_logging`: I don't think it is this class' responsibility to manager the root logger, and it should be left to user code if they want their application object to implement this.

jonathanrocher:
One final suggestion: now that this PR introduces a deep `Application` class hierarchy, I would suggest the following class renaming:
`BaseApplication` -> `Application` (to signify that that class can be used by end users, and is not just a base class)
`Application` -> `GuiApplication` (to signify that it contains GUI components)
`TaskApplication` -> `TaskGuiApplication` (to signify it is a subclass of `GuiApplication`)

examples/tasks/advanced/run.py
corranwebster: The reason for wanting to emphasize the event listener code is because this is the preferred way to handle "do this when everything is set up and the event loop is running" from an Envisage plugin's `start` method to avoid the particular hell that is getting the plugin order exactly right.

Unfortunately in the past couple of weeks I've been playing with things, and although the callback method works very nicely on Qt, it fails abysmally under Wx because Wx _requires_ that there is at least one window before the `wx.App` starts, and will quit automatically when the last window closes.

This means that any Wx-based app _must_ create a window as part of its `start` method.

So this means something of a re-design which will probably mean adding a `_create_windows` method that is called before the mainloop starts and which:

* is intended to be overridden in the`Application` class
* has a default implementation for `TaskApplication` which creates a task window (or more enthusiastically, follows the logic for task layouts from the Envisage version).

We can revisit the discussion after I've put something together.

pyface/base_application.py
corranwebster: The intent here was to provide a sane default for setting up logging, but allow the user to call or not call (and override, as desired) in their start method (and similarly with the teardown utilities).  The same idea was there for the excepthook and application home code.

A large part of this was to simplify _testing_ since logging, exception hooks and home directories are annoyingly global and having them as side-effects of start-up and teardown is painful.

Perhaps the right thing to do is to pull them out and make them helper functions (or better, context managers?), so something like:
```
app = Application()
with setup_logging(app), install_excepthook(app):
        app.run()
```
or, I guess,
```
app = Application()
with app.setup_logging(), app.install_excepthook():
        app.run()
```

corranwebster:
I'd be OK with the renaming.  Originally `BaseApplication` was intended to be abstract, but since it can be used directly, it makes sense to call it `Application`.

pyface/tasks/task_application.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+
+    # -------------------------------------------------------------------------
+    # 'TaskGuiApplication' interface
+    # -------------------------------------------------------------------------
+
+    # Tasks management --------------------------------------------------------
+
+    #: List of all running tasks
+    tasks = List("pyface.tasks.task.Task")
```
jonathanrocher: New bug found: this will lead toy a list of unicode elements, instead of a list of Task instances! Missing `Instance`.

corranwebster: Yep, found that this morning...

examples/tasks/advanced/run.py
jonathanrocher: Thanks, lots of things I didn't know here. Not sure I understand your focus on what envisage expects though (first paragraph) since this Application object is only useful when people don't use Envisage. I have never seen any issues using this class and creating windows in the `start` method under `qt` (`pyside` to be precise). Are you expecting this to fail sometimes? Otherwise, that could be the way to do things which would work for both Qt and WX.

pyface/base_application.py
jonathanrocher: Removed.

examples/tasks/advanced/run_fancy_app.py
jonathanrocher: Fixed.

pyface/base_application.py
jonathanrocher: Moved to the example (though not keeping the handles on the handlers).

corranwebster:
Looks like there are a number of imports which need to be fixed.

corranwebster:
Closed in favour of #345

PR #199, opened 2016-03-22
Fix exception when closing all tabs

stefanoborini:
Fixes behavior experienced in enthought/mayavi#321.
Performs a check before trying to set text color.


pyface/ui/qt4/workbench/workbench_window_layout.py
```diff
@@ -336,7 +336,8 @@ def _qt4_editor_tab_spinner(self, editor, name, new):
     def _qt4_active_editor_changed(self, old, new):
         """ Handle change of active editor """
         # Reset tab title to foreground color
-        self._qt4_editor_area.setTabTextColor(new.control)
+        if new:
```
corranwebster: Slight preference for `if new is not None:` here.

corranwebster:
LGTM

PR #207, opened 2016-04-08
Fix restoring show_editor_area from saved perspective

kitchoi:
Fix https://github.com/enthought/mayavi/issues/337

Include the info whether the editor_area is visible in the `window_memento` file.  This code still respects old `window_memento` files and assume `show_editor_area = True` in that case.

The more difficult thing to write was the test case.  A lot of mock objects are required for testing this bug fix.

### DIFF
--- a/pyface/ui/qt4/workbench/workbench_window_layout.py
+++ b/pyface/ui/qt4/workbench/workbench_window_layout.py
@@ -304,6 +304,8 @@ class WorkbenchWindowLayout(MWorkbenchWindowLayout):
                     if geometry is not None:
                         self.window.control.restoreGeometry(geometry)
 
+    def is_editor_area_visible(self):
+        return self._qt4_editor_area.isVisible()
 
     ###########################################################################
     # Private interface.
--- a/pyface/ui/wx/workbench/workbench_window_layout.py
+++ b/pyface/ui/wx/workbench/workbench_window_layout.py
@@ -248,6 +248,12 @@ class WorkbenchWindowLayout(MWorkbenchWindowLayout):
 
         return
 
+    def is_editor_area_visible(self):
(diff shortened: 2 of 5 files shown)


kitchoi:
The test coverage drops upon adding a test case!  Previously the workbench module has been ignored by codecov and now it is included.

kitchoi:
I am aware of traits (dynamic?) listener errors because the WorkbenchLayout is mocked.  It is a bit of an eyesore and I would appreciate any suggestion on how to silent these errors.  https://travis-ci.org/enthought/pyface/jobs/121676885#L492
codecov/path diff hit is not high enough because of codes for handling old mementos and WorkbenchLayout is mocked.  For this I have tested on my local machine with wx and qt and with old memento generated by the master branch.  Testing it in the CI tool would require doing a full initialisation of the ui controls for the WorkbenchLayout.

corranwebster:
Perhaps as a simple work-around for the error, just put a sanity-check guard here https://github.com/enthought/pyface/blob/fix-workbench-save-editor-area-visibility/pyface/workbench/workbench_window.py#L896 so that if the value of `editor` is nonsensical (`None` and `Undefined` being the most likely cases) then the code just doesn't try to do anything?

corranwebster:
Also, as far as test coverage goes, as long as there are reasonable tests added it doesn't matter if you can't hit 80%

kitchoi:
Thanks @corranwebster !  I will try the sanity-check approach.

kitchoi:
Done.

kitchoi:
@corranwebster can we merge this for the release?

corranwebster:
Yes - it looks good to go.

kitchoi:
Thanks.  Merging.

PR #208, opened 2016-04-15
Specify Enaml version to use

corranwebster:
I think a recent change to the Enaml imports in Enaml's master may have broken the tests.  So try specifying a known-good version of Enaml.


corranwebster:
I think errors were form some other cause.  Closing for now.

PR #211, opened 2016-04-19
MAINT: modernize raise syntax

grlee77:


jakirkham:
While there is a failure on Travis CI. It appears that failure is a segmentation fault. It is extremely unlikely that a small change in Python syntax would cause this. So, would argue it is totally unrelated.

corranwebster:
LGTM.  Merging, although it may be too late to get into the upcoming release.

jakirkham:
That's fine. We can hang on to the patch. Just let us know when it does go into a release, so we can drop the patch.

PR #217, opened 2016-05-24
Qt single choice dialog

corranwebster:
Add a Qt version of the single-choice dialog window that already exists for Wx.

screen shot 2016-06-08 at 5 36 07 pm


corranwebster:
This has been sitting around un-reviewed for 6 months.  It's not controversial, so I'm just going to merge this tomorrow unless someone has any objections.

Issue #219, opened 2016-06-11
Clarify Tasks wx support


fbender:
Per https://github.com/enthought/pyface/blame/4e454d3da537be74acf6919ecaa87555e6947579/docs/source/overview.rst#L256, Tasks do not support the wx backend. There have been a couple of releases since July 2015. 

Could you please clarify if Tasks support the wx backend by now?

Thanks!

jdmarch:
No, Tasks does not support wx and there are no plans for it to do so.

> Tasks, which is currently available only for Qt, but is under active development.

To be explicit, this does *not* mean that wx support is under active development. It means that tasks is under active development.

fbender:
Thank you for the quick reply. That's very unfortunate. 
There's an existing app I'd like to convert to ETS (and it currently uses wx directly), but this can only happen piece-by-piece so I need to stick with wx for the time being. Workbench it is, then!

@jdmarch are you going to clarify the Docs or should I send in a PR?

jdmarch:
Please feel free to submit a PR. 

Regarding your plans, perhaps others with workbench experience can comment. @corranwebster ?

fbender:
See #220

> Regarding your plans, perhaps others with workbench experience can comment. @corranwebster ?

That would be awesome! I'm having trouble finding good documentation on how to use Workbench via Envisage (I'm new to the ETS stack but absolutely love the concept). I will take a look at the plugins examples in enthought/envisage but I'm not sure how far this will get me, especially considering that the (non-Tasks) examples haven't been updated in a long time … 
FYI: The application is part of the [NANOBED project @ University of Strathclyde](http://craftprospect.com/nanobed/) and features a tree view of the sat, mission & env model plus a few "tabs" containing simulation and analysis "editors" that read and in some cases may modify the tree. So if there's any Open Source application that is similar to this structure and is using the Envisage Workbench API, I'd be glad to learn from it!

corranwebster:
@fbender There was an active effort to port Tasks to Wx by @robmcmullen (see the master branch of his fork of pyface: https://github.com/robmcmullen/pyface).  I don't know the current state of his effort, other than it never made it to a pull-request.  However it might provide a kicking-off point for you, particularly if it does most of what you need it to do in its current state.  If you or your team were willing to 

[...]

PR #220, opened 2016-06-11
Update docs to clarify the state of Workbench and Tasks

fbender:
Per #219


corranwebster:
Thanks for the clarified phrasing.

PR #221, opened 2016-06-14
Less TraitsUI in Pyface

corranwebster:
This is a start at untangling some of the interdependencies between Pyface and TraitsUI.  In an ideal world Pyface shouldn't need anything from TraitsUI.  It may not be feasible to fix some heavily entangled sections of code (eg. the `pyface.dock` package), but this PR should remove all of the dependencies when running under Qt.

This copies a few things from TraitsUI, so once we're good with this, there may be a point to making an equivalent PR on the TraitsUI side to remove duplication.

~~Tests are needed.~~

#### Update

This PR is ready for review.  It makes the following concrete changes:

* creates a `pyface.ui_traits` module which holds the traits from `traitsui.ui_traits` that are used by Pyface:
  - the `Image` trait for image resources and associated utility functions
  - the `HasMargin`, `HasBorder`, `Alignment` and `Position` traits and associated code
* moves the `traitsui.images` package for handling image libraries to `pyface.images`, rebuilds the standard image library zip files, and fixes the following bugs:
  - library timestamps were not in UTC, creating problems when libraries were created in different timezones than where they were being used
  - the library code assumed an `image_size` method on the `ImageResource` class, this PR provides the method
* moves the `PyMimeData` class used for Qt clipboard and DnD support from traitsui to pyface.
* using deferred imports of some TraitsUI classes used in `Instance` traits in Pyface classes that interoperate with TraitsUI.
* adds some support methods for testing to allow skipping tests if traitsui is not being used.
* tests for the above code.

This code doesn't try to untangle the Dock and Workbench code and the way that it interleaves with TraitsUI in the Wx backend.


pyface/ui/qt4/tasks/tests/test_split_editor_area_pane.py
```diff
@@ -321,6 +326,7 @@ def test_persistence(self):
         self.assertEquals(right_bottom.items[0].id, 1)
         self.assertEquals(right_bottom.items[1].id, 2)
 
+    @skip_if_no_traitsui
```
rkern: `ViewWithTabsEditor` should be rewritten to not use `traitsui` instead of skipping these tests. It should be straightforward to do in Qt. Probably even shorter.

pyface/ui/qt4/util/tests/test_modal_dialog_tester.py
```diff
@@ -115,6 +119,7 @@ def raise_error(tester):
                 tester.open_and_run(when_opened=raise_error)
             self.assertIn('ZeroDivisionError', alt_stderr)
 
+    @skip_if_no_traitsui
```
rkern: These should be rewritten to not use `traitsui`.

corranwebster:
Switch timezones and the timestamps are out of sync, so it tries to rebuild the image volumes on import and fails because of a bug in the size default handling.  Would not have noticed this if I hadn't switched timezones between recompiling the zip files and now.  The `pyface.image.image` needs some work.

corranwebster:
@itziakos and/or @jonathanrocher do either of you have time to do a review of this?

pyface/image/image.py
```diff
@@ -0,0 +1,1336 @@
...
+    )"""
+
+#-------------------------------------------------------------------------------
+#  Returns the contents of the specified file:
+#-------------------------------------------------------------------------------
+
+def read_file ( file_name ):
+    """ Returns the contents of the specified *file_name*.
+    """
+    fh = file( file_name, 'rb' )
```
itziakos: can we use a `with` context for this code?

pyface/image/image.py
```diff
@@ -0,0 +1,1336 @@
...
+from pyface.resource.resource_reference import (ImageReference,
+                                                ResourceReference)
+from pyface.ui_traits import HasMargin, HasBorder, Alignment
+
+#---------------------------------------------------------------------------
+#  Constants:
+#---------------------------------------------------------------------------
+
+# Standard image file extensions:
+ImageFileExts = ( '.png', '.gif', '.jpg', 'jpeg' )
```
itziakos: Should we change the style (e.g. `(  ` and `  )` ) to be more pep8?

pyface/image/image.py
```diff
@@ -0,0 +1,1336 @@
...
+#-------------------------------------------------------------------------------
+#  'FastZipFile' class:
+#-------------------------------------------------------------------------------
+
+class FastZipFile ( HasPrivateTraits ):
+    """ Provides fast access to zip files by keeping the underlying zip file
+        open across multiple uses.
+    """
+
+    # The path to the zip file:
```
itziakos: '#' -> '#:' so that they can be picked up by autodoc

pyface/image/image.py
```diff
@@ -0,0 +1,1336 @@
...
+
+        return (ImageVolumeInfoTextTemplate % ( description, self.copyright,
+                                                license, images ))
+
+    #-- Public Methods ---------------------------------------------------------
+
+    def clone ( self ):
+        """ Returns a copy of the ImageVolumeInfo object.
+        """
+        return self.__class__(**self.trait_get('description', 'copyright',
```
itziakos: Maybe use the `clone` method of HasTraits.

pyface/image/image.py
```diff
@@ -0,0 +1,1336 @@
...
+    def update ( self ):
+        """ Updates the contents of the image volume from the underlying
+            image store, and saves the results.
+        """
+        # Unlink all our current images:
+        for image in self.images:
+            image.volume = None
+
+        # Make sure the images are up to date by deleting any current value:
+        del self.images
```
itziakos: I would think that `self.images = []`  or self.reset_traits(['images']) is better than removing the `images` attribute

pyface/image/image.py
```diff
@@ -0,0 +1,1336 @@
...
+            # new file the original name. Note that unlocking the original zip
+            # file after the previous close sometimes seems to take a while,
+            # which is why we repeatedly try the rename until it either succeeds
+            # or takes so long that it must have failed for another reason:
+            temp_name = path + '.$$$'
+            for i in range( 50 ):
+                try:
+                    rename( path, temp_name )
+                    break
+                except:
```
itziakos: maybe `expect Exception`

pyface/image/image.py
```diff
@@ -0,0 +1,1336 @@
...
+        """ Returns the ImageVolumeInfo object that corresponds to the
+            image specified by **image_name**.
+        """
+        for info in self.info:
+            if ((len( info.image_names ) == 0) or
+                (image_name in info.image_names)):
+                return info
+
+        # Should never occur:
+        return None
```
itziakos: probably raise an error here?

pyface/image/image.py
```diff
@@ -0,0 +1,1336 @@
...
+                return image
+
+            # Make sure the correct image cache directory exists:
+            cache_dir = join( image_cache_path, self.volume_name )
+            if not exists( cache_dir ):
+                makedirs( cache_dir )
+
+            # Write the image data to the cache file:
+            cache_file = join( cache_dir, self.file_name )
+            fh         = file( cache_file, 'wb' )
```
itziakos: I think we should probably use `with open(...)`

pyface/ui_traits.py
```diff
@@ -0,0 +1,260 @@
...
+                 'or a tuple with 1, 2 or 4 integers in that range that can be '
+                 'used to define one')
+
+    def validate (self, object, name, value):
+        """ Validates that a specified value is valid for this trait.
+        """
+        if isinstance(value, int):
+            try:
+                value = self.klass(value)
+            except:
```
itziakos: `except Exception`

itziakos:
@corranwebster, I has a quick look, a few minor comments

pyface/image/image.py
corranwebster: No, I'm planning to do an autopep8-ing of Pyface at some point, but it shouldn't be part of this PR.

PR #222, opened 2016-06-21
pyface fixes not dependent on tasks

robmcmullen:
Here are a few changes that are not related to adding tasks support for wx. The tasks support will be a much larger pull request so I wanted to isolate the changes that don't depend on tasks.


corranwebster:
LGTM

Thanks for the contribution.

PR #223, opened 2016-06-21
Tasks support for wx

robmcmullen:
Here's my big patch for tasks support for wxPython. It requires wxPython (classic) 3.0.2; wxPython 2.8 is not supported. It uses the wx.lib.agw.aui toolkit for the central pane management and toolbars. The AuiManager class used as the basis for the TaskLayout is not quite as flexible as the Qt layout manager in that AUI can't nest VSplitter and HSplitters arbitrarily deeply; my implementation is limited to 2 levels: either a VSplitter inside an HSplitter or an HSplitter inside a VSplitter.

I've included a bunch of examples in the examples/tasks/steps directory, which includes a series of small programs that I used to build up tasks support for wx.


corranwebster:
Thanks for this very useful contribution!

Test failures on CI are presumably because Travis CI installs wx 2.8, I'm guessing.  I'll take a quick look into what it might take to get a wxPython 3 version built in the CI system; but even then it may be worth having a test decorator that skips tests if it detects wx < 2.8

pyface/wx/aui.py
```diff
@@ -0,0 +1,275 @@
...
+# capable of supporting the windowing flexibility needed by tasks. Therefore,
+# only AGW's pure-python AUI implementation is used.
+
+from wx.lib.agw import aui
+
+# AGW's library does need some patching for some usability differences desired
+# for pyface but not for the standard wxPython version
+
+class PyfaceAuiNotebook(aui.AuiNotebook):
+    SetPageToolTip = aui.AuiNotebook.SetPageTooltip
```
corranwebster: This seems to be causing some problems for running under wx < 3.  Can there perhaps be wrapped as standard methods (or even omitted entirely?).

corranwebster:
Also, a number of the failures seem to be tweaking an existing bug on this line here: https://github.com/enthought/pyface/blob/master/pyface/action/action_manager_item.py#L39 It should read:

    parent = Instance('pyface.action.group.Group')

I could fix in a separate PR, but it may be simpler to fix it in this PR.

pyface/tasks/i_dock_pane.py
```diff
@@ -36,6 +36,13 @@ class IDockPane(ITaskPane):
     # Whether the pane is currently visible.
     visible = Bool(False)
 
+    # Whether the title bar of the pane is currently visible.
+    caption_visible = Bool(True)
+
+    # AUI ring number; note that panes won't be movable out of their ring
+    # number.  This is a way to isolate panes
+    dock_layer = Int(0)
```
corranwebster: I think that this Trait should be on the Wx `DockPane` implementation, since its specific to AUI.

pyface/tasks/i_task_window_backend.py
```diff
@@ -31,6 +31,18 @@ def destroy(self):
         but this method may be used to perform additional cleanup.
         """
 
+    def begin_batch_window_updates(self):
```
corranwebster: Note hugely critical, but a context manager that calls this method on `__enter__` and the corresponding end method `__exit__` would make the Wx code a bit more robust, I think.

pyface/ui/wx/tasks/dock_pane.py
```diff
@@ -0,0 +1,255 @@
...
+        """ Create and return the toolkit-specific contents of the dock pane.
+        """
+        return wx.Window(parent, name=self.task.id + ':' + self.id)
+
+#    ###########################################################################
+#    # Protected interface.
+#    ###########################################################################
+#
+#    @contextmanager
+#    def _signal_context(self):
```
corranwebster: Commented-out code from Qt implementation.

pyface/ui/wx/tasks/editor_area_pane.py
```diff
@@ -0,0 +1,176 @@
...
+    ###########################################################################
+
+    def create(self, parent):
+        """ Create and set the toolkit-specific control that represents the
+            pane.
+        """
+        logger.debug("editor pane parent: %s" % parent)
+        # Create and configure the tab widget.
+        self.control = control = PyfaceAuiNotebook(parent, agwStyle=self.style)
+#        control.tabBar().setVisible(not self.hide_tab_bar)
```
corranwebster: This and the code below are presumably from attempts to hide/show the tab bar?  Do you want to keep around for reference, or can they be removed?

pyface/ui/wx/tasks/main_window_layout.py
```diff
@@ -0,0 +1,238 @@
...
+                if item.id == layout.active_tab:
+                    active_pane = dock_pane
+                dock_pane.add_to_manager(tabify_pane=first_pane)
+                if not first_pane:
+                    first_pane = dock_pane
+                dock_pane.visible = True
+
+            # Activate the appropriate tab, if possible.
+            if not active_pane:
+                # By default, Qt will activate the last widget.
```
corranwebster: Qt mentioned in comment, I assume that this is try as well for Wx?

pyface/ui/wx/tasks/main_window_layout.py
```diff
@@ -0,0 +1,238 @@
...
+    def _get_division_orientation(self, one, two, splitter=False):
+        """ Returns whether there is a division between two visible QWidgets.
+
+        Divided in context means that the widgets are adjacent and aligned along
+        the direction of the adjaceny.
+        """
+        united = one.united(two)
+        if splitter:
+            sep = self.control.style().pixelMetric(
+                QtGui.QStyle.PM_DockWidgetSeparatorExtent, None, self.control)
```
corranwebster: Qt code seems to have snuck in here.  Is this whole method needed.

pyface/ui/wx/tasks/main_window_layout.py
```diff
@@ -0,0 +1,238 @@
...
+        
+        return 0
+
+    def _get_tab_bar(self, dock_widget):
+        """ Returns the tab bar associated with the given QDockWidget, or None
+        if there is no tab bar.
+        """
+        dock_geometry = dock_widget.geometry()
+        for child in self.control.children():
+            if isinstance(child, QtGui.QTabBar) and child.isVisible():
```
corranwebster: Again, some Qt code remaining here.

corranwebster:
This is awesome work.  There are a few places where Qt code still seems to be around in the Wx implementation, and a few other minor issues, but if we can get the tests passing I'd be happy to merge and then fix things up afterwards.

robmcmullen:
Thanks for looking over the code. Before you commit, let me see if I can
address these issues you raised. One problem with wx 2.8 is that the aui
code included there is so old that the task window and editor area pane
will be completely broken. So I don't know if it's worth addressing the CI
failures because the actual task windows won't work. I could try to raise
an exception when using tasks with 2.8, I suppose. Or, would it be possible
to require wx 3.0 for the entire enthought suite?

Or should we wait for Phoenix before requiring 3.0? I do intend to begin
the process to port it to Phoenix when Robin releases it on pypi. There are
probably enough changes required in the move to Phoenix that supporting
both styles of wx won't be practical.

robmcmullen:
Removed all the Qt stuff that I could find still laying around, as well as that begin/end batch window stuff that I had originally coded to help with the speed when switching tasks, but now doesn't seem to make an appreciable difference. Maybe it was different under wx 2.8; I don't remember at this point.

corranwebster:
Looking at where things are breaking under 2.8, I think that the CI failures are independent of Tasks but caused by the modified Aui classes: I suspect most of the Tasks CI tests are currently skipped on Wx.

It would be nice to move to wxPython 3+, but we are still finding bugs with it (although it mostly works); and we aren't running CI tests against wxPython 3.

So I guess what I'd like, if its possible, is to have all the non-tasks CI tests that were running before to still be working in this PR.  But if the Tasks portion of the CI tests fail on 2.8, I don't care.

Failing that, I'd want to have a separate PR (which I wouldn't expect you to write) that gets CI tests working under wxPython 3.

I might play with the code a bit today to see if getting tests to pass is easy or hard.

corranwebster:
@robmcmullen I have a PR against your branch which will hopefully fix the tests under wxPython 2.8.  If you could merge that would be awesome.  If that means that the tests pass, I'd be willing to consider merging this branch.

robmcmullen:
OK, thanks. Merged!

robmcmullen:
Looks like the build passed, finally.

corranwebster:
Excellent!  Don't worry about the code coverage failing - this can be merged without that.

Thanks for working on this!

pyface/tasks/task_window.py
```diff
@@ -348,6 +347,9 @@ def _destroy_state(self, state):
             if state.menu_bar_manager:
                 state.menu_bar_manager.destroy()
             for tool_bar_manager in state.tool_bar_managers:
+                # FIXME: toolbars were not getting removed, but this isn't the right place for it
+                pane = self._window_backend.window._aui_manager.GetPane(tool_bar_manager.id)
```
corranwebster: @mdickinson pointed out that this seems to be assuming a Wx backend and will likely break under Qt.

PR #225, opened 2016-07-13
Fix task window title

### LINKED ISSUE #218: Updating a task's name doesn't trigger an update of the window title
It seems like that is because https://github.com/enthought/pyface/blob/master/pyface/tasks/task_window.py#L32
only listens to a change of the active task, not a change in its name.

corranwebster:
Fixes #218.  Includes tests.  We now listen to `active_task.name` instead of just `active_task`.

### DIFF
--- a/pyface/tasks/tests/test_task_window.py
+++ b/pyface/tasks/tests/test_task_window.py
@@ -1,0 +1,66 @@
+import unittest
+
+from traits.testing.unittest_tools import UnittestTools
+from pyface.tasks.api import Task
+from ..task_window import TaskWindow
+
+
+class TestTaskWindow(unittest.TestCase, UnittestTools):
+
+    def test_title(self):
+        task_1 = Task(name='Test Task')
+        task_2 = Task(name='Test Task 2')
+        task_window = TaskWindow(tasks=[task_1, task_2])
+
+        # initially nothing
+        self.assertEqual(task_window.title, '')
+
+        # activate task
+        with self.assertTraitChanges(task_window, 'title', count=1):
+            task_window.active_task = task_1
+        self.assertEqual(task_window.title, 'Test Task')
+
+        # change task name
+        with self.assertTraitChanges(task_window, 'title', count=1):
+            task_1.name = 'Changed Name'
+        self.assertEqual(task_window.title, 'Changed Name')
+
+        # change active task
+        with self.assertTraitChanges(task_window, 'title', count=1):
+            task_window.active_task = task_2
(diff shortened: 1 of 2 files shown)


corranwebster:
@jonathanrocher do you have time to review this?

jonathanrocher:
Yes, I will take a look today.

jonathanrocher:
Functionally, I confirm that this works. Reviewing code for styling now...

pyface/tasks/tests/test_task_window.py
```diff
@@ -0,0 +1,72 @@
...
+from traits.testing.unittest_tools import UnittestTools
+from pyface.tasks.api import Task
+from ..task_window import TaskWindow
+
+
+class TestTaskWindow(unittest.TestCase, UnittestTools):
+
+    def test_title(self):
+        task_1 = Task(name='Test Task')
+        task_2 = Task(name='Test Task 2')
```
jonathanrocher: Would move these 2 lines into a `setUp`.

pyface/tasks/tests/test_task_window.py
```diff
@@ -0,0 +1,72 @@
...
+        task_window = TaskWindow(tasks=[task_1, task_2])
+
+        # initially nothing
```
jonathanrocher: I would recommend to break each of these sections to a separate test to make them more granular.

pyface/tasks/tests/test_task_window.py
jonathanrocher: Actually the definition of `task_window` can also be moved to a `setUp`.

pyface/tasks/tests/test_task_window.py
```diff
@@ -0,0 +1,72 @@
...
+        with self.assertTraitChanges(task_window, 'title', count=1):
+            task_window.active_task = None
+        self.assertEqual(task_window.title, "Window title")
+
+        # change window title
+        with self.assertTraitChanges(task_window, 'title', count=1):
+            task_window.title = "New Window title"
+        self.assertEqual(task_window.title, "New Window title")
+
+        # go back to getting title from task
```
jonathanrocher: Again, I would break each of these sections into a separate test.

jonathanrocher:
Only minor recommendations on making the tests more granular. Other than that, LGTM.

corranwebster:
Fixes as suggested - didn't use a setUp since eventually we should have more tests of other aspects which need different set up, but instead moved object creation into a utility function.

corranwebster:
...and appveyor is failing because pip was upgraded, by the look of things.

PR #226, opened 2016-07-14
Fix Appveyor tests

corranwebster:
Replace `--no-use-wheel` with `--no-binary :all:` in appveyor pip install.


corranwebster:
Self-merging, as this is a simple change that needs an immediate fix.

PR #227, opened 2016-07-25
Split task application

corranwebster:
This splits up the `TaskApplication` from @jonathanrocher's #198, breaking out the things that are not Tasks-specific into their own base class that could be used for more basic applications (eg. pure Pyface, TraitsUI or Workbench).

It also cleans up the example somewhat.

PR #229, opened 2016-08-25
Split Task Application

corranwebster:
This splits up the TaskApplication from @jonathanrocher's #198, breaking out the things that are not Tasks-specific into their own base class that could be used for more basic applications (eg. pure Pyface, TraitsUI or Workbench).

It also cleans up the example somewhat.

This is a replacement PR for #227 which got mangled by a bad merge.


examples/tasks/advanced/run.py
```diff
@@ -1,66 +1,45 @@
...
+    app = TaskApplication(
+        id="MyPyfaceApplication",
+        name="Python Editor",
+        window_size=(800, 600),
+    )
+    # hook up listener to application initialized event
+    app.on_trait_change(
+        lambda event: create_task_window(event.application, files=argv[1:]),
+        "application_initialized"
+    )
```
jonathanrocher: Personally, I find this way of doing things too complicated, compared to adding a listener to a subclass.

examples/tasks/advanced/run.py
```diff
@@ -1,66 +1,45 @@
...
 def main(argv):
-    """ A more advanced example of using Tasks.
-    """
-    app = MyApplication()
+    """ A more advanced example of using Tasks. """
+    app = TaskApplication(
+        id="MyPyfaceApplication",
+        name="Python Editor",
+        window_size=(800, 600),
+    )
```
jonathanrocher: I am fine with going this route. I understand that for the things that my class `MyApplication ` override attributes that can just be passed. On the other hand, to teach people how to scale on top of `TaskApplication`, they will need to subclass it at some point. That's what I see be the purpose of the `advanced` example.

jonathanrocher: Never mind, since there is the `run_fancy_app` file. Please ignore. +1

examples/tasks/advanced/run_fancy_app.py
```diff
@@ -33,28 +33,33 @@ class MyApplication(TaskApplication):
...
-        return True
+    def _on_window_closing(self, window, trait, old, new):
+        """ Ask confirmation when a window is closed. """
+        from pyface.api import confirm, YES
+
+        msg = "Are you sure you want to close the window?"
+        return_code = confirm(None, msg)
+        if return_code != YES:
+            logger.debug("Window closing even was veto-ed")
+            new.veto = True
```
jonathanrocher: Thanks! I had the same code to add to the example, and was waiting for your PR. +1

examples/tasks/advanced/run.py
corranwebster: I agree we should probably have one example which shows how to use subclassing.

corranwebster:
These need tests before merging (at least for `BaseApplication`), I think.

examples/tasks/advanced/run_fancy_app.py
```diff
@@ -33,28 +33,33 @@ class MyApplication(TaskApplication):
...
     # MyApplication interface
     # -------------------------------------------------------------------------
 
-    #: Hook to add global schema additions to tasks/windows
-    extra_actions = List(Instance(
-        'pyface.tasks.action.schema_addition.SchemaAddition'
-    ))
+    logdir = Str
 
+    logdir_path = Directory
```
jonathanrocher: Why use 2 different attributes for the log directory? `logdir` doesn't seem to be used...

pyface/base_application.py
```diff
@@ -0,0 +1,286 @@
...
+    # 'BaseApplication' interface
+    # -------------------------------------------------------------------------
+
+    # Branding ----------------------------------------------------------------
+
+    #: Human-readable application name
+    name = Str('Default Application')
+
+    #: Human-readable company name
+    company = Str('Enthought')
```
jonathanrocher: Personally, I would remove `Enthought` and use a more neutral default like for your `name`.

pyface/base_application.py
```diff
@@ -0,0 +1,286 @@
...
+            if stopped:
+                self.stopped = ApplicationEvent(application=self,
+                                                event_type='stopped')
+                logger.info('---- Application stopped ----')
+
+                # exit normally
+                sys.exit()
+
+        # exit with an error
+        sys.exit(1)
```
jonathanrocher: What is the point of these calls to `sys.exit`, as opposed to exiting with an exception?

examples/tasks/advanced/run_fancy_app.py
```diff
@@ -75,21 +80,34 @@ def _setup_logging(self):
...
-        return_code = confirm(None, msg)
-        if return_code != YES:
-            logger.debug("Window closing even was veto-ed")
-            new.veto = True
+        return SGroup(
+            Action(name='Exit' if IS_WINDOWS else 'Quit',
+                   accelerator='Alt+F4' if IS_WINDOWS else 'Ctrl+Q',
+                   on_perform=self.exit),
+            id='QuitGroup', name='Quit',
+        )
```
jonathanrocher: Personally, I find this general enough to be in the `Application` class.

jonathanrocher: Or `TaskApplication` since these are `TaskAction`?

pyface/application.py
```diff
@@ -0,0 +1,259 @@
...
+class Application(BaseApplication):
+    """ A basic Pyface application.
+
+    This handles setting up starting up the GUI, window management, logging,
+    and other common features that we want when creating a GUI application.
+
+    Often you will want to subclass this application, but it can be used as-is
+    by hooking a listener to the `application_initialized` event::
+
+
```
jonathanrocher: I would add the imports here `from pyface.api import ApplicationWindow, HeadingText`

pyface/application.py
```diff
@@ -0,0 +1,259 @@
...
+        except Exception:
+            # something went wrong, use base class excepthook
+            super(Application, self)._excepthook(type, value, traceback)
+        finally:
+            sys.exit(1)
+
+    # Destruction methods -----------------------------------------------------
+
+    def _can_exit(self):
+        """ Check with each window to see if it can be closed """
```
jonathanrocher: I would add that to use this, a listener on `window.closing` has been added in `add_window`.

pyface/application.py
```diff
@@ -0,0 +1,259 @@
...
+    def _setup_logging(self):
+        """ Initialize logger.
+
+        By default send logging to a stream and set the log level to info.
+        Most applications will want to replace this with something more
+        appropriate to their needs.
+        """
+        logger = logging.getLogger()
+        logger.addHandler(logging.StreamHandler())
+        logger.setLevel(logging.INFO)
```
jonathanrocher: Personally, I would probably leave this empty, and leave this to the user to set.

jonathanrocher:
Done with review. This looks great, thanks for the contribution @corranwebster ! 

I have one more general question about the class structure: I love the idea of breaking things up into `TaskApplication` vs `Application`. The need for the BaseApplication is less clear to me: can you think of situations where someone would want to subclass from `BaseApplication` but not `Application`?

corranwebster:
The immediate use-case for `BaseApplication` are command-line apps.  But you can certainly think of a code-base where there is a GUI app and a command-line app which share a lot of the same code. The other use-case is for situations where you need to incorporate some other event loop: eg. Twisted, Tornado, or a number of others, particularly web servers.

Technically that probably doesn't belong in Pyface, but there's nowhere better I can think of right now.  

It also makes testing easier, since we can test basic functionality without getting a GUI involved.

jonathanrocher:
Makes sense. Does that motivate to rename `Application` to `GuiApplication` to make the division of labor clearer?

jonathanrocher:
@corranwebster Shall we move this along? I have a few comments left, but this is mostly ready to be merged into my branch...

jonathanrocher:
@corranwebster This is getting stale. I had a few comments. Could you respond so that I can merge that into my branch and add unit tests. I believe this is still valuable work.

examples/tasks/advanced/run_fancy_app.py
```diff
@@ -33,28 +33,33 @@ class MyApplication(TaskApplication):
...
-        self.create_new_task_window()
-        return True
+    def _on_window_closing(self, window, trait, old, new):
+        """ Ask confirmation when a window is closed. """
+        from pyface.api import confirm, YES
+
+        msg = "Are you sure you want to close the window?"
+        return_code = confirm(None, msg)
+        if return_code != YES:
+            logger.debug("Window closing even was veto-ed")
```
itziakos: even -> event

pyface/application.py
```diff
@@ -0,0 +1,259 @@
...
+
+        # Keep a handle on all windows created so that non-active windows don't
+        # get garbage collected
+        self.windows.append(window)
+
+        # set up window listeners
+        # NOTE: 'activated' is not fired on a window when the window first
+        # opens and gets focus. It is only fired when the window comes from
+        # lower in the stack to be the active window.
+        window.on_trait_change(self._on_activate_window, 'activated')
```
itziakos: There should be no need to setup these change handlers here. The application should setup extended trait handlers looking into the `windows` list

e.g.
```
@on_trait_change('windows:activate'):
def _on_activate_window(self, object, name, value):
      window = object
      ...
```

pyface/application.py
```diff
@@ -0,0 +1,259 @@
...
+        their start method, and should call the superclass start() method
+        before doing any work themselves.
+        """
+        ok = super(Application, self).start()
+        if ok:
+            # set up logging so messages can be displayed on splash screen
+            self._setup_logging()
+
+            # create the GUI so that the splash screen comes up first thing
+            self.gui
```
itziakos: This code assumes that `self.gui` will trigger the default initializer which as a side effect will create a GUI will the desired splash screen. 

I would suggest the following

- There is no _gui_default ,method
- The gui trait is set only once here `self.gui = GUI(splash_screen=self.splash_screen)` and reset on stop.
- The gui trait is considered readonly

pyface/application.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def _on_window_closing(self, window, trait, old, new):
+        """ Custom behavior when the closing of a window is requested."""
+        pass
+
+    def _on_window_closed(self, window, trait, old, new):
+        """ Listener that ensures window handles are released when closed.
+        """
+        self.windows.remove(window)
+        window.on_trait_change(self._on_activate_window, 'activated', )
```
itziakos: these methods do not remove the handler but hook it again

corranwebster:
@itziakos If you could look this over at your convenience (no rush) I would appreciate it.

Test failures are unrelated segfaults in the dialog testing code, I think.

itziakos:
> Test failures are unrelated segfaults in the dialog testing code, I think.

The failures mainly happen on pyqt so it could pyface usage of qt that causes the segfaults.

jonathanrocher:
Note: this is targeting my original branch, not master. I am also ok with merging into it. I have a couple more comments, but can make them there in the hope to get this over the finish line faster... So 👍 for me too.

corranwebster:
I'm just going to merge this - any issues with tests (I don't expect any) can be fixed in the original PR.

PR #230, opened 2016-08-30
MAINT: Update EndEdit signature.

jvkersch:
In wxPython 3 the signature and implementation of `EndEdit` is slightly different: it is now a method that takes 5 arguments and validates the result, with a new method `ApplyEdit` actually applying the edits.

There was one place in Pyface where `EndEdit` was implemented, and this PR updates that definition to match the changes. I didn't write a test, but below is a minimal working example. Running the example without this PR will result in a table that cannot be edited, with the message `TypeError: EndEdit() takes exactly 4 arguments (5 given)` being printed to the command line.

``` python
from traits.etsconfig.api import ETSConfig
ETSConfig.toolkit = 'wx'
from traitsui.api import View
from traits.api import HasTraits, Str, List


class Employee(HasTraits):
    first_name = Str
    last_name = Str


class Company(HasTraits):
    employees = List(Employee)
    view = View('employees')


company = Company()
company.employees.append(Employee(first_name="A", last_name="B"))
company.employees.append(Employee(first_name="C", last_name="D"))
company.configure_traits()
```


jvkersch:
Note that this supposes that only wxPython 3 is supported. If backwards compatibility with version 2 is needed, this will have to be a bit more subtle.

corranwebster:
I'm not sure that we want to drop Wx 2.8+ compatibility just yet, but I could be convinced otherwise.

jdmarch:
@RobinD42 ?? ^^

RobinD42:
2.8.10 is 7 years old, although 2.8 is still found out in the wild now and then.  (But I think a large chunk of them are are getting it from us.)  3.0.2 is about 2 years old and most users have moved on to it a long time ago.

jvkersch:
@corranwebster In light of Robin's comment, what do you think we should do here? I'm tempted to leave as is, but I could add some code for 2.8, too.

corranwebster:
I've checked, and we have clients who are still on 2.8.  I think we still need 2.8 support for at least a little longer.

What is the failure mode on Wx < 3 for this PR?

jvkersch:
> What is the failure mode on Wx < 3 for this PR?

@corranwebster This seems to work fine even with wx 2.9 (which is what I used for testing), presumably because the signature of `ApplyEdit` hasn't changed. Do you think this is good to go?

corranwebster:
Yep, if it works on 2.9 I'm good to go with it.

PR #231, opened 2016-09-01
Remove -toolkit references from docs

corranwebster:
See https://github.com/enthought/traits/pull/326.  Should wait until that is merged.


rkern:
LGTM!

PR #247, opened 2016-12-18
Fix for toolbars in non-AUI windows

### LINKED ISSUE #246: Toolbars don't work with TraitsUI with Wx Toolkit
There is a regression in TraitsUI introduced by #223 (the Tasks for Wx PR) where toolbars no longer work under Wx.  The cause is that the underlying control was changed from a `wx.ToolBar` to an `wx.aui.AuiToolBar`, which while more featureful, is not a subclass of `wx.ToolBar` and so can't be set as the toolbar of a non-AUI window.

I don't think it is practical to convert TraitsUI to using AUI, particularly given that it uses DockWindow which covers similar ground.

Possible solutions include:
- carefully reverting the changes in #223 around the Toolbar.
- allowing the window to request different toolbar controls types depending on need (with the default being a wx.ToolBar)

corranwebster:
Fixes #246. AUI windows must explicitly ask for an AUI toolbar; other code is adapted to the possibility of getting one or the other type.

### DIFF
--- a/pyface/ui/wx/action/action_item.py
+++ b/pyface/ui/wx/action/action_item.py
@@ -370,11 +370,16 @@ class _Tool(HasTraits):
         # Set the initial checked state.
         tool_bar.ToggleTool(self.control_id, action.checked)
 
-        # Set the initial enabled/disabled state of the action.
-        tool_bar.EnableTool(self.control_id, action.enabled)
+        if hasattr(tool_bar, 'ShowTool'):
+            # Set the initial enabled/disabled state of the action.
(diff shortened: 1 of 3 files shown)


robmcmullen:
Great! I was searching my notes for why I had to change to the AUI toolbars; IIRC it was because the wx toolbars wouldn't work as children of the AUI manager. And I remember having difficulty trying to get multiple toolbars to work as well as showing or hiding whole toolbars. The AUI toolbars fixed that, but I didn't think to check traitsui.

Sorry, I had never thought to check traitsui because it didn't occur to me that you could generate toolbars with it. Thanks for figuring out the fix.

PR #255, opened 2017-01-11
Appveyor EDM support

corranwebster:
This PR uses EDM to create test environments for Appveyor CI runs, the same way we are for TravisCI.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,18 +1,20 @@
 build: false
 shallow_clone: true
+platform: x64
 
 environment:
 
   global:
     distutils_use_sdk: 1
-    ets_toolkit: "qt4"
 
   matrix:
 
-    - python: "C:/Python27-x64"
+    - python: "2.7"
+      toolkit: "pyside"
       sdkver: "v7.0"
 
-    - python: "C:/Python34-x64"
+    - python: "3.5"
+      toolkit: "pyqt"
       sdkver: "v7.1"
 
 cache:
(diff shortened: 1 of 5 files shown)


corranwebster:
Looks like there is an issue with finding the `edm` command while running from invoke (at least at first approximation).

appveyor.yml
```diff
@@ -21,13 +23,12 @@ cache:
...
 
 install:
-  - ps: if ((Test-Path "c:/temp") -eq 0) { mkdir c:/temp }
-  - ps: python -m pip install --upgrade --no-binary wheel pip
-  - ps: pip install --upgrade wheel
-  - ps: pip --version
-  - cmd /v:on /e:on /c ".\appveyor-install.cmd"
+  - cmd: appveyor-bootstrap-edm.cmd setup
+  - ps: edm info
+  - cmd /e:on /v:on /c ".\appveyor-install.cmd"
```
itziakos: there is probably no need for a separate install and test batch files. We can probably call edm run from appveyor.yml directly

appveyor.yml
```diff
@@ -21,13 +23,12 @@ cache:
 init:
   - ps: $Env:sdkbin = "C:\Program Files\Microsoft SDKs\Windows\" + $Env:sdkver + "\Bin"
```
itziakos: These environment variables are not needed

itziakos:
failures are intermittent #3215 and #3307

corranwebster:
Closed in favour of #298.

PR #260, opened 2017-01-12
Allow plugin toolkits

corranwebster:
This PR makes the toolkit selection mechanism a lot cleaner, clearer and more standard.  We use the `pkg_resources` mechanisms to look for `pyface.toolkit` entry points.  This allows third-party libraries to add new backends.  The core library provides entry points for `qt4`, `wx` and `null` backends.

We also provide a base class for the `toolkit_object` importers which allows multiple packages to be searched for backend implementations of toolkit objects.  This allows applications to more easily support custom widgets for multiple toolkits.

### DIFF
--- a/pyface/base_toolkit.py
+++ b/pyface/base_toolkit.py
@@ -9,34 +9,26 @@ toolkit, but plugin authors are free to use whatever methods they like.
 import os
 import sys
 
-
-class Unimplemented(object):
-    """ An unimplemented toolkit object
-
-    This is returned if an object isn't implemented by the selected
-    toolkit.  It raises an exception if it is ever called, ie. used as a
-    toolkit object factory.
-    """
-
-    def __init__(self, toolkit, name):
-        self._toolkit = toolkit
-        self._name = name
-
-    def __call__(self, *args, **kwargs):
-        msg = "the {} pyface backend doesn't implement {}"
-        raise NotImplementedError(msg.format(self._toolkit, self._name))
+from traits.api import HasTraits, List, Str
 
 
-class Toolkit(object):
-    """ A basic toolkit implementation for use by specific toolkits
+class Toolkit(HasTraits):
+    """ A basic toolkit implementation for use by specific toolkits.
 
     This implementation uses pathname mangling to find modules and objects in
-    those modules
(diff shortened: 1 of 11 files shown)


pyface/base_toolkit.py
```diff
@@ -0,0 +1,82 @@
...
+            separated by a colon.
+        """
+        from importlib import import_module
+
+        mname, oname = name.split(':')
+
+        for package in self.packages:
+            try:
+                module = import_module('.' + mname, package)
+                obj = getattr(module, oname, None)
```
itziakos: I would move the getattr into an `else` section

pyface/toolkit.py
```diff
@@ -49,105 +72,82 @@ def provisional_toolkit(toolkit_name):
...
-                warnings.warn(msg.format(toolkit_name), RuntimeWarning)
-            except ImportError as exc:
+    # Try known toolkits first.
+    known_toolkits = ('qt4', 'wx')
+    for tk in known_toolkits:
+        try:
+            with provisional_toolkit(tk):
+                return import_toolkit(tk)
+        except RuntimeError as exc:
+            if logger.getEffectiveLevel() <= logging.INFO:
```
itziakos: It is better to have a single logging message. I would suggest something like

```python
exc_info = logger.getEffectiveLevel() <= logging.INFO
msg = "Could not import Pyface backend %r"
level = 'ERROR' if exc_info else 'INFO'
logger.log(level, msg, tk, excinfo=exc_info)
```

(it is also advised to use `%` for log messages)

pyface/toolkit.py
```diff
@@ -49,105 +72,82 @@ def provisional_toolkit(toolkit_name):
...
+        return toolkit_object
+    except ImportError as exc:
+        logger.exception(exc)
+        raise ImportError("Unable to import a pyface backend for any toolkit")
 
+    return None
 
 # Do this once then disappear.
-_init_toolkit()
+toolkit_object = _init_toolkit()
```
itziakos: We are setting this variable twice?

corranwebster:
I'm fairly happy with this as it stands.  Any further comments?

PR #264, opened 2017-03-10
Qt5 v3

corranwebster:
This is another attempt at Qt5 compatibility.  Following #256 we should have removed all old-style signals and slots, so at the object level, we have compatibility with Qt5.  This PR takes a different approach to the previous PRs like #171 and is inspired by a comment by @pankajp there.

This PR tries to make the PyQt5 API compatible with PyQt4, so that code which depends on pyface.qt does not need to be modified.  We may, at some future point, add a pyface.qt5 module or something similar which attempts forward compatibility instead of backwards compatibility.

To do:

- [x] CI integration for Qt5.  This is still a blocker for merging.

### DIFF
--- a/pyface/qt/__init__.py
+++ b/pyface/qt/__init__.py
@@ -9,7 +9,16 @@
 # Description: Qt API selector. Can be used to switch between pyQt and PySide
 #------------------------------------------------------------------------------
 
+import importlib
 import os
+import sys
+
+QtAPIs = [
+    ('pyside2', 'PySide2'),
+    ('pyside', 'PySide'),
+    ('pyqt5', 'PyQt5'),
+    ('pyqt', 'PyQt4'),
+]
 
 def prepare_pyqt4():
     # Set PySide compatible APIs.
@@ -22,23 +31,33 @@ def prepare_pyqt4():
     sip.setapi('QUrl', 2)
     sip.setapi('QVariant', 2)
 
+qt_api = None
+
+# have we already imported a Qt API?
+for api_name, module in QtAPIs:
+    if module in sys.modules:
+        qt_api = api_name
+        break
+
+# does our environment give us a preferred API?
 qt_api = os.environ.get('QT_API')
+if qt_api not in {api_name for api_name, module in QtAPIs}:
+    msg = ("Invalid Qt API %r, valid values are: " +
+           "'pyside, 'pyside2, 'pyqt' or 'pyqt5'") % qt_api
+    raise RuntimeError(msg)
 
+# if we have no preference, is a Qt API available? Or fail with ImportError.
 if qt_api is None:
-    try:
-        import PySide
-        qt_api = 'pyside'
-    except ImportError:
+    for api_name, module in QtAPIs:
         try:
-            prepare_pyqt4()
-            import PyQt4
-            qt_api = 'pyqt'
+            importlib.import_module(module)
+            qt_api = api_name
+            break
         except ImportError:
-            raise ImportError('Cannot import PySide or PyQt4')
+            continue
+        else:
+            raise ImportError('Cannot import PySide2, PySide, PyQt5 or PyQt4')
 
-elif qt_api == 'pyqt':
+if qt_api == 'pyqt':
+    # set the PyQt4 APIs
     prepare_pyqt4()
-
-elif qt_api != 'pyside':
-    raise RuntimeError("Invalid Qt API %r, valid values are: 'pyqt' or 'pyside'"
-                       % qt_api)
--- a/pyface/qt/__init__.py
+++ b/pyface/qt/__init__.py
@@ -41,10 +41,6 @@ for api_name, module in QtAPIs:
 
 # does our environment give us a preferred API?
 qt_api = os.environ.get('QT_API')
-if qt_api not in {api_name for api_name, module in QtAPIs}:
-    msg = ("Invalid Qt API %r, valid values are: " +
-           "'pyside, 'pyside2, 'pyqt' or 'pyqt5'") % qt_api
-    raise RuntimeError(msg)
 
 # if we have no preference, is a Qt API available? Or fail with ImportError.
 if qt_api is None:
@@ -57,6 +53,13 @@ if qt_api is None:
             continue
         else:
             raise ImportError('Cannot import PySide2, PySide, PyQt5 or PyQt4')
+
+# otherwise check QT_API value is valid
+elif qt_api not in {api_name for api_name, module in QtAPIs}:
+    msg = ("Invalid Qt API %r, valid values are: " +
+           "'pyside, 'pyside2, 'pyqt' or 'pyqt5'") % qt_api
+    raise RuntimeError(msg)
+
 
 if qt_api == 'pyqt':
     # set the PyQt4 APIs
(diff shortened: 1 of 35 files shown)


pankajp:
For more api changes you can take inspiration from https://github.com/rwarren/qt_backport https://github.com/rwarren/qt_backport/blob/master/qt_backport/qt5_backport.py

pankajp:
Thanks for PR @corranwebster . I really think this approach will involve less code and help in gradual move to Qt5.
Btw, i could get `pyqt5` tests working on python3 in travis-ci since it now has wheels for linux via the `manylinux1` platform tag, except that those wheels do not bind QtWebkit.
https://travis-ci.org/enthought/pyface/jobs/209678577

corranwebster:
Qt5 tests are running, but failing with segfaults as noted by @pankajp. Suspicion should fall on the modal dialog testing code first.

pyface/toolkit.py
```diff
@@ -80,18 +80,23 @@ def import_toolkit(tk):
             modules = ', '.join(plugin.module_name for plugin in plugins)
             logger.warning(msg, tk, modules)
 
+        exception = None
         while plugins:
             plugin = plugins.pop(0)
             try:
                 tk_object = plugin.load()
                 return tk_object
             except ImportError as exc:
```
rkern: I think you could just change this to `except ImportError as exception:`

rkern: I think you could just change this line to `except ImportError as exception:`

rkern:
I'm somewhat in favor for moving to something like [`QtPy`](https://pypi.python.org/pypi/QtPy) to move the Qt4-style organization of classes to Qt5-style in the longer term. But I do recognize that it causes significant short-term work to update all of the existing code.

pankajp:
Just for the kicks: https://github.com/enthought/pyface/commits/feature/qtpy & https://github.com/enthought/traitsui/commits/feature/qtpy

corranwebster:
I think a reasonable proposal (and what I suspect that @pankajp is working towards) is:

* keep `pyface.qt` as a Qt4 compatible API so that we don't break all of our existing code that depends on the current API.
* incrementally move libraries to using a Qt5 compatible API imported from some other location (whether it be QtPy or we roll our own).

I think using QtPy is probably a good idea, but it may make sense to have a shim in Pyface (`pyface.qt5`?) so we can easily replace it with something else if there is a better alternative or QtPy goes bad in the future.

corranwebster:
At this point the segfault is happening on Travis CI, but not the Ubuntu VM I've been using for testing, so it is getting hard to diagnose.  I will branch here and play around to see what works.

pankajp:
@corranwebster I'm happy with the approach of this PR (obviously).

Once it is reviewed I think we should be able to get it merged without blocking on the segfaulting modal dialog tests imho because:

- This PR is backward compatible with no changes for existing pyside/pyqt4 codebase so no existing users would be impacted
- The segfault only happens on travisci, and not on both my and your linux machines
- This would give a nice starting point for users to begin trying out PyQt5 without waiting for everything to be perfect

In a subsequent PR we can add something like `pyface.qt5` as you mentioned, so that people can start migrating to qt5 modular api without having to wait on all exiting code being migrated.

corranwebster:
@pankajp I've merged in the test fixes branch, making the Qt5 tests "allowed failures" for now, since we still get occasional segfaults which can't be locally reproduced.

Assuming CI runs successfully, do you think this is good to merge into master?

pyface/tests/test_about_dialog.py
```diff
@@ -8,16 +8,26 @@
...
 no_modal_dialog_tester = (ModalDialogTester.__name__ == 'Unimplemented')
 
 
-class TestAboutDialog(unittest.TestCase):
+@unittest.skipIf(no_gui_test_assistant, 'No GuiTestAssistant')
+class TestAboutDialog(unittest.TestCase, GuiTestAssistant):
 
     def setUp(self):
-        self.gui = GUI()
+        GuiTestAssistant.setUp(self)
```
pankajp: Can `super` be used if `GUITestAssistant` was the first base class? I think mixin classes in Python have to be first base class to play nice with inheritance and super

tasks.py
```diff
@@ -89,13 +90,17 @@
 extra_dependencies = {
     'pyside': {'pyside'},
     'pyqt': {'pyqt'},
-    'wx': {'wxpython'},
+    # XXX once pyqt5 is available in EDM, we will want it here
+    'pyqt5': set(),
+    # XXX temporary workaround for bug in recent EDM wxpython build on OS X
+    'wx': {'wxpython' if sys.platform != 'darwin' else 'wxpython==3.0.2.0-3'},
```
pankajp: I think this is fixed with latest wxpython

pyface/tests/test_about_dialog.py
corranwebster: I prefer to do explicit call here, rather than super, because I found it surprising that `TestCase` _didn't_ implement a trivial `setUp` and `tearDown`, so the explicit call flags the future readers of the code that something not usual is happening here.

tasks.py
corranwebster: No, this is an issue with the EPD/EDM build of wxpython-3.0.2.0-4 OS X which is still the current version being served.  We can remove this if/when there is a 3.0.2.0-5 or higher build for OS X.

pyface/qt/__init__.py
```diff
@@ -22,23 +31,36 @@ def prepare_pyqt4():
...
+qt_api = None
+
+# have we already imported a Qt API?
+for api_name, module in QtAPIs:
+    if module in sys.modules:
+        qt_api = api_name
+        break
+
+# does our environment give us a preferred API?
 qt_api = os.environ.get('QT_API')
```
corranwebster: Good catch.

corranwebster:
And yes, tests still segfault eg. https://travis-ci.org/enthought/pyface/jobs/214604026

I think there may be a problem with the `ModalDialogTester` which is particularly apparent on Qt5, but @itziakos and I went through it fairly carefully a couple of weeks ago and couldn't find any obvious problems with it.  I think the situation we are in is:

* tests of `QMessage` based dialogs reliably segfault under Qt5 with the `ModalDialogTester`
* other tests of modal dialogs _occasionally_ segfault under Qt5.

But I think that the problem is not with Qt5, but more with the testing infrastructure.

pankajp:
Ok. LGTM. I'd have preferred letting pyside2 be if it doesn't harm, just in case someone wishes to try (pyside2 unofficial wheels are available on bintray). pyface could be yet another package pyside2 devs can test against

corranwebster:
The problem with Pyside2 was that it was first on the list and was not supported, so if it was installed it would be preferentially chosen and then fail (actually, it would try to use original Pyside).

I also see we are back to getting occasional PyQt4 segfaults, but that is a known issue.

pankajp:
Ok, let it be gone then. What about pyqt4 segfaults, i am not aware. Any pointers to issue/build failures?
And I think this PR is good to go, a great start to Qt5 support.

corranwebster:
Yes, various Qt segfaults have been collected here: #251 A lot of them cleared up when we removed old-style signals and slots; I think we had situations where old-style and new style were being used together.

I will merge this in a little bit.

thomasaarholt:
This is great! Over at hyperspy/hyperspy#1323 we're very excited for this, and it seems to be working well.

PR #265, opened 2017-03-15
Fix tests on Qt5

corranwebster:
Breaking out into separate PR to reduce noise.  We are at point of experimentation for what works and doesn't in Travis CI environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,20 @@ language: generic
 sudo: false
 matrix:
   include:
+    - env: RUNTIME=2.7 TOOLKIT=pyqt
+    - env: RUNTIME=2.7 TOOLKIT=pyside
+    - env: RUNTIME=2.7 TOOLKIT=wx
+    - env: RUNTIME=3.5 TOOLKIT=pyqt
+    - env: RUNTIME=3.5 TOOLKIT=pyqt5
+    - os: osx
+      env: RUNTIME=2.7 TOOLKIT=pyqt
+    - os: osx
+      env: RUNTIME=2.7 TOOLKIT=pyside
+    - os: osx
+      env: RUNTIME=2.7 TOOLKIT=wx
+    - os: osx
+      env: RUNTIME=3.5 TOOLKIT=pyqt
+  allow_failures:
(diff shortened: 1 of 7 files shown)


pankajp:
Copying comment from previous PR in case you missed:

> The tests skipped due to segfaults do not segfault on my machine's distro supplied pyqt5, but only on the wheels on pypi, i suspect some compilation issue or so, though one segfault in traitsui table_editor was definitely buggy code: enthought/traitsui@ae05ee9

Travis's docker containers are available for testing locally https://docs.travis-ci.com/user/common-build-problems/#Troubleshooting-Locally-in-a-Docker-Image

Also, in a previous attempt to debug, valgrind found a jump depending on unintialized value in some font codepath when the segfaults occured, and the segfault happens in `QPushButton.text()` method when the modal dialog tester attempts to click a button in the dialog as reported by faulthandler, found by ipython tracing and `python3 -X faulthandler`

corranwebster:
I had managed to replicate a codepath via which I was getting segfaults in the `text()` method on my VM, but the careful use of `GuiTestAssistant` and explicit tear down of created components seems to have solved that on my VM, but not on Travis CI 😞 

I'm aware of the ability to run docker images locally, but I don't have docker and associated infrastructure set up on my machine, and so it's more efficient time-wise just to work in a branch than to go through setting up the infrastructure.

corranwebster:
Based on above, and some judicious re-running of tests, I think there are two issues:

- `click_widget` is suspect under PyQt5 when used with the confirmation dialog (possibly because of interactions with underlying UI dialog).  We get a repeatable segfault every time in these situations on Travis CI.
- `open_and_wait` is suspect under Qt5 (and possibly Qt4 as well).  Errors are replicable, but not on every run.  This would be consistent with some sort of race condition, accessing something from the wrong thread, or any of the usual issues that you can get from concurrency.

It is possible that the `click_widget` problem is related to the concurrency issues, but it is also possibly just a bug in the Qt or PyQt versions we get on Travis CI.

Proposal: for now, with Qt5 we just disable the `ModalDialogTester` until we can get it working reliably on Qt5.  The GUI test assistants could likely do with an overhaul in any case.

PR #266, opened 2017-03-15
Wx Test Assistants

corranwebster:
These are Wx versions of the GuiTestAssistant and other related code.  This should not be merged until tests run cleanly.  We expect significant breakage since this is running tests that would otherwise be impossible on Wx based code.  They may also need work to make the tools better, safer and more cross-platform.

### DIFF
--- a/pyface/ui/wx/util/event_loop_helper.py
+++ b/pyface/ui/wx/util/event_loop_helper.py
@@ -1,0 +1,151 @@
+
+# (C) Copyright 2014-15 Enthought, Inc., Austin, TX
+# All right reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in enthought/LICENSE.txt and may be redistributed only
+# under the conditions described in the aforementioned license.  The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Code modified from pyface.ui.qt4.util.event_loop_helper
+import contextlib
+import threading
+
+import wx
+from traits.api import HasStrictTraits, Instance
+from pyface.api import GUI
+from pyface.util.guisupport import start_event_loop_wx
+
+
+class ConditionTimeoutError(RuntimeError):
+    pass
+
+
+@contextlib.contextmanager
+def dont_quit_when_last_window_closed(wx_app):
+    """
+    Suppress exit of the application when the last window is closed.
+
+    """
+    flag = wx_app.GetExitOnFrameDelete()
+    wx_app.SetExitOnFrameDelete(False)
+    try:
+        yield
+    finally:
+        wx_app.SetExitOnFrameDelete(flag)
+
+
+class EventLoopHelper(HasStrictTraits):
+    """ A helper class to provide basic event loop functionality. """
+
+    #: The gui toolkit application.
+    gui_app = Instance(wx.App)
+
+    #: The pyface gui implementation
+    gui = Instance(GUI, ())
+
+    #: Reference to a dummy wx event handler that will help bind events for
+    #: the timers.
+    _handler = Instance(wx.EvtHandler, ())
+
+    @contextlib.contextmanager
+    def event_loop(self, repeat=1):
+        """ Emulates an event loop `repeat` times.
+
+        Parameters
+        ----------
+        repeat : int
+            Number of times to process events. Default is 1.
+
+        """
+        yield
+        for i in range(repeat):
+            self.gui_app.ProcessPendingEvents()
+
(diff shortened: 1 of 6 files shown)


pyface/ui/wx/util/event_loop_helper.py
```diff
@@ -0,0 +1,151 @@
...
+    def event_loop(self, repeat=1):
+        """ Emulates an event loop `repeat` times.
+
+        Parameters
+        ----------
+        repeat : int
+            Number of times to process events. Default is 1.
+
+        """
+        yield
```
kitchoi: Should this be in a try-finally block?  If there is an exception from within the context, we would still want to process the fired events right?

pyface/ui/wx/util/gui_test_assistant.py
```diff
@@ -0,0 +1,272 @@
...
+                window.Destroy()
+                wx.WakeUpIdle()
+
+        wx.CallLater(50, _cleanup)
+        self.gui_app.MainLoop()
+        if len(titles) > 0:
+            raise RuntimeError(
+                'Not all windows closed after test method, {}'.format(titles))
+        del self.event_loop_helper
+        del self.gui_app
```
kitchoi: Are these about removing references to the GUI objects? Should `del` be called before raising RuntimeError?  I am not sure if these references are kept alive when multiple test cases are being run.

itziakos: yes we should call del before raising

pyface/ui/wx/util/gui_test_assistant.py
```diff
@@ -0,0 +1,272 @@
...
+            for window in windows:
+                titles.append(window.GetTitle())
+
+        def _cleanup():
+            for window in wx.GetTopLevelWindows():
+                window.Destroy()
+                wx.WakeUpIdle()
+
+        wx.CallLater(50, _cleanup)
+        self.gui_app.MainLoop()
```
kitchoi: Could we do `self.event_loop_helper.event_loop_until_condition(<callable-that-returns-true-when-all-windows-are-closed>)` instead?  There is a danger of this MainLoop call hanging there forever.

pyface/ui/wx/util/event_loop_helper.py
itziakos: Good point, I think a try-finally might be a good thing to do

pyface/ui/wx/util/gui_test_assistant.py
itziakos: The wx event loop exists only when all the windows are closed. Event_loop_until_codition does the same so there is no advantage using it.

itziakos: oops I was wrong

itziakos: So the right thing here is to use a wx.Timer 

something like
```
handler = wx.EvtHandler()
kill_timer = wx.Timer(handler) 
handler.Bind(wx.EVT_TIMER, lambda event:self.gui_app.Exit(), kill_timer) 
kill_timer.Start(30000), True) 
try:
self.gui_app.MainLoop()
finally:
      kill_timer.Stop()
```

pyface/ui/wx/util/event_loop_helper.py
```diff
@@ -0,0 +1,151 @@
...
+        timer = wx.Timer(self._handler)
+        self._handler.Bind(
+            wx.EVT_TIMER, lambda event:self.gui_app.Exit(), timer)
+        yield
+        timer.Start(int(timeout * 1000), True)
+        self._start_event_loop()
+        if not timer.IsRunning():
+            # We exited the event loop on timeout
+            raise ConditionTimeoutError(
+                'Could not destroy widget before timeout: {!r}'.format(widget))
```
kitchoi: `widget` is not defined here.  `window` instead?

itziakos: That is probably true

pyface/ui/wx/util/gui_test_assistant.py
kitchoi: Does the timer have the same effect as `wx.CallLater(30000, self.gui_app.Exit)`?

itziakos: It is not the same. Since with the time we can deactivate it while the CallLater will be called later probably while on another is running.

kitchoi: Ah!  Good point!

PR #269, opened 2017-04-07
Fixes for a few small issues when using wx

robmcmullen:
Fixed a few minor bugs when using wx

### DIFF
--- a/pyface/toolkit.py
+++ b/pyface/toolkit.py
@@ -108,7 +108,7 @@ def _init_toolkit():
                 return import_toolkit(tk)
         except RuntimeError as exc:
             exc_info = logger.getEffectiveLevel() <= logging.INFO
-            level = 'ERROR' if exc_info else 'INFO'
+            level = logger.ERROR if exc_info else logging.INFO
             msg = "Could not import Pyface backend %r"
             logger.log(level, msg, tk, excinfo=exc_info)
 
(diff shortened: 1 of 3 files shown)


corranwebster:
Thanks. Looks good.

pyface/toolkit.py
```diff
@@ -108,7 +108,7 @@ def import_toolkit(tk):
                 return import_toolkit(tk)
         except RuntimeError as exc:
             exc_info = logger.getEffectiveLevel() <= logging.INFO
-            level = 'ERROR' if exc_info else 'INFO'
+            level = logger.ERROR if exc_info else logging.INFO
```
rkern: `logging.ERROR`

pyface/ui/wx/action/action_item.py
```diff
@@ -74,7 +74,7 @@ def __init__(self, parent, menu, item, controller):
         action  = item.action
         label   = action.name
         kind    = _STYLE_TO_KIND_MAP[action.style]
-        longtip = action.description
+        longtip = action.description or action.tooltip
```
rkern: This is a change in semantics. If we want to do this, then the same logic should be applied on the `qt4` side as well.

robmcmullen: This is the comment I was basing the patch on:

https://github.com/enthought/pyface/blob/master/pyface/action/action.py#L45
```
    #: A longer description of the action (used for context sensitive help etc).
    #: If no description is specified, the tooltip is used instead (and if there
    #: is no tooltip, then well, maybe you just hate your users ;^).
    description = Unicode
```

rkern: Sure, that's fine; it just needs to be propagated to the `qt4` version of this code as well.

robmcmullen: Oh, having never looked at the Qt side (because Qt is a total mystery) I thought wx was just not quite in sync.

PR #271, opened 2017-05-25
Update QtOpenGL.py

nprksh:
Use elif instead of a new if, otherwise will throw error --> PyQt4 will never be used


corranwebster:
LGTM.  Thanks.

PR #274, opened 2017-06-06
Fixed unnecessary full re-layout of AUI when only changing a pane's name

robmcmullen:
Simple fix that prevents a time-consuming re-layout of the entire AUI manager. The pane name can be changed with a simple refresh of the pane. Removes flicker and is sooooo much faster.


robmcmullen:
Found a problem. Will submit a new one.

PR #275, opened 2017-06-06
Fixed unnecessary full re-layout of AUI when only changing a pane's name

robmcmullen:
OK, for realz this time. The simple fix of the last pull request has issues on different platforms, so it require a little more code. It still prevents a time-consuming re-layout of the entire AUI manager, although it does refresh everything. Still much faster than a layout, though.


corranwebster:
LGTM

PR #279, opened 2017-07-06
Add missing QtWebkit import for PyQt5 compatibility

sjagoe:
Add missing import for PyQt5 compatibility.

### DIFF
--- a/pyface/qt/QtWebKit.py
+++ b/pyface/qt/QtWebKit.py
@@ -4,6 +4,7 @@ if qt_api == 'pyqt':
     from PyQt4.QtWebKit import *
 
 elif qt_api == 'pyqt5':
+    from PyQt5.QtWebKit import *
     from PyQt5.QtWidgets import *
     from PyQt5.QtWebKitWidgets import *
 


corranwebster:
Qt test failures look like not your fault. (We are still getting segfaults in the code that inspects the modal dialogs, it seems).

corranwebster:
Change looks reasonable.  Feel free to merge. Do we have a PyQt5 egg in EDM that we can use to test yet?

sjagoe:
> Do we have a PyQt5 egg in EDM that we can use to test yet?

An internal experimental build only that I'm using for this; it's also not yet available for all platforms.

sjagoe:
> `Failure: RuntimeError (No plugin could be loaded for qt4.) ... ERROR`

These error's don't look good. That's new in this PR, although I'm not quite sure how. Our experimental Qt5 builds are Qt 5.6, I'll see if there's something there.

sjagoe:
I don't see those errors running locally on Ubuntu 15.10, using otherwise the same setup as travis-ci

sjagoe:
I did a rebuild of a master build, and the same errors occur there: https://travis-ci.org/enthought/pyface/jobs/244937331

@corranwebster I propose we merge this (it's a trivial change, and critical to actually being able to _use_ `QtWebKit` with `Qt5`), and separately we can look into the failures. Probably when we have a build of PyQt5 available in the EDM repositories, which should be Soon.

corranwebster:
@sjagoe I am Ok with this being merged, as I don't think test failures have anything to do with this.

PR #281, opened 2017-08-22
Be more careful about object cleanup when tests finish

corranwebster:
This is an attempt to fix the issues with excess objects after tests have finished as identified by #254.  There is some hope that these may address the segfault issues we are seeing under PyQt.

### DIFF
--- a/pyface/tests/test_about_dialog.py
+++ b/pyface/tests/test_about_dialog.py
@@ -55,14 +55,6 @@ class TestAboutDialog(unittest.TestCase, GuiTestAssistant):
             parent.destroy()
         self.gui.process_events()
 
-    def test_create_ok_renamed(self):
-        # test that creation and destruction works as expected with ok_label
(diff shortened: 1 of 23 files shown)


corranwebster:
@itziakos Again, no rush on this, but if you could give this a quick sanity check before I merge.

pyface/tests/test_about_dialog.py
```diff
@@ -26,17 +26,20 @@ def tearDown(self):
...
                 self.dialog.destroy()
+        self.dialog = None
         GuiTestAssistant.tearDown(self)
 
     def test_create(self):
         # test that creation and destruction works as expected
         self.dialog._create()
         self.gui.process_events()
         self.dialog.destroy()
+        self.gui.process_events()
```
itziakos: if a gui_test_assistant exists the correct thing is to use `self.event_loop()` or `with self.event_loop()`

PR #284, opened 2017-08-25
Add visible and enabled traits to base Widget interface

corranwebster:
Apart from general utility, this will permit better integration with TraitsUI `visible_when` and `enabled_when` (the long-term aim here is to eventually generalise the TraitsUI `View` buttons with Pyface objects).

### DIFF
--- a/pyface/i_widget.py
+++ b/pyface/i_widget.py
@@ -15,7 +15,7 @@
 
 
 # Enthought library imports.
-from traits.api import Any, Interface
+from traits.api import Any, Bool, HasTraits, Interface
 
 
 class IWidget(Interface):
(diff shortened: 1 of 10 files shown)


pyface/i_widget.py
```diff
@@ -24,24 +24,46 @@ class IWidget(Interface):
...
     parent = Any
 
-    ###########################################################################
+    #: Whether or not the control is visible
+    visible = Bool(True)
+
+    #: Whether or not the control is enabled
+    enabled = Bool(True)
+
+    # ------------------------------------------------------------------------
```
rkern: Really, I'd just turn off that `flake8` warning instead. :-)

PR #287, opened 2017-11-03
example conversion to wxPython 4

robmcmullen:
Merging this would break compatibility with wxPython 3, so I was wondering what the upgrade strategy might be?

I was thinking about trying to add a new toolkit called "wx4" where the normal "wx" would then refer to wxPython 3, but there is stuff that doesn't lie behind the pyface/wx and pyface/ui/wx, like pyface/dock/*, pyface/sizers/flow.py, and pyface/tree/node_tree.py. I'm not sure how to deal with that.

Note: a lot of this work was done by Eric Carlson here: https://github.com/braidedlogix/ETS_py3 but that code had some python 3 specific stuff as well as a bunch of whitespace/styling changes that I left out in this.

There are also some changes needed in the traitsui package, so if you have a different way you'd like me to show those changes, let me know, otherwise I can submit a "pull request" that I'm not intending for you to actually pull.

### DIFF
--- a/pyface/__init__.py
+++ b/pyface/__init__.py
@@ -23,7 +23,7 @@ except ImportError:
 
 __requires__ = ['traits']
 __extras_require__ = {
-    'wx': ['wxpython>=2.8.10', 'numpy'],
+    'wx': ['wxpython>=4', 'numpy'],
     'pyqt': ['pyqt>=4.10', 'pygments'],
     'pyqt5': ['pyqt>=5', 'pygments'],
     'pyside': ['pyside>=1.2', 'pygments'],
--- a/pyface/dock/dock_sizer.py
+++ b/pyface/dock/dock_sizer.py
@@ -193,7 +193,7 @@ def set_standard_font ( dc ):
     global standard_font
 
     if standard_font is None:
-        standard_font = wx.SystemSettings_GetFont( wx.SYS_DEFAULT_GUI_FONT )
+        standard_font = wx.SystemSettings.GetFont(wx.SYS_DEFAULT_GUI_FONT)
 
     dc.SetFont( standard_font )
 
@@ -207,9 +207,10 @@ def clear_window ( window ):
     """ Clears a window to the standard background color.
     """
     bg_color = SystemMetrics().dialog_background_color
-    bg_color = wx.Colour(bg_color[0]*255, bg_color[1]*255, bg_color[2]*255)
+    bg_color = wx.Colour(
+        int(bg_color[0] * 255), int(bg_color[1] * 255), int(bg_color[2] * 255))
 
-    dx, dy = window.GetSizeTuple()
+    dx, dy = window.GetSize().Get()
     dc     = wx.PaintDC( window )
     dc.SetBrush( wx.Brush( bg_color, wx.SOLID ) )
     dc.SetPen( wx.TRANSPARENT_PEN )
@@ -224,14 +225,14 @@ def get_dc ( window ):
     """
     if is_mac:
         dc     = wx.ClientDC( window )
-        x, y   = window.GetPositionTuple()
-        dx, dy = window.GetSizeTuple()
+        x, y = window.GetPosition().Get()
+        dx, dy = window.GetSize().Get()
         while True:
             window = window.GetParent()
             if window is None:
                 break
-            xw, yw   = window.GetPositionTuple()
-            dxw, dyw = window.GetSizeTuple()
+            xw, yw = window.GetPosition().Get()
+            dxw, dyw = window.GetSize().Get()
             dx, dy   = min( dx, dxw - x ), min( dy, dyw - y )
             x += xw
             y += yw
@@ -240,7 +241,7 @@ def get_dc ( window ):
 
         return ( dc, 0, 0 )
 
-    x, y = window.ClientToScreenXY( 0, 0 )
+    x, y = window.ClientToScreen(0, 0)
     return ( wx.ScreenDC(), x, y )
 
 #-------------------------------------------------------------------------------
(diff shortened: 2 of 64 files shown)


corranwebster:
@robmcmullen and @braidedlogix thanks for this!

I'm really unsure about how best to handle the transition.

It would be awesome if we could have some way of getting it to work with at least WxPython3 and WxPython4 in the same codebase, as that would allow a staggered introduction of the code (ie. we could get Pyface and TraitsUI stable and then worry about projects like Mayavi, Apptools and Enable later).

One possible approach would be to simply do an if-then on the WxPython version whenever we have an instance of an incompatible API, which is probably a fair bit of work, but it doesn't look to be totally insane.  It might be possible to include something like what we do in pyface.qt (or what QtPy does) where we have a package that maps WxPython submodules to a common API where possible (eg. we have a `pyface.wx4` tree which provides the common API depending on which version we are using; and replace a lot of `import wx` statements with `import pyface.wx4 as wx`).

Having a separate `wx4` toolkit probably doesn't help much: it would mean essentially duplicated code in `pyface.ui.wx` and `pyface.ui.wx4` which we would then need to keep synchronized; plus as you note, we'd still need to deal with all the wx code that isn't in `pyface.ui.wx.*`. This is a particular problem for TraitsUI which relies on the code in `pyface.dock` in a fairly heavy way.

Or it may just be best to make the switch to WxPython4 cleanly at a major version release and not worry about backwards compatibility, but if we do that then we need to have WxPython4 branches ready to go for downstream projects as well.

Since we aren't using Wx heavily at Enthought right now, I'm more than willing to follow community advice from those who _are_ using it about what would be the best way forward.

robmcmullen:
I'll take a look at how the qt4/qt5 stuff is set up and start to think about how that might work for wx3/wx4.

Another possibility is that, since we're about 2 years away from the python 2 end-of-life and wx3 doesn't support python 3: do you have a plan on when to drop python 2 support? That would probably be a good time to go cold turkey if so. My projects will certainly outlast that switchover so I'd be able to keep the wx4 branch up-to-date and ready to go at that time.

prabhuramachandran:
I vote that we just go with wx4 and slowly port the rest.  Mayavi is not an issue and compatible already, I think traitsui is a big one as is envisage and chaco.  If we can get pyface, traitsui and chaco going that is a great start I'd say.

prabhuramachandran:
@robmcmullen -- what are your thoughts on this?

robmcmullen:
I've already ported my stuff to this pyface/wx4 branch, so I'd of course say "yes definitely change to wx4!" but I don't know how many folks are still using wx3 with pyface. My guess is that wx4 will increase its market share because it is now pip installable. wx3 wasn't, which was a barrier to entry.

If Enthought organization isn't using wx much and the wx3 and wx4 ports can't coexist easily, can the next major version bump replace wx3 with wx4?

braidedlogix:
It seems a no-brainer to me to just move on to wx4. That said, going from 2
to 3 was considerably more complicated than going from 3 to 4, other than
vast numbers of segfaults that may have been patched up since last year.
Even with 2 to 3, much of that involved event binding that could work in
both 2 and 3. There are a few incompatibilities that affect other packages
(TreeListControl, for example) but are not used by ETS. In fact, I saw much
more need for special cases going from qt4 to qt5, and even between
branches of qt5, than I found for wx (once binding was updated). Much of
the special case branching for wx2 is there to accommodate versions that
date back to Win97, and I have to wonder where anyone would possibly find
these older versions.

prabhuramachandran:
I agree about the compatibility with wx2 and suspect we can drop it but am not sure if Enthought needs it, I think not.  My issue is with wxPython 3 -- as you point out the changes from 3 to 4 are minor in comparison. I am personally happy with moving to 4.  It is just that Debian/Ubuntu and the others will still be shipping wx 3.x (16.04 for example ships 3.0.2). As you point out, wx4 has wheels although not on Linux yet.  So I would really like things to work with wx3 as well just to avoid pain on Linux.  I suspect it is possible to have both 3 and 4 working with the current changes?  Thoughts?  Also @corranwebster -- what are your thoughts?

robmcmullen:
The changes from wxPython 3 may be minor, but there's lots of them and it would take lots of effort. Looking at the changes again, without adding a new toolkit API flag like "wx4" vs "wx", it would be a super-pain. As for me, I'm abandoning wx3. wx4 builds fine on linux with the dependencies, and the deps are available on 16.04, at least.

corranwebster:
How's this for a possible way forward which doesn't totally strand anyone still on Wx3: we have a release really soon now (which has been waiting 6 months for me to have the time, but there is the possibility in the next month) which keeps the Wx code as it currently is. Immediately after that we can merge Wx4 changes without worrying too much about Wx3 backwards compatibility.  If it can be done easily, great, otherwise the release after that will drop Wx3 support.

prabhuramachandran:
@corranwebster -- I think your plan sounds great, but I think you may need to merge #294 as it fixes a bug.  @robmcmullen -- I guess if we go with Corran's plan you may need to update this PR to fix any conflicts and updates, I imagine you would be fine with that assuming the release happens soon?

robmcmullen:
@prabhuramachandran I'm invested enough that I'll make sure the wx4 stuff is working when we do get to the big merge time; no worries.

corranwebster:
@robmcmullen @braidedlogix @prabhuramachandran Now that 6.0 is out, you should feel free to put forward PRs for Pyface and TraitsUI that provide WxPython 4 support.  If you can do it in a backwards compatible way with WxPython 3.x, awesome, but if not then that's OK.

robmcmullen:
I'll check out the 6.0 release and start looking at merging these changes. wx3 is dead to me, so my work will be wx4 only. :grin:

lfriedri:
As a user of python3 + wxpython4 I would like to vote for wxpython4 support. Currently I am cut off from chaco and mayavi, which is a pity.

prabhuramachandran:
@robmcmullen -- any update on this?  Thanks!

corranwebster:
Everything in this has now been merged as part of #489.

Thanks all!

PR #288, opened 2017-11-13
changes for wxpython update

reckoner:
small tweaks for changes in wxpython.


corranwebster:
Hi, thanks for the contribution!

Unfortunately, I don't think we can break backwards compatibility with WxPython 3 just yet.  See #287 for a similar PR and some discussion about the way we might go forward with this.  You might talk with @robmcmullen to see if you can pool your resources on this.

corranwebster:
This should be resolved by #489.  Thanks for the contribution!

PR #290, opened 2017-12-14
Fix: Ordering of items being appended to the default group of ActionManager

### LINKED ISSUE #289: MenuManager does not respect order of appending Action
I modified the example script in `example/progress_dialog.py` to use `MenuManager.append` for adding `Action`

```
from __future__ import print_function

import time
from pyface.api import GUI, ApplicationWindow, ProgressDialog
from pyface.action.api import Action, MenuManager, MenuBarManager

def _main():
    print("main")

    """ The main application window. """

    ######################################################################
    # 'object' interface.
    ######################################################################

    def __init__(self, **traits):
        """ Creates a new application window. """

        # Base class constructor.
        super(MainWindow, self).__init__(**traits)

        # Add a menu bar.
        menu = MenuManager(name="&File")
        menu.append(Action(name='Close', on_perform=self.close))        
        menu.append(Action(name='DoItAgain', on_perform=_main))
        menu.append(Action(name='DoIt', on_perform=_main))
        menu.append(Action(name='Quit', on_perform=self.close))        
        self.menu_bar_manager = MenuBarManager(menu)


if __name__ == "__main__":
    gui = GUI()
    # Create and open the main window.
    window = MainWindow()
    window.open()
    _main()
    # Start the GUI event loop!
    gui.start_event_loop()
```

You would expect the Action to follow this order:
Close, DoItAgain, DoIt, Quit

This is what I get on Windows 7 64-bit (with PySide 1.2.2-14, PyQt 4.12-1 and wxPython 3.0.2.0-5):
menu_manager_order

And on Mac OSX 10.11 (with PySide 1.2.2-14):
screen shot 2017-12-13 at 15 50 03
Note that on Mac OSX the "Quit" Action is not even there.

On Mac OSX 10.11 (with wxPython 3.0.2.0-5):
screen shot 2017-12-13 at 16 01 20
Quit is there with wxPython

kitchoi:
Fix #289 

When item that is not a `Group` is being added to `ActionManager` using `append`, the index is computed from the number of groups (by default there is one Group) rather than the number of items in the default group.  So items were being inserted before index 1, hence the order was wrong.

This PR fixes this.

(Note: `Group` does not provide `__len__`, it would be nice if it did.)

### DIFF
--- a/pyface/action/action_manager.py
+++ b/pyface/action/action_manager.py
@@ -332,7 +332,7 @@ class ActionManager(HasTraits):
 
         Parameters
         ----------
-        item : Group instance or ActionManagerItem instance
+        item : string, Group instance or ActionManagerItem instance
             The item to be added to this ActionManager
 
         Returns
--- a/pyface/action/action_manager.py
+++ b/pyface/action/action_manager.py
@@ -162,8 +162,9 @@ class ActionManager(HasTraits):
         of groups.  If the item is an ActionManagerItem then the item is
         appended to the manager's defualt group.
         """
-
-        return self.insert(len(self._groups), item)
+        group = self._get_target_group(item)
+        group.append(self._prepare_item(item))
+        return group
 
     def destroy(self):
         """ Called when the manager is no longer required.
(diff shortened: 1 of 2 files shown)


corranwebster:
LGTM

PR #294, opened 2018-02-07
BUG: Fixing use of `control.bind`.

prabhuramachandran:
Running Mayavi with the latest wxPython or older versions like here
fails: https://travis-ci.org/enthought/mayavi/jobs/337913143
with wxPython 2.8.x.  Looks like it should be `control.Bind`.


corranwebster:
Failures here look to be EDM related.  Probably CI is downloading a too-old version of EDM.

prabhuramachandran:
OK, but looks like I needn't have done this as #287 does this quite nicely.

PR #298, opened 2018-02-15
Click- and EDM-based Test Runners

corranwebster:
Add click-based CI for Travis, based on TraitsUI version.  Also will convert the Appveyor system to use etstool.py.


corranwebster:
@itziakos do we need to merge to master for the new `appveyor.yml` to take effect or something?

itziakos:
It should not matter.

PR #303, opened 2018-02-19
Basic PySide2 support

corranwebster:
Add PySide2 support to the Qt backend.


corranwebster:
Tests fail with segfault from deallocating `None`!  Fun!

corranwebster:
I'm going to merge this and open a ticket for the test failures.  This means that people can experiment with PySide2, which will help with feedback.

PR #304, opened 2018-02-20
Add 'qt' as a synonym for 'qt4' for ETS_TOOLKIT

corranwebster:
This is the first step in allowing setting `ETS_TOOLKIT=qt` instead of `qt4`.  Currently wont work because TraitsUI will break

Also generalize `Toolkit` for use in other packages (eg. TraitsUI, Enable, etc.), as this will be useful in rolling out support for `pkg_resource`-based toolkit discovery.

### DIFF
--- a/pyface/base_toolkit.py
+++ b/pyface/base_toolkit.py
@@ -20,15 +20,22 @@ class Toolkit(HasTraits):
     class that raises NotImplementedError when it is instantiated.
     """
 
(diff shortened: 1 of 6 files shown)


corranwebster:
Going to self-merge this; code is well tested, API not yet in a release, so no breakage; other than that only adds an extra entrypoint.

PR #305, opened 2018-02-21
Common Toolkit Selection Code

corranwebster:
Refactor the logic for toolkit discovery loading out into base_toolkit for reusability across other ETS libraries.


corranwebster:
It feels like a lot of this code belongs in `traits.etsconfig`, rather than here, since this really isn't Pyface-specific.  It's only here because Pyface is more-or-less the bottom of our current ETS GUI stack.

PR #319, opened 2018-04-28
qt: Actually raise if we can't find at QT api

bpteague:
I am writing a package where I would like the Qt GUI to be optional.  I expect that saying `import pyface.qt` without a Qt API will fail -- but it seems that the `else` clause is attached to the `try / except` construct, not the `for` loop.

This PR removes 8 spaces and restores the expected behavior.


corranwebster:
Good catch!  Thanks for the fix.

PR #322, opened 2018-05-17
BUG: Use open() instead of file().

### LINKED ISSUE #321: pyface.image.image is not Python3 compatible
pyface.image.image.py uses the old "file()" built-in which is not compatible in Python3.  Should be an easy fix to update for Python3 & backwards compatibility to Python2.

rkern:
Fixes #321


cfarrow:
LGTM

PR #328, opened 2018-07-21
Bugfix the Closes #327 for split_editor_area_pane not supporting Qt5 …

### LINKED ISSUE #327: Pyface Tasks split_editor_area_pane calls Qt methods removed in Qt5
My tasks-based GUI crashes when dragging tabs within the split-editor pane using pyface 6.0.0 and PyQt5.11.2.
split_editor_area_pane uses QPixmap.grabWidget which has been removed in Qt5.

A simple fix is to replace that line with "widget.grab()"
`pixmap1 = QtGui.QPixmap.grabWidget(tabBar, tab_rect)`
becomes
`pixmap1 = tabBar.grab()`

Additionally, the QStyleOptionTabBarBaseV2 is called which has also been removed in Qt5.  It appears simply replacing it with QStyleOptionTabBarBase is fine.  I found a list of functions that are obsolete in Qt5 from someone's backport code - I'm currently patching all of these to use the 'non-versioned' ones that are still available in Qt5.

Sorry for not trying to fix this and do a pull-request, I'm not sure how you are handling the pyqt5 migration and don't know how to best make this backwards compatible.

ericjmcd:
…when dragging tabs

### DIFF
--- a/pyface/ui/qt4/tasks/split_editor_area_pane.py
+++ b/pyface/ui/qt4/tasks/split_editor_area_pane.py
@@ -929,16 +929,23 @@ class TabDragObject(object):
         painter.fillRect(result_pixmap.rect(), QtCore.Qt.lightGray)
         painter.setCompositionMode(QtGui.QPainter.CompositionMode_SourceOver)
 
-        optTabBase = QtGui.QStyleOptionTabBarBaseV2()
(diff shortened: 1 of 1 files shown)


corranwebster:
Thanks for the fix - missed this going through my e-mails or I would have merged it earlier.

PR #336, opened 2018-08-10
Fix usage of event_loop() in tests

corranwebster:
Should have been being used as a context manager, rather than as straight call.  Hopefully this may lead to better testing results in some environments.

### DIFF
--- a/pyface/tests/test_about_dialog.py
+++ b/pyface/tests/test_about_dialog.py
@@ -11,13 +11,14 @@ from ..window import Window
 GuiTestAssistant = toolkit_object('util.gui_test_assistant:GuiTestAssistant')
 no_gui_test_assistant = (GuiTestAssistant.__name__ == 'Unimplemented')
 
-ModalDialogTester = toolkit_object('util.modal_dialog_tester:ModalDialogTester')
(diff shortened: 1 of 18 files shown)


pyface/tests/test_about_dialog.py
```diff
@@ -31,29 +32,28 @@ def tearDown(self):
...
     def test_create(self):
         # test that creation and destruction works as expected
-        self.dialog._create()
-        self.event_loop()
-        self.dialog.destroy()
-        self.event_loop()
+        with self.event_loop():
+            self.dialog._create()
+        with self.event_loop():
+            self.dialog.destroy()
```
itziakos: Note that in qt to destroy widgets one needs to run the real event loop instead of simulated event processing that `self.event_loop` does. I suggest that you use https://github.com/enthought/pyface/blob/master/pyface/ui/qt4/util/gui_test_assistant.py#L98

corranwebster: Recording an in-person conversation: I'm aware of the limitations here; proper clean-up should be happening in the `tearDown` method.

Many of there tests are smoke tests until we have better ways of testing the the UI is doing what we want in a cross-platform way.

PR #344, opened 2018-11-06
Make sure selected QtAPI actually provides QT support

cv3d:
Maybe related issues: enthought/mayavi#448, enthought/mayavi#460, enthought/mayavi#483

PyFace might select PyQt5 as the default API if it is installed even if Qt5 itself is missing.
This PR fix such an issue by ensuring that at least `QtCore` can be imported from the being-selected API, otherwise try the next one.

### DIFF
--- a/pyface/qt/__init__.py
+++ b/pyface/qt/__init__.py
@@ -61,7 +61,7 @@ else:
 if qt_api is None:
     for api_name, module in QtAPIs:
         try:
-            importlib.import_module(module)
+            importlib.import_module('.QtCore', module)
             qt_api = api_name
             break
         except ImportError:


corranwebster:
LGTM

Thanks for the contribution!

cv3d:
@corranwebster You are welcome!
BTW: The order [here](https://github.com/cv3d/pyface/blob/master/pyface/qt/__init__.py#L16) is quite strange! `pyqt` should be considered before `pyqt5`, as per the current strategy to prefer Qt4 over 5.
Shall I PR on that as well?

corranwebster:
I think that order is OK - it doesn't need to match the order in the files like `QtCore.py`, as they aren't expressing any sort of order/preference.  In general, we prefer PySide over PyQt, and more recent over older, _except_ that PySide2 is currently experimental for ETS (eg. currently tests are failing on windows).

cv3d:
I see! I thought the warning `Warning: The default toolkit if none is supplied is qt4.` in the README implies that Qt4 is preferred..

corranwebster:
Ah! No, that refers to the `ETS_TOOLKIT` not the `PYFACE_API`, which for historical reasons is called `qt4` even if the `PYFACE_API` is `pyqt5`.  We're planning to change that to `qt` in the future (and you can do that right now for _some_ ETS projects), but that requires care since there is so much infrastructure we have that assumes `ETS_TOOLKIT` is `qt4`.

For the record, the priorities for qt vs wx are here: https://github.com/enthought/pyface/blob/master/pyface/base_toolkit.py#L94

PR #351, opened 2019-01-23
FIX: Fix deprecation warnings

larsoner:
Fixes warnings of this type:
```
  /home/larsoner/python/pyface/build/lib/pyface/i_clipboard.py:17: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```
And also fixes the `stacklevel` in an internal `PendingDeprecationWarning`.

### DIFF
--- a/pyface/i_clipboard.py
+++ b/pyface/i_clipboard.py
@@ -14,7 +14,10 @@
 
 """ The interface for manipulating the toolkit clipboard.
 """
-from collections import Sequence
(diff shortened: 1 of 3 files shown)


corranwebster:
Thanks for the contribution, and apologies for the delay in reviewing.  This looks good to me, and the test failure is unrelated.

PR #356, opened 2019-03-06
BUG: Only display context menu when mouse is not on page.

rkern:
In the `SplitEditorAreaPane`, a right click will bring up a context menu asking if you want to split the pane. When a tab is displaying a Chaco plot, a right click on the Chaco plot will interrupt still bring up that context menu. It is the case that Enable/Chaco is not `accept()`ing the Qt `QMouseEvent`s when it handles them as it should, but even when I fix that, I still get these context menus. I am not sure what's going with that.

In this PR, I only allow the context menu when it's the default empty page being displayed (e.g. in a fresh pane split) or when the mouse is in the tab bar, which is where I would expect to have pane-related actions to be available.

I invite feedback if this change in the user experience is acceptable to your applications.

### DIFF
--- a/pyface/ui/qt4/tasks/split_editor_area_pane.py
+++ b/pyface/ui/qt4/tasks/split_editor_area_pane.py
@@ -814,10 +814,15 @@ class DraggableTabWidget(QtGui.QTabWidget):
     def contextMenuEvent(self, event):
         """ To show collapse context menu even on empty tabwidgets
         """
-        global_pos = self.mapToGlobal(event.pos())
-        menu = self.editor_area.get_context_menu(pos=global_pos)
-        qmenu = menu.create_menu(self)
-        qmenu.exec_(global_pos)
+        local_pos = event.pos()
+        if (self.empty_widget is not None or
+                self.tabBar().rect().contains(local_pos)):
+            # Only display if we are in the tab bar region or the whole area if
+            # we are displaying the default empty widget.
+            global_pos = self.mapToGlobal(local_pos)
+            menu = self.editor_area.get_context_menu(pos=global_pos)
+            qmenu = menu.create_menu(self)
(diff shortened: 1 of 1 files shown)


corranwebster:
This doesn't feel quite right - this isn't a mouse event (which Chaco would catch by setting `accept`), but a `QContextMenuEvent` which would imply that the Enable `Window` class has the wrong `contextMenuPolicy` if we don't want to generate these events - perhaps it should be set to `Qt.PreventContextMenu`? (See, eg. https://doc.qt.io/archives/qt-4.8/qt.html#ContextMenuPolicy-enum)

rkern:
That's a good point. Nonetheless, it seems to me that, from a UX perspective, I still wouldn't want a context menu for the `SplitEditorAreaPane` manipulations to show up when I right-click inside of my page (except perhaps for that empty widget).

jonathanrocher:
+1 on the desired behavior. More specifically, +1 for only allowing `the context menu when it's the default empty page being displayed (e.g. in a fresh pane split) or when the mouse is in the tab bar` .

The implementation discussion is largely above my head but I would add that, even without Chaco plots, the pane could contain a TraitsUI view which has its own handling of right click events (for example a TreeEditor). So it feels to me like a better solution to suppress the event handling in the pane rather than in enable.

rkern:
Most Qt widgets that care about context menus, one way or the other, properly set their context menu policy properly, so it's not an issue for them. It's only the case of the empty space between widgets and custom `QWidget`s like Enable's that neglect to override the default. Enable can and should be fixed, but I still think that the "empty space between widgets" inside one of these pages still should not expose the `SplitEditorAreaPane` context menu.

corranwebster:
That seems fair enough.  I have some concerns about how this might affect some internal codebases, but I think we can handle it if we need to.

PR #357, opened 2019-03-06
Add missing create_percent method

serwy:
See: https://github.com/enthought/traitsui/issues/487

### DIFF
--- a/pyface/ui/qt4/progress_dialog.py
+++ b/pyface/ui/qt4/progress_dialog.py
@@ -222,6 +222,12 @@ class ProgressDialog(MProgressDialog, Window):
         self._message_control = label
         return
 
+    def _create_percent(self, dialog, parent_sizer):
+        if not self.show_percent:
+            return
+
+        raise NotImplementedError
+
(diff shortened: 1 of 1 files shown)


pyface/ui/qt4/progress_dialog.py
```diff
@@ -222,6 +222,12 @@ def _create_message(self, dialog, layout):
         self._message_control = label
         return
 
+    def _create_percent(self, dialog, parent_sizer):
+        if not self.show_percent:
+            return
+
+        raise NotImplementedError
```
corranwebster: I think I'd be tempted to log or warn and do nothing rather than raise here.

serwy: I copied the method as it exists in the `wx` backed. Are we going to fix it in both places?

corranwebster: If this is the same as Wx, it's fine.

serwy:
@corranwebster Thank you for the review

PR #358, opened 2019-03-08
FIX: proper import of PySide2 QtWidgets, include QWebSettings, exclud…

bergtholdt:
…e QWebHistory (not available in current PySide2)


corranwebster:
Thanks - good catch.  Obviously we haven't been testing much with Pyside2 :)

PR #360, opened 2019-03-29
Convenience classes for widget Actions

### LINKED ISSUE #349: Allow Widgets in Toolbars
There should be a straightforward way to include a widget inside a toolbar.  This would presumably be a new type of `ActionManagerItem` that has a factory for the widget.

A complete implementation would include an appropriate Task `Schema` for these sorts of items.

corranwebster:
This PR adds:

- some simple field widgets for data entry
- an Action subclass that supports embedding them in toolbars
- an Action subclass that supports embedding TraitsUI views in toolbars

Screen Shot 2019-03-29 at 4 05 21 pm

Fixes #349

### DIFF
--- a/examples/python_editor.py
+++ b/examples/python_editor.py
@@ -134,10 +134,11 @@ class MainWindow(ApplicationWindow):
         if self.control:
             try:
                 self._editor.save()
-            except IOError as e:
+            except IOError:
                 # If you are trying to save to a file that doesn't exist,
                 # open up a FileDialog with a 'save as' action.
-                dlg = FileDialog(parent=self.control, action='save as', wildcard="*.py")
+                dlg = FileDialog(parent=self.control, action='save as',
+                                 wildcard="*.py")
                 if dlg.open() == OK:
                     self._editor.save(dlg.path)
 
@@ -162,7 +163,7 @@ if __name__ == '__main__':
     gui = GUI()
 
     # Create and open the main window.
-    window = MainWindow()
+    window = MainWindow(size=(800, 600))
     window.open()
 
     # Start the GUI event loop!
--- a/examples/python_editor.py
+++ b/examples/python_editor.py
@@ -149,9 +149,6 @@ class MainWindow(ApplicationWindow):
     def on_show_line_numbers(self):
         self._editor.show_line_numbers = not self._editor.show_line_numbers
 
-    def on_go_to_line(self, line_number):
-        self._editor.select_line(line_number+1)
-
     def on_style_changed(self, value):
         from pygments.styles import get_style_by_name
 
--- a/examples/python_editor.py
+++ b/examples/python_editor.py
@@ -13,67 +13,22 @@
 #------------------------------------------------------------------------------
 """ Python editor example. """
 
+from __future__ import print_function
 
 # Enthought library imports.
-from traits.api import Constant
+from traits.api import (
+    Any, Constant, Dict, Instance, Str, Type, on_trait_change
+)
 
 from pyface.api import (
     ApplicationWindow, FileDialog, GUI, OK, PythonEditor
 )
-from pyface.toolkit import toolkit_object
 from pyface.action.api import (
-    Action, ActionEvent, Group, MenuManager, MenuBarManager, ToolBarManager
+    Action, ActionEvent, FieldAction, Group, MenuManager, MenuBarManager,
+    ToolBarManager
 )
-
-
-class IntFieldAction(Action):
-    """ A widget action containing an integer textbox
-
-    This shows one way of creating a widget Action by overriding the
-    create_control method and overloading the `perform` methods to pass
-    the entered value to a handler.
-    """
-
-    style = Constant("widget")
-
-    def create_control(self, parent):
-        if toolkit_object.toolkit == 'wx':
-            import wx
-            from wx.lib.intctrl import IntCtrl
-            control = IntCtrl(
-                parent,
-                value=1,
-                style=wx.TE_PROCESS_ENTER
-            )
-
-            def handle_enter(event):
-                value = control.GetValue()
-                action_event = ActionEvent(value=value)
-                self.perform(action_event)
-
-            control.Bind(wx.EVT_TEXT_ENTER, handle_enter)
-
-        elif toolkit_object.toolkit == 'qt4':
-            from pyface.qt import QtGui
-            control = QtGui.QLineEdit(parent)
-            control.setText("1")
-            control.setValidator(QtGui.QIntValidator())
-
-            def handle_enter():
-                value = int(control.text())
-                action_event = ActionEvent(value=value)
-                self.perform(action_event)
-
-            control.returnPressed.connect(handle_enter)
-
-        else:
-            control = None
-
-        return control
-
-    def perform(self, event):
-        if self.on_perform is not None:
-            self.on_perform(event.value)
+from pyface.fields.api import ComboField
+from pyface.toolkit import toolkit_object
 
 
 class MainWindow(ApplicationWindow):
@@ -115,6 +70,9 @@ class MainWindow(ApplicationWindow):
 
         # Add a tool bar if we are using qt4 - wx has layout issues
         if toolkit_object.toolkit == 'qt4':
+            from pygments.styles import STYLE_MAP
+            styles = list(STYLE_MAP)
+
             self.tool_bar_manager = ToolBarManager(
                 Group(
                     Action(
@@ -133,15 +91,21 @@ class MainWindow(ApplicationWindow):
                 ),
                 Group(
                     Action(
-                        name="Show Line Numbers",
+                        name="Lines",
                         style='toggle',
                         on_perform=self.on_show_line_numbers,
                         checked=True,
                     ),
-                    IntFieldAction(
-                        name="Go to line",
-                        on_perform=self.on_go_to_line,
-                    )
+                    FieldAction(
+                        name='Style',
+                        field_type=ComboField,
+                        field_defaults={
+                            'values': styles,
+                            'value': u'default',
+                            'tooltip': u'Style',
+                        },
+                        on_perform=self.on_style_changed,
+                    ),
                 )
             )
 
@@ -164,7 +128,7 @@ class MainWindow(ApplicationWindow):
         """ Open a new file. """
 
         if self.control:
-            dlg = FileDialog(parent=self.control, wildcard="*.py")
+            dlg = FileDialog(parent=self.control, wildcard='*.py')
 
             if dlg.open() == OK:
                 self._editor.path = dlg.path
@@ -188,6 +152,17 @@ class MainWindow(ApplicationWindow):
     def on_go_to_line(self, line_number):
         self._editor.select_line(line_number+1)
 
+    def on_style_changed(self, value):
+        from pygments.styles import get_style_by_name
+
+        # XXX surface this to a proper API on the editor widget
+        # XXX Qt backend only
+        highlighter = self._editor.control.code.highlighter
+        highlighter._style = get_style_by_name(value)
+        highlighter._brushes = {}
+        highlighter._formats = {}
+        highlighter.rehighlight()
+
 
 # Application entry point.
 if __name__ == '__main__':
@@ -200,5 +175,3 @@ if __name__ == '__main__':
 
     # Start the GUI event loop!
     gui.start_event_loop()
-
-##### EOF #####################################################################
--- a/examples/python_editor.py
+++ b/examples/python_editor.py
@@ -15,16 +15,11 @@
 
 from __future__ import print_function, unicode_literals
 
-# Enthought library imports.
-from traits.api import (
-    Any, Constant, Dict, Instance, Str, Type, on_trait_change
-)
-
 from pyface.api import (
     ApplicationWindow, FileDialog, GUI, OK, PythonEditor
 )
 from pyface.action.api import (
-    Action, ActionEvent, FieldAction, Group, MenuManager, MenuBarManager,
+    Action, FieldAction, Group, MenuManager, MenuBarManager,
     ToolBarManager
 )
 from pyface.fields.api import ComboField
--- a/examples/python_editor.py
+++ b/examples/python_editor.py
@@ -13,7 +13,7 @@
 #------------------------------------------------------------------------------
 """ Python editor example. """
 
-from __future__ import print_function
+from __future__ import print_function, unicode_literals
 
 # Enthought library imports.
 from traits.api import (
@@ -101,8 +101,8 @@ class MainWindow(ApplicationWindow):
                         field_type=ComboField,
                         field_defaults={
                             'values': styles,
-                            'value': u'default',
-                            'tooltip': u'Style',
+                            'value': 'default',
+                            'tooltip': 'Style',
                         },
                         on_perform=self.on_style_changed,
                     ),
--- a/pyface/action/field_action.py
+++ b/pyface/action/field_action.py
@@ -11,6 +11,7 @@ from traits.api import Any, Constant, Dict, Str, Type
 
 from pyface.fields.i_field import IField
 from .action import Action
+from .action_event import ActionEvent
 
 
 class FieldAction(Action):
--- a/pyface/action/field_action.py
+++ b/pyface/action/field_action.py
@@ -27,7 +27,7 @@ class FieldAction(Action):
     #: The field to display.
     field_type = Type(IField)
 
-    #: The field to display.
+    #: The default trait values for the field.
     field_defaults = Dict(Str, Any)
 
     def create_control(self, parent):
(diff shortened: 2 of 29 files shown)


examples/python_editor.py
```diff
@@ -188,6 +152,17 @@ def on_show_line_numbers(self):
     def on_go_to_line(self, line_number):
         self._editor.select_line(line_number+1)
```
jwiggins: This method is dead code now

examples/python_editor.py
```diff
@@ -133,15 +91,21 @@ def __init__(self, **traits):
...
-                        name="Go to line",
-                        on_perform=self.on_go_to_line,
-                    )
+                    FieldAction(
+                        name='Style',
+                        field_type=ComboField,
+                        field_defaults={
+                            'values': styles,
+                            'value': u'default',
+                            'tooltip': u'Style',
```
jwiggins: It's probably safe to import `unicode_literals` here instead of using `u`-prefixes, right?

pyface/action/field_action.py
```diff
@@ -0,0 +1,74 @@
...
+
+        Returns
+        -------
+        control : toolkit control
+            A toolkit control or None.
+        """
+        field = self.field_type(parent=parent, **self.field_defaults)
+        field._create()
+        field.on_trait_change(self.value_updated, 'value')
+        field.control._field = field
```
jwiggins: Are you making a reference to `field` here because no other object will get a reference? (because you're returning `field.control`)

pyface/action/traitsui_widget_action.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # ------------------------------------------------------------------------
+    # Action interface
+    # ------------------------------------------------------------------------
+
+    def create_control(self, parent):
+        """ Called when creating a "widget" style action.
+
+        This constructs an IField-based control directly and binds changes to
+        the value to the `value_updated` method.
```
jwiggins: Looks like copy pasta here

pyface/fields/i_combo_field.py
```diff
@@ -0,0 +1,90 @@
...
+# under the conditions described in the aforementioned license.  The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+# Thanks for using Enthought open source!
+#
+# Author: Enthought, Inc.
+# Description: <Enthought pyface package component>
+#------------------------------------------------------------------------------
+""" The text field interface. """
+
+from six import text_type
```
jwiggins: Using `six`... so, how do you feel about importing `__future__.[absolute_import|division|print_function|unicode_literals]`?

pyface/fields/i_spin_field.py
```diff
@@ -0,0 +1,80 @@
...
+class ISpinField(IField):
+    """ The spin field interface.
+
+    This is for spinners holding integer values.
+    """
+
+    #: The current value of the spinner
+    value = Range(low='minimum', high='maximum')
+
+    #: The current value of the spinner
```
jwiggins: This is a copied comment

pyface/fields/i_spin_field.py
```diff
@@ -0,0 +1,80 @@
...
+    #: The maximum value
+    maximum = Property(Int, depends_on='bounds')
+
+
+class MSpinField(HasTraits):
+
+    #: The current value of the spinner
+    value = Range(low='minimum', high='maximum')
+
+    #: The current value of the spinner
```
jwiggins: Same here

pyface/fields/i_spin_field.py
```diff
@@ -0,0 +1,80 @@
...
+            self.bounds = (value, value)
+        else:
+            self.bounds = (value, self.maximum)
+        if value > self.value:
+            self.value = value
+
+    def _get_maximum(self):
+        return self.bounds[1]
+
+    def _set_minimum(self, value):
```
jwiggins: You mean `_set_maximum`?

pyface/ui/qt4/fields/spin_field.py
```diff
@@ -0,0 +1,77 @@
...
+from traits.api import Bool, Enum, Str, Trait, provides
+
+from pyface.fields.i_spin_field import ISpinField, MSpinField
+from pyface.qt.QtGui import QSpinBox
+from .field import Field
+
+
+@provides(ISpinField)
+class SpinField(MSpinField, Field):
+    """ The Qt-specific implementation of the text field class """
```
jwiggins: text field class?

pyface/action/field_action.py
corranwebster: This is a hack to get a long-lived reference to `field` so that it isn't GC'd during the lifetime of the underlying toolkit control.  This is done in part because `Action`s don't know whether they have a 1-to-1 or 1-to-many relationship with the concrete toolbar controls... not how I would code all of this from scratch.

I do a similar thing with the `UI` instance for the TraitsUI code for the same reason.

pyface/ui/qt4/fields/spin_field.py
corranwebster: Copy pasta

pyface/action/field_action.py
jwiggins: I suspected that was what was going on.

pyface/fields/i_combo_field.py
corranwebster: Done

examples/python_editor.py
```diff
@@ -13,67 +13,22 @@
 #------------------------------------------------------------------------------
 """ Python editor example. """
 
+from __future__ import print_function, unicode_literals
 
 # Enthought library imports.
-from traits.api import Constant
+from traits.api import (
+    Any, Constant, Dict, Instance, Str, Type, on_trait_change
+)
```
jwiggins: None of these are being used here.

examples/python_editor.py
```diff
@@ -13,67 +13,22 @@
...
 
 from pyface.api import (
     ApplicationWindow, FileDialog, GUI, OK, PythonEditor
 )
-from pyface.toolkit import toolkit_object
 from pyface.action.api import (
-    Action, ActionEvent, Group, MenuManager, MenuBarManager, ToolBarManager
+    Action, ActionEvent, FieldAction, Group, MenuManager, MenuBarManager,
```
jwiggins: `ActionEvent` is unused

pyface/action/traitsui_widget_action.py
```diff
@@ -0,0 +1,76 @@
...
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in enthought/LICENSE.txt and may be redistributed only
+# under the conditions described in the aforementioned license.  The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+# Thanks for using Enthought open source!
+
+from traits.api import Constant, HasTraits, Instance
+
+from pyface.fields.i_field import IField
```
jwiggins: `IField` is unused here

pyface/ui/qt4/fields/field.py
```diff
@@ -0,0 +1,94 @@
...
+    # IWidget interface
+    # ------------------------------------------------------------------------
+
+    def _add_event_listeners(self):
+        """ Set up toolkit-specific bindings for events """
+        super(Field, self)._add_event_listeners()
+        if self.control is not None:
+            self.control.customContextMenuRequested.connect(
+                    self._handle_context_menu)
+            self.control.contextMenuPolicy(Qt.CustomContextMenu)
```
jwiggins: This is different from the behavior down below in `_context_menu_changed`. Was that intentional?

pyface/ui/qt4/fields/spin_field.py
```diff
@@ -0,0 +1,79 @@
...
+        super(SpinField, self)._remove_event_listeners()
+
+    # ------------------------------------------------------------------------
+    # Private interface
+    # ------------------------------------------------------------------------
+
+    def _handle_update(self):
+        if self.control is not None:
+            value = self.control.value()
+            self._value_updated(value)
```
jwiggins: This method appears to have leaked in from the `TextField` implementation

pyface/ui/qt4/fields/text_field.py
```diff
@@ -0,0 +1,119 @@
...
+            self.control.setText(new)
+
+    def _update_text_changed(self, new):
+        if self.control is not None:
+            if new == 'editing_finished':
+                self.control.editingFinished.disconnect(self._handle_update)
+                self.control.textEdited.connect(self._handle_update)
+            else:
+                self.control.textEdited.disconnect(self._handle_update)
+                self.control.editingFinished.connect(self._handle_update)
```
jwiggins: This method reads a bit funny if you're not paying close attention. Can you add a comment explaining what's happening?

corranwebster:
I'm starting to write tests and finding some interesting issues, so there are likely to be some updates on this (currently `SpinField` is behaving oddly in tests).

corranwebster:
This is being delayed a bit more: dispatch of traits listeners needs to be on the UI thread so I need to refactor away from `_*_changed` forms :(

corranwebster:
... except UI dispatch requires TraitsUI to be imported, which means tests have errors unless there happens to have been an import of TraitsUI somewhere.

I am knee-deep in yak hair right now.

corranwebster:
@jwiggins This has been thoroughly worked over, so a re-review would be appreciated at some point.

pyface/action/tests/test_traitsui_widget_action.py
```diff
@@ -70,19 +119,40 @@ class ComplexEnumAction(TraitsUIWidgetAction):
...
 
         editor = control._ui.get_editors('value')[1]
 
         with self.assertTraitChanges(action, 'value', count=1):
-            editor.control.setCurrentIndex(2)
+            if toolkit.toolkit in {'qt', 'qt4'}:
+                editor.control.setCurrentIndex(2)
+                editor.control.activated.emit(2)
+            elif toolkit.toolkit == 'wx':
+                event = wx.CommandEvent(wx.EVT_CHOICE.typeId,
```
jwiggins: Missing `import wx`?

pyface/action/tests/test_traitsui_widget_action.py
```diff
@@ -70,19 +119,40 @@ class ComplexEnumAction(TraitsUIWidgetAction):
...
+                                        editor.control.GetId())
+                event.SetString('c')
+                wx.PostEvent(editor.control.GetEventHandler(), event)
+            else:
+                self.skipTest("Unknown toolkit")
+            GUI.process_events()
```
jwiggins: Why not `self.gui.process_events()`?

pyface/fields/tests/test_combo_field.py
```diff
@@ -1,163 +1,114 @@
...
 
 
-class TestComboField(unittest.TestCase):
+class TestComboField(unittest.TestCase, FieldMixin):
 
-    def test_combo_field(self):
-        widget = ComboField(
-            parent=None,
+    def setUp(self):
+        FieldMixin.setUp(self)
```
jwiggins: Will this not be called automatically?

pyface/fields/tests/test_combo_field.py
```diff
@@ -1,163 +1,114 @@
...
+        self.widget.values = ['four', 'five', 'one', 'six']
+        self.gui.process_events()
+
+        # XXX different results in Wx and Qt
+        # As best I can tell, difference is at the Traits level
+        # On Qt setting 'values' sets 'value' to "four" before combofield
+        # handler sees it.  On Wx it remains  as "one" at point of handler
+        # call.  Possibly down to dictionary ordering or something
+        # similar.
+        self.assertIn(self.widget.value, {'one', 'four'})
```
jwiggins: Is this non-deterministic?

pyface/fields/tests/test_spin_field.py
```diff
@@ -1,96 +1,60 @@
...
-        widget._create()
-        try:
-            widget.show(True)
-            widget.value = 5
-
-            self.assertEqual(widget._get_control_value(), 5)
-        finally:
-            widget.destroy()
+    def setUp(self):
+        FieldMixin.setUp(self)
```
jwiggins: Same question. Why?

pyface/fields/tests/test_text_field.py
```diff
@@ -1,87 +1,70 @@
...
-        try:
-            widget.show(True)
-            widget.value = 'new value'
+class TestTextField(unittest.TestCase, FieldMixin):
 
-            self.assertEqual(widget._get_control_value(), 'new value')
-        finally:
-            widget.destroy()
+    def setUp(self):
+        FieldMixin.setUp(self)
```
jwiggins: Is there some reason to avoid calling `unittest.TestCase.setup()`?

pyface/ui/wx/fields/field.py
```diff
@@ -0,0 +1,88 @@
...
+# Thanks for using Enthought open source!
+#
+# Author: Enthought, Inc.
+# Description: <Enthought pyface package component>
+#------------------------------------------------------------------------------
+""" The Wx-specific implementation of the text field class """
+
+from __future__ import absolute_import, print_function, unicode_literals
+
+from traits.api import Any, Bool, Enum, Instance, Trait, Unicode, provides
```
jwiggins: `Bool`, `Enum`, and `Trait` are unused

pyface/action/tests/test_traitsui_widget_action.py
corranwebster: No, it's imported on line 132 above.

pyface/action/tests/test_traitsui_widget_action.py
corranwebster: Because I wrote this line before I added `self.gui`.  Will fix.

pyface/fields/tests/test_combo_field.py
corranwebster: No, because `unittest.TestCase` has an empty `setUp` that doesn't call `super`.  :(

However, I could fix by swapping the order of the superclasses, which would be cleaner.

pyface/fields/tests/test_combo_field.py
corranwebster: No, it seems to always be one way in the Qt implementation and the other for Wx but the difference occurs _prior_ to the toolkit code being called (ie. on one backend the `_values_updated` handler gets called with `self.value == 'one'` and on the other it gets called with `self.value == 'four'`.  So my suspicion is that there is a trait handler to reset the `value` trait on changes to `values` and it is not specified which one gets called first.

So my fear is that changing the structure of each trait class, or running in a different context, may case non-deterministic behaviour; but I haven't dug deeper, because even if I do find out what is causing it, I suspect there is nothing much I can _do_ about it.

pyface/fields/tests/test_text_field.py
corranwebster: Yes, see above.

pyface/ui/wx/fields/text_field.py
```diff
@@ -0,0 +1,100 @@
...
+        self.control.SetHint(placeholder)
+
+    def _get_control_echo(self):
+        """ Toolkit specific method to get the control's echo. """
+        return self.echo
+
+    def _set_control_echo(self, echo):
+        """ Toolkit specific method to set the control's echo. """
+        # Can't change echo on Wx after control has been created."
+        pass
```
jwiggins: Do you think it would be annoying to warn here?

pyface/ui/wx/fields/field.py
corranwebster: Thanks - will fix.

pyface/action/tests/test_traitsui_widget_action.py
jwiggins: Ugh... I hate Python's scope rules sometimes...

pyface/fields/tests/test_combo_field.py
jwiggins: How delightful! I'll leave it to you if you want to change it...

pyface/ui/wx/fields/text_field.py
corranwebster: Yes.  I had a warning there for a while and it was annoying :)

It gets called by the control initialization method, so every time you create a control it emits the warning.

pyface/fields/tests/test_combo_field.py
jwiggins: OK. I suppose this note will have to suffice for the time being.

jwiggins:
It's good that this is covered by tests now. I won't push too hard for big changes, because I'd rather see this merged soon.

That said, the repetition in the testing code is a bit of a smell. It would be nice to track that in a separate issue and clean it up later. (I see that `GuiTestAssistant` only exists for Qt...)

corranwebster:
I think the testing of the field classes is about as tight as you are going to be able to get it, although it would be nice to have the GuiTestAssistant available to make some of the event processing nicer and a bit more robust.  The action testing is not as good, but it should really be done as part of re-writing the tests for all the actions.

pyface/ui/wx/fields/text_field.py
jwiggins: Works for me

corranwebster:
Looks good.  Merging.

PR #368, opened 2019-04-24
Add some missing tests on text field

corranwebster:
Forgot tests for a couple of attributes.

### DIFF
--- a/pyface/fields/i_text_field.py
+++ b/pyface/fields/i_text_field.py
@@ -44,8 +44,6 @@ class ITextField(IField):
 class MTextField(HasTraits):
     """ The text field mix-in. """
 
-    # this currently does nothing
-
     #: The value held by the field.
     value = Unicode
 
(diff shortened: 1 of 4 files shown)


jwiggins:
This is good? The tests you added failed. TDD!

corranwebster:
The code was written on a computer where I couldn't easily do testing (didn't particularly want to install EDM on it).  But it _is_ good that it failed, as it looks like there may be an actual bug or two.

corranwebster:
Test cases hopefully now working on all platforms.  This is why we write tests.

pyface/fields/i_text_field.py
```diff
@@ -144,6 +142,7 @@ def _observe_control_editing_finished(self, remove=False):
     # Trait change handlers --------------------------------------------------
 
     def _placeholder_updated(self):
+        print(self.control, self.placeholder)
```
jwiggins: Debug droppings

PR #387, opened 2019-05-30
Update CI

corranwebster:
CI updates and fixes:

- Use PyQt5 from EDM
- Handle QtWebKit in PyQt5 5.6

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -107,7 +107,7 @@ extra_dependencies = {
     # XXX once pyside2 is available in EDM, we will want it here
     'pyside2': set(),
     'pyqt': {'pyqt<4.12'},  # FIXME: build 1 of.4-12 appears to be bad
-    'pyqt5': 'pyqt5',
+    'pyqt5': {'pyqt5'},
     'wx': {'wxpython'},
     'null': set()
 }
--- a/etstool.py
+++ b/etstool.py
@@ -107,8 +107,7 @@ extra_dependencies = {
     # XXX once pyside2 is available in EDM, we will want it here
     'pyside2': set(),
     'pyqt': {'pyqt<4.12'},  # FIXME: build 1 of.4-12 appears to be bad
-    # XXX once pyqt5 is available in EDM, we will want it here
-    'pyqt5': set(),
+    'pyqt5': 'pyqt5',
     'wx': {'wxpython'},
     'null': set()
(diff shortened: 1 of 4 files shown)


etstool.py
```diff
@@ -107,8 +107,7 @@
     # XXX once pyside2 is available in EDM, we will want it here
     'pyside2': set(),
     'pyqt': {'pyqt<4.12'},  # FIXME: build 1 of.4-12 appears to be bad
-    # XXX once pyqt5 is available in EDM, we will want it here
-    'pyqt5': set(),
+    'pyqt5': 'pyqt5',
```
rahulporuri: need to be a `set`

corranwebster: Not having a good day PR-wise :(

pyface/qt/QtWebKit.py
```diff
@@ -4,16 +4,18 @@
...
-        QWebEnginePage as QWebPage,
-        QWebEngineView as QWebView,
-    )
+    try:
+        from PyQt5.QtWebEngine import *
+        from PyQt5.QtWebEngineWidgets import (
+            QWebEngineHistory as QWebHistory,
+            QWebEngineHistoryItem as QWebHistoryItem,
+            QWebEnginePage as QWebPage,
+            QWebEngineView as QWebView,
```
rahulporuri: question : is `QWebEngineSettings` not available via PyQt5 because we import it from PySide2 but not here.

pyface/qt/QtWebKit.py
```diff
@@ -27,6 +29,7 @@
             QWebEngineSettings as QWebSettings,
```
rahulporuri: Q : Is there a reason why we can't import QWebEngineHistory from PySide2 because we import it from PyQt5.

corranwebster: I haven't investigated fully, but I recall that PySide2 didn't provide `QWebEngineHistory` at least at some point.  Unfortunately PyQt and PySide don't always have completely compatible APIs.  I'll do a quick check to see if it is in the current PySide2 release.

corranwebster: Looks like it now works.

corranwebster:
i think everything here has been resolved by other PRs.  I'm going to close this.

PR #391, opened 2019-06-06
Update unittest imports

rahulporuri:
Import from the standard library instead of importing it from `traits.testing.unittest_tools`


corranwebster:
Generally LGTM

PR #410, opened 2019-06-22
BUG: fix error during tab completion on shell.

prabhuramachandran:
This seems like an error from Python-2 specific code.


corranwebster:
LGTM

Issue #408, opened 2019-06-19
Issue with timer API changes


prabhuramachandran:
With latest master there is a problem with the timer API it appears, here is a minimal example:
```
from pyface.timer.api import Timer

t = Timer(1000, lambda: None)
t.IsRunning()
```
This will produce an `AttributeError: 'Timer' object has no attribute 'is_running'`.  I apologize for not testing this sooner. I am only now getting around to updating to the latest pyface/traits/traitsui.

prabhuramachandran:
There is another issue with the new API:
```
from pyface.timer.api import Timer
t = Timer(1000, lambda: None)
t.active = False
```
Produces this traceback:
```
i_timer.py in _set_active(self, value)
    238
    239     def _set_active(self, value):
--> 240         if self.value:
    241             self.start()
    242         else:

AttributeError: 'Timer' object has no attribute 'value'
```

corranwebster:
Thanks for the report - fixes look simple, but unfortunately we just missed a bugfix release.  Are you able to switch to using a `CallbackTimer` instead for your use-case?

corranwebster:
Thanks for the report - fixes look simple, but unfortunately we just missed a bugfix release.  Are you able to switch to using a `CallbackTimer` instead for your use-case?

prabhuramachandran:
@corranwebster -- no worries, I will work around the issue, I just wanted to point the error out.  I found something else as well and will report that a bit later.  Thanks.

corranwebster:
@prabhuramachandran #409 hopefully should fix the problems - would you be able to check?

prabhuramachandran:
Thanks!  I just did a quick check and it seems fine, my old code I have workarounds for anyway.  I will be sure to report any issues I do find.  I think you can safely merge #409.  I just sent a trivial PR for another issue I found in #410.

prabhuramachandran:
Am reopening as something is still busted here with the qt timer.
```
from pyface.timer.api import Timer

t = Timer(1000, lambda: None)
assert t.IsRunning()
assert t._timer.interval() == 1000

t.Stop()
t.Start(100)
print(t._timer.interval())
assert t._timer.interval() == 100
```
So, it looks like the interval is not being set correctly and it appears if the interval is changed, also if a user changes the `interval` trait, the underlying timer is still using the same interval.
```
t = Timer(1000, lambda: None)
assert t.IsRunning()
assert t._timer.interval() == 1000

t.interval = 100
assert t.IsRunning()
[...]

PR #414, opened 2019-07-04
Fix timer issue #408.

prabhuramachandran:
This adds a test and fixes the issue.  It doesn't address setting the interval trait when the timer is running.

### DIFF
--- a/pyface/timer/tests/test_timer.py
+++ b/pyface/timer/tests/test_timer.py
@@ -373,14 +373,24 @@ class TestTimer(TestCase, GuiTestAssistant):
     def test_restart(self):
         handler = ConditionHandler()
 
-        timer = Timer(250, handler.callback)
+        timer = Timer(20, handler.callback)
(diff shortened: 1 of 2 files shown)


corranwebster:
Thanks!  LGTM.

pyface/timer/tests/test_timer.py
```diff
@@ -373,14 +373,24 @@ def test_basic(self):
...
+        self.event_loop_helper.event_loop()
+        time.sleep(0.1)
+        self.assertEqual(handler.count, count)
 
         timer.Start()
         try:
             self.assertTrue(timer.IsRunning())
-            self.event_loop_helper.event_loop()
+            self.event_loop_helper.event_loop_until_condition(
+                lambda: handler.count > 0
```
prabhuramachandran: Noticed a small bug, this should be `handler.count > count` in the off chance that count is already > 0.

PR #448, opened 2019-09-17
Fix relative imports for wx tasks

janvonrickenbach:
These imports failed for me when trying to import envisage.plugins.ipython_kernel.ipython_kernel_ui_plugin.IPythonKernelUIPlugin


corranwebster:
Thanks for the bugfix!

PR #451, opened 2019-09-23
python_shell: Pass __init__ keyword arguments also to _create_control

janvonrickenbach:
Not really sure if this is the best way to do it. I would like to pass arguments to the PyShell.__init__ function.


corranwebster:
Thanks for the contribution!

As you note, this probably isn't the best way to do it; in particular the keyword args are going to be wxPython-specific and so will break compatibility with the equivalent Qt toolkit widgets; and they also conflict with the way Traits initialises objects (notice that `**traits` is used in the `super` call - that is no accident).

Which keyword arguments (ie. which features) are you particularly needing?  With a little bit more work it may be possible to support them on both back-ends, probably by adding some additional Traits that reflect those particular needs.

Issue #450, opened 2019-09-20
Namespaced sip


opoplawski:
Forwarded from: https://bugzilla.redhat.com/show_bug.cgi?id=1753414

python-pyface currently includes code that explicitly loads non-namespaced sip module,

yface/ui/qt4/workbench/split_tab_widget.py:            import sip
pyface/ui/qt4/workbench/split_tab_widget.py:            if sip.isdeleted(self):
pyface/qt/__init__.py:    import sip
pyface/qt/__init__.py:        sip.setapi('QDate', 2)
pyface/qt/__init__.py:        sip.setapi('QDateTime', 2)
pyface/qt/__init__.py:        sip.setapi('QString', 2)
pyface/qt/__init__.py:        sip.setapi('QTextStream', 2)
pyface/qt/__init__.py:        sip.setapi('QTime', 2)
pyface/qt/__init__.py:        sip.setapi('QUrl', 2)
pyface/qt/__init__.py:        sip.setapi('QVariant', 2)

Recent versions of sip now use name-spaced modules, so either this code can either be dropped or use PyQt4.sip instead.

(Or ideally, appears it at least has *some* support for PyQt5 which is better supported).

See also bug
https://bugzilla.redhat.com/show_bug.cgi?id=1753069

corranwebster:
Hi, thanks for the report, but I'm not quite sure what the problem is - the reference to the redhat issue is not particularly enlightening.  Most of the code shown above is only run in the case that the Qt toolkit being used is PyQt4 and is not run when using the PyQt5 toolkit.  We are doing essentially the same thing as, eg. qtpy here https://github.com/spyder-ide/qtpy/blob/master/qtpy/__init__.py#L181

Can you provide a bit more detail about what the issue is, with pointers to sip documentation showing best practices if we are not following them.

opoplawski:
Honestly, I don't think I can - I'm really just relaying the bug that was filed in fedora.  But apparently one should be using PyQt4.sip.... instead of sip.... ?

corranwebster:
OK, I have dug a little deeper and this appears to be a very new, backwards incompatible change in PyQt's API.  Importing `PyQt4.sip` does not work on versions of PyQt earlier than 4.12.2 or PyQt 5.11, and we need to be able to support earlier versions.

[...]

PR #452, opened 2019-09-29
Support PyQt 4.12.2

corranwebster:
PyQt 4.12.2 changes to use a vendorized `sip`.  See https://www.riverbankcomputing.com/static/Docs/PyQt4/incompatibilities.html#pyqt-v4-12-2

This also makes an update to only call `prepare_pyqt4` on Python 2, as it isn't needed in Python 3.

See #450 for more discussion.

### DIFF
--- a/pyface/qt/__init__.py
+++ b/pyface/qt/__init__.py
@@ -39,19 +39,15 @@ def prepare_pyqt4():
         sip.setapi('QUrl', 2)
         sip.setapi('QVariant', 2)
     except ValueError as exc:
-        if sys.version_info[0] <= 2:
-            # most likely caused by something else setting the API version
-            # before us: try to give a better error message to direct the user
-            # how to fix.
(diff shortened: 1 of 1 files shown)


pyface/qt/__init__.py
```diff
@@ -56,7 +60,7 @@ def prepare_pyqt4():
 for api_name, module in QtAPIs:
     if module in sys.modules:
         qt_api = api_name
-        if qt_api == 'pyqt':
+        if qt_api == 'pyqt' and sys.version_info[0] <= 2:
```
rahulporuri: In that case, there's the `if/else` clause in `prepare_pyqt4` which handles the exception differently for python 2 and 3 can also be removed.

Or do we want to keep it in case someone explicitly calls `prepare_pyqt4`?

corranwebster:
@rahulporuri Any other comments, or does this look good to go?

PR #455, opened 2019-11-05
Python3 compatibility updates in fix_introspect_bug

David-Baddeley:
- types.TupleType is gone in python3. isinstance(object, tuple) should work equally well on python 2.7. 

- isinstance check with object is also required to avoid infinite recursion on python 3

- parameter name object changed  to obj to avoid conflict with builtin object


corranwebster:
Thanks for the contribution!

We probably never noticed this because the wx backend isn't properly supported on Python 3, which is the only place this is used currently, but we are planning to move off of Python 2.7 soon, so this is useful.

PR #457, opened 2019-11-11
MAINT: Update deprecated attribute name.

jvkersch:
This changes the deprecated `traits` attribute name into `traits_name`, from the (privately imported) `TraitsChangeCollector`. This was renamed in https://github.com/enthought/traits/pull/343.

I realize that this code may be obsoleted by https://github.com/enthought/pyface/pull/447, but having one less `DeprecationWarning` for pyface end users would be a good thing, in the meantime.


jvkersch:
@corranwebster Quick review?

corranwebster:
LGTM.  Thanks!

PR #467, opened 2019-12-19
Include additional copyright notices in AboutDialog

### LINKED ISSUE #466: Allow to customise copyright notice in AboutDialog
I found myself in a situation where I needed to replace the copyright notice provided in `AboutDialog` with a custom one, but there was no easy way to do so apart from copy-pasting the `AboutDialog` class from `pyface.ui.qt4.about_dialog.py` and directly editing its `_DIALOG_TEXT`. 

Would it be possible to provide a simpler way to customise the copyright lines?

robertopreste:
Closes #466 

As per offline discussion with @corranwebster , we decided to leave the existing Enthought and Riverbank copyright lines as they are, and add an optional trait that will allow users to include their own copyright notices. This new trait works exactly as the `additions` trait, but prepends each line with `Copyright (c) `.

I also had to refactor a bit both `ui/qt4/about_dialog.py` and `ui/wx/about_dialog.py` to accommodate a new `AboutDialog._create_html()` method, which builds the html shown in the dialog. This allows to properly test its content, something that was missing so far.

### DIFF
--- a/pyface/ui/qt4/about_dialog.py
+++ b/pyface/ui/qt4/about_dialog.py
@@ -1,6 +1,6 @@
 #------------------------------------------------------------------------------
 # Copyright (c) 2007, Riverbank Computing Limited
-# Copyright (c) 2019, Enthought, Inc.
+# Copyright (c) 2020, Enthought, Inc.
 # All rights reserved.
 #
 # This software is provided without warranty under the terms of the BSD license.
@@ -121,7 +121,7 @@ class AboutDialog(MAboutDialog, Dialog):
         qt_version = QtCore.__version__
 
         # The additional copyright strings.
-        copyrights = "<br />".join(["Copyright (c) %s" % line
+        copyrights = "<br />".join(["Copyright &copy; %s" % line
                                     for line in self.copyrights])
 
         return _DIALOG_TEXT % (path, additions, py_version, qt_version,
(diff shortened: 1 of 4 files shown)


rkern:
I think I'd prefer that we permit a complete replacement of all of the copyright lines. The dates given for the Enthought copyright are not going to be correct for the application this dialog will be shown in. They are already out of date, even if you restrict them to just the copyright on PyFace.

pyface/ui/qt4/about_dialog.py
```diff
@@ -112,3 +107,22 @@ def _create_contents(self, parent):
...
+
+        # The additional strings.
+        additions = '<br />'.join(self.additions)
+
+        # Get the version numbers.
+        py_version = platform.python_version()
+        qt_version = QtCore.__version__
+
+        # The additional copyright strings.
+        copyrights = "<br />".join(["Copyright (c) %s" % line
```
rkern: We should be consistent about using the full `©` symbol.

robertopreste:
> I think I'd prefer that we permit a complete replacement of all of the copyright lines. 

I agree, but my understanding is that there might be some legal issue related to removing those lines too, especially related to the Riverbank one (@corranwebster can better explain this). 

> The dates given for the Enthought copyright are not going to be correct for the application this dialog will be shown in. They are already out of date, even if you restrict them to just the copyright on PyFace.

Do you mean this line? `Copyright &copy; 2003-2010 Enthought, Inc.<br>` 
I guess I can update this one as well as I did in the header.

pyface/ui/qt4/about_dialog.py
robertopreste: You mean replacing `(c)` with `&copy;`, right?

robertopreste:
What's the situation about this one? Is it mergeable?

PR #470, opened 2020-01-11
Test against Traits master

corranwebster:
Test against the master branch of Traits rather than the latest release since there have been a lot of changes.

This means turning off Python 2.7 testing.  We also stop running CI against 3.5 because it is not giving much information that testing on 3.6 isn't already giving us.


corranwebster:
PyQt5 appears to be having unrelated issues on Windows.  Making an allowed failure for now.

Issue #483, opened 2020-02-18
Action images don't update


corranwebster:
If you create an `Action` with an `image`, create a toolbar or menu item with it, and then update the image, it does not change on the screen.

These sections of code:
https://github.com/enthought/pyface/blob/03cbcfbe6e36167576a97462fa86ea9ef1b0ea79/pyface/ui/qt4/action/action_item.py#L323-L330
and
https://github.com/enthought/pyface/blob/03cbcfbe6e36167576a97462fa86ea9ef1b0ea79/pyface/ui/qt4/action/action_item.py#L405-L428
handle updating the UI when something on the underlying `Action` changes; it does not include updating if the `image` trait changes.

Similar code for Menus and the WxBackend also need to be updated.

Corwinpro:
Thank you @corranwebster. Should the [`ListeningAction`](https://github.com/enthought/pyface/blob/master/pyface/action/listening_action.py) also be updated? This class implements `_enabled_name_changed` and `_visible_name_changed` but doesn't implement the `_image_changed` method.

Thank you!

corranwebster:
No, `ListeningAction` doesn't need updating.  If a user wants to update the image based on the state of another object, they can subclass `ListeningAction` as needed.

Corwinpro:
I could fix the missing `_on_action_image_changed` method for the  `_Tool` class, and it seems to be working fine for the original issue I had. I am happy to work on the PR for this issue, if that's alright. I could not find the test suite for the `action_item.py` however.

PR #484, opened 2020-02-18
Action image (icon) attribute UI update

Corwinpro:
This PR approaches #483 

We add `_on_action_image_changed` to the Qt `_MenuItem`, `_Tool` and `_Menu`, and to the wx `_MenuItem`. We add `_on_action_tooltip_changed` to the Qt `_MenuItem` and `_Tool`.

### Done

- Implemented the `_on_action_image_changed` method that sets the control icon to the icon provided by the `action.image.create_icon()`.
- Implemented the `_on_action_tooltip_changed` method that sets the tooltip provided by the `action.tooltip`.

### DIFF
--- a/pyface/ui/qt4/action/menu_manager.py
+++ b/pyface/ui/qt4/action/menu_manager.py
@@ -118,6 +118,7 @@ class _Menu(QtGui.QMenu):
         self._manager.on_trait_change(self._on_enabled_changed, 'enabled')
         self._manager.on_trait_change(self._on_visible_changed, 'visible')
         self._manager.on_trait_change(self._on_name_changed, 'name')
+        self._manager.on_trait_change(self._on_image_changed, 'image')
         self.setEnabled(self._manager.enabled)
         self.menuAction().setVisible(self._manager.visible)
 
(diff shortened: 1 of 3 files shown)


pyface/ui/qt4/action/menu_manager.py
```diff
@@ -189,6 +190,13 @@ def _on_name_changed(self, obj, trait_name, old, new):
 
         return
 
+    def _on_image_changed(self, obj, trait_name, old, new):
+        """ Dynamic trait change handler. """
+
+        self.menuAction().setIcon(new.create_icon())
+
+        return
```
corranwebster: These dangling returns are part of an old style that we don't use any more.

pyface/ui/wx/action/action_item.py
```diff
@@ -244,6 +245,14 @@ def _on_action_name_changed(self, action, trait_name, old, new):
 
         return
 
+    def _on_action_image_changed(self, action, trait_name, old, new):
+        """ Called when the name trait is changed on an action. """
+
+        if self.control is not None:
+            self.control.SetIcon(action.image.create_icon())
+
+        return
```
corranwebster: Same comment here.

corranwebster:
Going to merge this anyway.  The "return at end of method" stuff can be resolved in a separate PR.

PR #492, opened 2020-02-22
Remove six and __future__ imports

corranwebster:
This is now a Python 3.5+ only library.


corranwebster:
Something went badly wrong with search and replace of `__future__` imports.  Closing and trying again.

PR #501, opened 2020-04-15
Pyface 7.0.0 release

corranwebster:
This is the release PR for Pyface 7.0.0.

This is not yet ready to merge.


CHANGES.txt
```diff
@@ -2,6 +2,82 @@
...
+* Allow live update of Action images (#484)
+* Add Qt Multimedia APIs to pyface.qt (#485)
+* Allow additional copyright notices in the AboutDialog class (#467)
+* Replace default AboutDialog and Splashscreen images with Python logo (#454)
+* Allow etstool.py to run in non EDM environments (#443)
+* Add assertEventuallyTrueInGui method to GuiTestAssistant (#419)
+
+Fixes
+
+* Avoid imports of Wx in interface classes (#500`)
```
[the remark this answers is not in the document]
corranwebster: Thanks

README.rst
```diff
@@ -12,20 +12,17 @@ pyface: traits-capable windowing framework
...
 the details of displaying them.
 
 The following GUI backends are supported:
 
-- wxPython
-- PyQt
-- PySide
-
-**Warning:** The default toolkit if none is supplied is ``qt4``.
-   This changed from ``wx`` in Pyface 5.0..
```
kitchoi, line 28: There is still a default toolkit and it might be worth mentioning in the README anyway, without the "warning" prefix and the "changed from wx" suffix? This is minor and should not have to block the release though.

corranwebster: I decided that it had been long enough that the Qt default doesn't need to be called out in the readme any more (it is talked about in the docs).

ci-src-requirements.txt
```diff
@@ -1,2 +1 @@
-git+http://github.com/enthought/traits.git#egg=traits
-git+http://github.com/enthought/traitsui.git#egg=traitsui
+git+http://github.com/enthought/traitsui.git@release/7.0#egg=traitsui
```
kitchoi, line 1: Do we still want to pin this version on master?

corranwebster: Good catch.  It doesn't hurt right now, but we should fix.

PR #503, opened 2020-04-24
Update documentation for 7.0.0 release

corranwebster:
Update the documentation for 7.0.0 and move old documentation to release-6.1.0 subdirectory.


_sources/api/pyface.util.rst.txt
```diff
@@ -2,26 +2,22 @@ pyface.util package
...
+   :show-inheritance:
 
 Subpackages
 -----------
 
 .. toctree::
 
-    pyface.util.grid
-    pyface.util.tests
+   pyface.util.tests
```
kitchoi, line 14: Aha apparently we have been including test cases in the API reference:
https://docs.enthought.com/pyface/api/pyface.action.tests.test_action_controller.html

Given this is an existing problem, we should open an issue for fixing this programmatically and not have this block the release.

PR #506, opened 2020-04-24
Fix size and position of line status markers in CodeWidget

### LINKED ISSUE #505: CodeWidget shows status markers in wrong size and location
`CodeWidget` uses a fixed 3-pixel height and incorrect scaling for the right-side status markers

bad-gutter

ffishburn:
Closes #505. Fixes the size and position of line status markers in `CodeWidget`


corranwebster:
See comments on the main issue.  I don't think we want to make this change.

Issue #510, opened 2020-04-27
Drop support for PyQt4 and PySide


corranwebster:
PySide is not compatible with any Python version supported by Traits.  PyQt4 is, but has been EOL since 2018.

We should drop support for these.

PR #515, opened 2020-04-30
Use unittest discover for running tests

### LINKED ISSUE #514: Stop using Nose
The Nose test runner is no longer supported.  Ideally we would like to move to using `unittest discover`.  Nose is not imported anywhere in the test suite, but we do run it with `--exclude` to force the omission of non-relevant toolkit-specific packages.

This means that if we run as-is, we get a lot of spurious test failures, eg. where qt tests are attempted with the wx backend.

To make the move successful we either need to make it so that the toolkit-specific tests are skipped when the wrong toolkit is used for running the tests, or re-organize the test suites into general and toolkit-specific packages and handle accordingly.

ievacerny:
Closes #514 

The PR adds a custom test loader functions that is used by `unittest discover` which allows excluding tests that match with a regex pattern specified through environment variable `EXCLUDE_TESTS`.

This is an attempt to simulate the behaviour of `--exclude` flag in nosetests.

Caveat: unittest attempts to import all test modules during loading, therefore, it will try to import test modules in `ui/qt4` when running tests for `wx`. The import will fail and the loader will return a `unittest.loader._FailedTest` which has the path to the test module in its ID. Because the path contains excluded pattern `qt` the test will be filtered out.


ievacerny:
I've been running the test suite in many different ways to check if the behaviour of `unittest` and `nose` is the same, but I'm struggling to get consistent results.

It might be due to test interactions. Investigating it now but it might take a while as it's not an easy thing to debug.

pyface/util/testing.py
```diff
@@ -8,6 +8,21 @@
...
+
+def filter_tests(test_suite, exclusion_pattern):
+    filtered_test_suite = TestSuite()
+    for item in test_suite:
+        if isinstance(item, TestSuite):
+            filtered = filter_tests(item, exclusion_pattern)
+            filtered_test_suite.addTest(filtered)
+        else:
+            match = re.search(exclusion_pattern, item.id())
+            if match is None:
```
corranwebster, line 23: It would be nice to be able to record excluded tests (perhaps by creating fake skipped tests with the same ID), as that would help guard against spurious matches, but this is OK.

pyface/__init__.py
```diff
@@ -26,3 +26,43 @@
...
+    pattern : str
+        An inclusion pattern used to match test files (test*.py default)
+
+    Returns
+    -------
+    filtered_package_tests : unittest.TestSuite
+        TestSuite representing all package tests that did not match specified
+        exclusion pattern.
+    """
+    this_dir = dirname(__file__)
```
corranwebster, line 56: Perhaps do an `import pyface.toolkit` here to ensure that we have the right toolkit up and running before we try and import anything else.  This should ensure the correct tests actually run.

corranwebster:
Wx test failures are because:

1. the linux wheel referenced in etstool.py is no longer there, and we need another one
2. pip is installing wxPython 4.1 on mac and it has introduced test failures

pyface/util/testing.py
ievacerny: I implemented this, but the module of skipped tests usually has import errors so they are logged a bit differently:
```
pyface.ui.qt4.workbench.tests.test_workbench_window_layout (unittest.loader._FailedTest) ... skipped 'Test excluded via pattern qt'
test_get_column_count (pyface.ui.wx.grid.tests.test_composite_grid_model.CompositeGridModelTestCase) ... ok
```

pyface/__init__.py
ievacerny: Added the import.

>It might even be possible to use the pyface.toolkit value to select the filter rather than having to rely on an environment variable (but maybe still have the choice overridable by an envirionment variable for safety, if you do this). But isn't needed for this PR.

I tried this out but I think setting the exclusion pattern in `load_tests` makes it very easy to forget that this feature exists. I think it's better if all the configuration for the test run (including exclusion pattern) stays in `etstool.py` where it's easy to find.

corranwebster:
With #510 merged, this needs a merge of master which will hopefully then make the tests pass.

ievacerny:
The tests on Travis seem to work fine so I'm merging this PR...

PR #522, opened 2020-05-28
Timed messages in StatusBar

jonathanrocher:
This PR exposes the ability to control the duration of a StatusBar message (qt only for now). I am creating this PR to start the discussion:
- Would others agree that this is desirable? 
- How to handle Wx since its `wx.StatusBar` doesn't expose the same ability? What's `pyface`'s policy for exposing toolkit specific capabilities? If that's not acceptable, a similar behavior could be implemented using a `wx.Timer`: https://stackoverflow.com/questions/23188042/wxpython-statusbar-temporary-text . That feels like a lot of complexity brought into pyface though.

### DIFF
--- a/pyface/ui/qt4/action/status_bar_manager.py
+++ b/pyface/ui/qt4/action/status_bar_manager.py
@@ -17,7 +17,7 @@
 from pyface.qt import QtGui
 
 
-from traits.api import Any, Bool, HasTraits, Int, List, Property, Str
+from traits.api import Any, Bool, Float, HasTraits, List, Property, Str
 
 
 class StatusBarManager(HasTraits):
@@ -38,8 +38,8 @@ class StatusBarManager(HasTraits):
     # Whether the status bar is visible.
     visible = Bool(True)
 
-    # Number of millisecond to display new messages for [default: indefinitely]
-    message_duration = Int
+    # Number of seconds to display new messages for [default: indefinitely]
+    message_duration_sec = Float
 
     # ------------------------------------------------------------------------
     # 'StatusBarManager' interface.
@@ -52,11 +52,7 @@ class StatusBarManager(HasTraits):
             self.status_bar = QtGui.QStatusBar(parent)
             self.status_bar.setSizeGripEnabled(self.size_grip)
             self.status_bar.setVisible(self.visible)
-
-            if len(self.messages) > 1:
-                self._show_messages()
-            else:
-                self.status_bar.showMessage(self.message)
+            self._show_messages()
 
         return self.status_bar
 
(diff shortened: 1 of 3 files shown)


pyface/ui/qt4/action/status_bar_manager.py
```diff
@@ -126,4 +129,5 @@ def _show_messages(self):
         # decide to put all but the first message into separate widgets.  We
         # probably also need to extend the API to allow a "message" to be a
         # widget - depends on what wx is capable of.
-        self.status_bar.showMessage("  ".join(self.messages))
+        self.status_bar.showMessage("  ".join(self.messages),
+                                    msecs=self.message_duration)
```
corranwebster, line 133: I think duration in seconds would be better.  That is what `pyface.timer` uses.

corranwebster (commented):
This generally looks fine as an addition.  I would prefer to measure time in seconds and then convert to integer milliseconds at the toolkit level.

For the wx case, we have timers exposed in a toolkit-independent fashion in `pyface.timer`, so it should be fairly easy to add the capability to remove after a fixed amount of time with a toolkit-independent timer object.

pyface/ui/qt4/action/status_bar_manager.py
jonathanrocher: Changed.

jonathanrocher:
This is ready for a second look. WX support added, implementation redone for Qt, unit test added.

jonathanrocher:
Test failure only happens with `pyqt`, not `pyqt5` and is triggered by the following exception in the listener:
```
Exception occurred in traits notification handler.
Please check the log file for details.
Exception occurred in traits notification handler for object: <pyface.ui.qt4.action.status_bar_manager.StatusBarManager object at 0x127802888>, trait: messages, old value: ['goodbye world'], new value: []
  File "/Users/jrocher/.edm/envs/bootstrap3/lib/python3.6/site-packages/traits/trait_notifiers.py", line 381, in __call__
    self.handler(*args)
  File "/Users/jrocher/Projects/ETS_source/pyface/pyface/ui/qt4/action/status_bar_manager.py", line 100, in _messages_changed
    self._show_messages()
  File "/Users/jrocher/Projects/ETS_source/pyface/pyface/ui/qt4/action/status_bar_manager.py", line 142, in _show_messages
    self.status_bar.showMessage("  ".join(self.messages))
RuntimeError: wrapped C/C++ object of type QStatusBar has been deleted
```
Can't make sense of it. Any suggestions?

kitchoi:
If the UI is closed before the timer gets to run its action, then when the timer's action is run, it would be accessing deleted objects. I think in the `remove_status_bar` method, we should stop and remove the timer if there is one. See if that resolves the error you saw?

jonathanrocher:
Thanks for the suggestion @kitchoi . Didn't help with the test failure, but good suggestion anyway, so added.

PR #525, opened 2020-06-01
Initial implementation of new table and tree widget

corranwebster:
This is a first, sketchy cut of the proposed new table and tree widget system, incomplete, but sufficient to display a table or tree structure.

The current status is:

- the index manager design and implementation is essentially complete
- the core methods of the data model are there, but a lot needs to be added
- there is scaffolding for the widget and toolkit code, but it is just enough to prove the concept

### DIFF
--- a/pyface/data_view/abstract_data_model.py
+++ b/pyface/data_view/abstract_data_model.py
@@ -1,0 +1,180 @@
+# (C) Copyright 2005-2020 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+"""
+Abstract Data Model
+===================
+
+This module provides an ABC for all data view data models.  This specifies
+the API that the data view widgets expect, and which the underlying
+data is adapted to by the concrete implementations.  Data models are intended
+to be toolkit-independent
+"""
+from abc import abstractmethod
+
+from traits.api import ABCHasStrictTraits, Event, Instance
+
+from .index_manager import AbstractIndexManager
+
+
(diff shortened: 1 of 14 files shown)


kitchoi:
In https://github.com/enthought/ets/pull/34/files, there is this mention of the limitation of TabularEditor:

> Perhaps most importantly, the TabularEditor does not permit the use
> of different styles of editors in the cells.

At the end of this document, there is a mention of using combobox for `Enum` and spinbox for `Range` in a cell:
> 
>     class Song(HasStrictTraits):
>         name = Str()
>         artist = Str()
>         album = Str()
>         genre = Enum(GENRES)
>         duration = Float()
>         rating = Range(0, 5)
> 
>     class Collection(HasStrictTraits):
>         songs = List(Instance(Song))
> 
>         view = View(
>             Item('songs', editor=DataViewEditor())
>         )
> 
> This should display a flat table with columns for each editable trait of the
> Song class in arbitary order.  This may require some effort in the Editor
> class to introspect the object being edited to select an appropriate model and
> configure it.  This should ideally include using a combobox editor for genre,
> a text editor with float conversion for duration, and an integer spinbox for
> rating.

Do you already have an idea for how the combobox / spinbox / anything else going to fit in this new table and tree widget?

pyface/data_view/abstract_data_model.py
```diff
@@ -0,0 +1,284 @@
...
+        """
+        # XXX needs to be a richer object
+        # eg should have bounds for spinbox, choices for Combo etc.
+        return "text"
+
+    @abstractmethod
+    def get_cell_editor(self, row, column):
+        """ Return editor information for the row and column.
+        """
+        return self.get_column_type(column)
```
kitchoi, line 247: > Do you already have an idea for how the combobox / spinbox / anything else going to fit in this new table and tree widget?

These two look like the relevant methods. But that's about as much as this goes. We probably need QAbstractItemDelegate for these, right?

corranwebster:
I think everything that was sketched here is now implemented.  Closing.

Issue #258, opened 2017-01-11
Audit for signal connections which need to be disconnected


corranwebster:
This line is worrying, in particular: https://github.com/enthought/pyface/blob/master/pyface/ui/qt4/workbench/split_tab_widget.py#L52

But there may be other places where we need to carefully disconnect explicitly connected signal handlers to avoid cycles.

kitchoi:
Offline discussion: We will limit this issue to Qt only, and open a separate one for Wx.

ievacerny:
For reference, the line from the original issue description (code has changed a bit since then): https://github.com/enthought/pyface/blob/2e5fc045dc8d022e519bd654b3ad72b5006f6c84/pyface/ui/qt4/workbench/split_tab_widget.py#L52


Audit done on most of Qt code (PRs #536, #537, #540, #541, #542). Two subpackages that remain unaudited - `wizard` and `workbench`.

Issue #517, opened 2020-05-11
Sporadic test_array_to_qimage_rgb failure using pyside2


ievacerny:
The test sporadically fails both on Travis and during a manual test suite run ~after switching to unittest runner (#515)~ (edit: happens with both unittest and nose runners). I haven't managed to reproduce the error when running just that single test or the test module, seems to only happen when the whole test suite is run. 

For some reason the following pixels have different values than expected (`0xff4488cc`):
```
i=0, j=0: 0xff000000
i=1, j=0: 0xff000000
i=2, j=0: value differs between failures
i=4, j=0: 0xff000000
i=5, j=0: 0xff000000
i=6, j=0: value differs between failures
```

<details>
<summary>Non-informative failure traceback:</summary>

```
FAIL: test_array_to_qimage_rgb (pyface.ui.qt4.util.tests.test_image_helpers.TestImageHelpers)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/travis/.edm/envs/pyface-test-3.6-pyside2/lib/python3.6/site-packages/pyface-7.1.0.dev10-py3.6.egg/pyface/ui/qt4/util/tests/test_image_helpers.py", line 79, in test_array_to_qimage_rgb
    for i in range(32) for j in range(64)
AssertionError: False is not true
```
</details>

corranwebster:
That's really odd.  There really shouldn't be anything more than a (channel-shuffled) memory copy for these pixel values.

I'm tempted to say that this is a Qt/PySide bug, but maybe I'm missing something.

corranwebster:
Idea for debugging these issues: re-write the test to save the `QImage` out if it doesn't match what we expect so we can visually inspect it.

ievacerny:
Here's the image from a failed test:

![qimage](https://user-images.githubusercontent.com/25366751/81679839-6c149380-944a-11ea-9468-a6447fee1edc.png)

It should be all blue, but some pixels at the top are different. Strangely enough, the pixels in the image don't seem to match printed out hex values.

I thought that maybe it is another test interaction brought upon by the change in test order, but I managed to replicate the error with nose runner, so it is most likely an issue with Qt/PySide.

ievacerny:
I think I found the reason behind the failure. It's not a Qt bug as the behaviour is mentioned in the [documentation](https://doc.qt.io/qt-5/qimage.html#QImage-6):
[...]

PR #536, opened 2020-06-10
Audit qt dialogs connections

ievacerny:
Part of #258 

Adds additional cleanup (disconnecting qt signals) to `destroy` methods of some qt dialogs.

Also adds one additional smoke test for `ProgressDialog`.

### DIFF
--- a/pyface/ui/qt4/dialog.py
+++ b/pyface/ui/qt4/dialog.py
@@ -153,10 +153,9 @@ class Dialog(MDialog, Window):
     # -------------------------------------------------------------------------
 
     def destroy(self):
-        if self.control is not None:
-            while self._connections_to_remove:
-                signal, handler = self._connections_to_remove.pop()
-                signal.disconnect(handler)
+        while self._connections_to_remove:
+            signal, handler = self._connections_to_remove.pop()
+            signal.disconnect(handler)
 
         super(Dialog, self).destroy()
 
--- a/pyface/ui/qt4/dialog.py
+++ b/pyface/ui/qt4/dialog.py
@@ -15,8 +15,9 @@
 from pyface.qt import QtCore, QtGui
 
 
-from traits.api import Bool, Enum, Int, provides, Str
-
+from traits.api import (
+    Any, Bool, Callable, Enum, Int, List, provides, Str, Tuple
+)
 
 from pyface.i_dialog import IDialog, MDialog
 from pyface.constant import OK, CANCEL, YES, NO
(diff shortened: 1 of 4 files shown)


ievacerny:
Something strange is happening with the tests. The first 2 runs failed because of #517. Third time I saw the same failure while the job was still running (very slowly) but now that the job is done (timed out) I no longer see the failure.

pyface/ui/qt4/progress_dialog.py
```diff
@@ -97,6 +97,20 @@ def close(self):
 
         super(ProgressDialog, self).close()
 
+    # -------------------------------------------------------------------------
+    # 'IWidget' interface.
+    # -------------------------------------------------------------------------
+
+    def destroy(self):
+        if self.control is not None:
+            buttons = self.control.findChild(QtGui.QDialogButtonBox)
```
kitchoi, line 106: I agree that this is smelly, could we hold a reference to the button created by this class and use that, and then remove the reference in `destroy`.

ievacerny:
I don't think this PR is doing anything worse, quite likely it's just causing imbalance between other test interactions. So I don't think #517 shouldn't be a blocker for this and I suggest skipping the failing test until #517 is fixed.

ievacerny:
> > Something strange is happening with the tests. The first 2 runs failed because of #517. Third time I saw the same failure while the job was still running (very slowly) but now that the job is done (timed out) I no longer see the failure.
> 
> I don't think this PR is doing anything worse, quite likely it's just causing imbalance between other test interactions. So I don't think #517 shouldn't be a blocker for this and I suggest skipping the failing test until #517 is fixed.

The tests finally passed, so I take my suggestion back. The problematic tests can still be skipped until the issue is fixed but that no longer needs to be done in this PR

pyface/ui/qt4/dialog.py
```diff
@@ -138,6 +148,18 @@ def _show_modal(self):
         retval = dialog.exec_()
         return _RESULT_MAP[retval]
 
+    # -------------------------------------------------------------------------
+    # 'IWidget' interface.
+    # -------------------------------------------------------------------------
+
+    def destroy(self):
+        if self.control is not None:
```
kitchoi, line 156: If the list is nonempty, it would be best to disconnect them than keeping the references.
(Likewise for the other method)

kitchoi:
Oops. The build was definitely green when this PR was merged. That was my mistake!  I thought I was rebuilding master but I actually rebuilt the CI job for *the PR*, which then failed because the branch was deleted.

kitchoi:
Let me see if restoring the branch could resurrect the history and mask my silly mistake...

kitchoi:
Nope, because the merge commit is gone. Not sure how I can put the history right. My apologies!

PR #537, opened 2020-06-10
Disconnect qt PythonEditor downstream widget signals

ievacerny:
Part of #258 

Adds `disconnect_event_listeners` methods to downstream widgets which are called from within `_remove_event_listeners` of `PythonEditor`. 

Right after `_remove_event_listeners` a `deleteLater` is called on control, which according to qt documentation should disconnect all signals of downstream objects, so this might not be needed.

### DIFF
--- a/pyface/ui/qt4/code_editor/replace_widget.py
+++ b/pyface/ui/qt4/code_editor/replace_widget.py
@@ -36,6 +36,11 @@ class ReplaceWidget(FindWidget):
 
         self.setLayout(layout)
 
+    def disconnect_event_listeners(self):
+        close_button = self.findChild(QtGui.QPushButton)
+        if close_button is not None:
+            close_button.clicked.disconnect(self.hide)
+
     def _create_replace_control(self):
         control = QtGui.QWidget(self)
 
--- a/pyface/ui/qt4/code_editor/replace_widget.py
+++ b/pyface/ui/qt4/code_editor/replace_widget.py
@@ -38,7 +38,7 @@ class ReplaceWidget(FindWidget):
 
     def disconnect_event_listeners(self):
         close_button = self.findChild(QtGui.QPushButton)
-        if close_button is not None:
+        if close_button is not None and close_button.text() == "Close":
             close_button.clicked.disconnect(self.hide)
 
     def _create_replace_control(self):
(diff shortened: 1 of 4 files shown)


corranwebster:
This looks fine, although probably not necessary.  The code where you are finding the button is a but of a smell.  We should be storing the close button as an attribute, but that's orthogonal to this PR.

ievacerny:
In this case all signals are connected to methods in main `control` hierarchy, so as long as `control` is deleted, the signals should be disconnected as well according to qt documentation. The benefits of adding `disconnect` are very small (or completely non-existent).

It's quite straightforward and safe (I think) to disconnect most of them. Since the work is already done, it wouldn't hurt to add this.

The exception is the `close_button` which has to be found between the children of a control. As pointed out by @corranwebster, this is a smell and it is risky because someone else might have added another "Close" button and connected it to something else. Existing `disconnect` would then throw an exception.

A few different approaches to minimise the risk regarding the `close_button`:
- In a different PR save a reference to the `close_button` and then disconnect using that reference (leave this PR open until then)
- Wrap `disconnect` of `close_button`  in `try-except`
- Do not disconnect the signal of `close_button`

pyface/ui/qt4/code_editor/replace_widget.py
```diff
@@ -36,6 +36,11 @@ def __init__(self, parent):
 
         self.setLayout(layout)
 
+    def disconnect_event_listeners(self):
+        close_button = self.findChild(QtGui.QPushButton)
+        if close_button is not None and close_button.text() == "Close":
+            close_button.clicked.disconnect(self.hide)
+
```
kitchoi, line 43: I thought we were going to store the reference to the button and then get rid of it like in the other PRs. 
If we can't do this confidently, let's not do it then.
I am sorry I assumed we resolved this in the other PRs and that could be extended here. I should have commented about it again while this was a draft.

ievacerny: I don't think `_connections_to_remove` would work here because `ReplaceWidget.__init__()` calls `FindWidget.__init__()` which sets up a button in the layout that is then overwritten by the `ReplaceWidget` layout. Unless we save only one `connection_to_remove` that is also overwritten in `ReplaceWidget.__init__()`.

kitchoi: OK, that looks complicated. Let's not do that then.

ievacerny: Went with the 3rd approach and removed `close_button` disconnects completely

PR #540, opened 2020-06-12
Disconnect qt PythonShell downstream widget signals

ievacerny:
Part of #258

Adds `disconnect_event_listeners` methods to downstream widgets which are called from within `_remove_event_listeners` of `PythonShell`.

Right after `_remove_event_listeners` a `deleteLater` is called on control. In this case all signals are connected to methods in main `control` hierarchy, so as long as control is deleted, the signals should be disconnected as well according to qt documentation. The benefits of adding disconnect are very small (or completely non-existent).

It's quite straightforward and safe (I think) to disconnect most of the signals here. And since the work is already done, it wouldn't hurt to add this.

### DIFF
--- a/pyface/ui/qt4/console/console_widget.py
+++ b/pyface/ui/qt4/console/console_widget.py
@@ -349,6 +349,10 @@ class ConsoleWidget(QtGui.QWidget):
 
         return QtCore.QSize(width, height)
 
+    # ---------------------------------------------------------------------------
+    # 'ConsoleWidget' public interface
+    # ---------------------------------------------------------------------------
+
     def can_copy(self):
         """ Returns whether text can be copied to the clipboard.
         """
--- a/pyface/ui/qt4/console/console_widget.py
+++ b/pyface/ui/qt4/console/console_widget.py
@@ -122,11 +122,6 @@ class ConsoleWidget(QtGui.QWidget):
         [QtCore.Qt.Key_C, QtCore.Qt.Key_G, QtCore.Qt.Key_O, QtCore.Qt.Key_V]
     )
 
-    #: A list of connected Qt signals to be removed before destruction.
-    #: First item in the tuple is the Qt signal. The second item is the event
-    #: handler.
-    _connections_to_remove = []
-
     # ---------------------------------------------------------------------------
     # 'QObject' interface
     # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


pyface/ui/qt4/python_shell.py
```diff
@@ -118,9 +118,11 @@ def _add_event_listeners(self):
     def _remove_event_listeners(self):
         if self.control is not None:
             # Disconnect signals for events.
-            self.control.executed.connect(self._on_command_executed)
+            self.control.executed.disconnect(self._on_command_executed)
             self._event_filter.signal.disconnect(self._on_obj_drop)
 
+            self.control.disconnect_event_listeners()
+
```
kitchoi, line 125: Nit:  Rename `disconnect_event_listeners` to `_remove_event_listeners`?

pyface/ui/qt4/console/console_widget.py
```diff
@@ -340,6 +340,28 @@ def sizeHint(self):
...
+        control = self._control
+        control.cursorPositionChanged.disconnect(self._cursor_position_changed)
+        control.customContextMenuRequested.disconnect(
+            self._custom_context_menu_requested
+        )
+        control.copyAvailable.disconnect(self.copy_available)
+        control.redoAvailable.disconnect(self.redo_available)
+        control.undoAvailable.disconnect(self.undo_available)
+        layout = control.document().documentLayout()
+        layout.documentSizeChanged.disconnect(self._adjust_scrollbars)
```
kitchoi, line 363: There are quite a lot of signals, and it requires keeping some text between this method and `__init__` in sync. The `_connections_to_remove` pattern might be more maintainable and easier to confirm it is right?

ievacerny: I did switch to use `_connections_to_remove` but without the use of Traits. The module wasn't using anything from Traits before and I wasn't sure if it's a good idea to bring it in.

pyface/ui/qt4/python_shell.py
ievacerny: I also moved the methods closer to `__init__` and removed the "public interface" headers

pyface/ui/qt4/console/console_widget.py
```diff
@@ -122,6 +122,11 @@ class ConsoleWidget(QtGui.QWidget):
         [QtCore.Qt.Key_C, QtCore.Qt.Key_G, QtCore.Qt.Key_O, QtCore.Qt.Key_V]
     )
 
+    #: A list of connected Qt signals to be removed before destruction.
+    #: First item in the tuple is the Qt signal. The second item is the event
+    #: handler.
+    _connections_to_remove = []
```
kitchoi, line 128: I did not read the scope careful enough. This instantiation needs to live inside `__init__`, otherwise we get one list that is shared among all instances of `ConsoleWidget`.

ievacerny:
The document layout doesn't stay the same throughout console widget lifecycle, so a reference to it cannot be saved in `self._connections_to_remove`. Since that connection is some kind of Qt hack, I decided to leave the connection alone (added a comment as an explanation).

PR #541, opened 2020-06-12
Disconnect trait change handlers from downstream ApplicationWindow widgets

ievacerny:
Part of #258 

Before destroying control of `ApplicationWindow`, trait change handlers from downstream widgets are disconnected.

The structure of these underlying widgets and their managers is fairly complex so it's hard to say if it is necessary to disconnect these trait change handlers and if it helps with garbage collection. I attempt to evaluate the risks and the benefits but there might be things that I've missed.
____
**`_Menu` in `menu_manager.py` and `_ToolBar` in `tool_bar_manager.py`**

They connect their methods to Python `MenuManager` and `ToolBarManager` object traits, but as far as I know that shouldn't prevent any of the objects from being garbage collected. And when either side is deleted, the connection should break. However, if for some reason those connections remain after an attempt to destroy the whole control hierarchy, they might cause problems because there are no safeguards in the handlers.

This PR removes `on_trait_change` connections, however, to do so it has to find these objects in the children of the control, since no proper reference is held. The code also only checks that the objects is an instance of expected Qt object, not the custom `_Menu` or `_ToolBar` because those are private classes. This is risky because there might be other `QMenu` and `QToolBar` objects that do not have expected `disconnect_event_listeners` method. One solution would be to wrap these in a `try-except`.

I'm open to suggestion on how to better approach this situation and if the potential benefits are worth the risk.
____
**`_MenuItem` and `_Tool` in `action_item.py`**

All handlers in `action_item.py` have protective `if self.control is not None` check so from code inspection it seems that leaving these handlers here shouldn't cause problems. But a disconnect structure was already implemented for `_MenuItem` in 80855481982ce433324123, which suggests that not removing the handlers was actually causing problems. Although it was only executed on refresh, not on destruction.

This PR connects that structure from `_MenuItem` to the main `destroy` method. The PR doesn't add a similar structure for `_Tool` because (1) its `clear` method behaves differently, (2) it would introduce more changes outside of disposal code. Instead trait change handlers of `_Tool` are disconnected in a slot that is connected to the `destroyed` signal of its control. I would say that the code added here is low risk because it disconnects `on_trait_change` from Python objects. And if the code from the section above doesn't work out, the disconnect in `_MenuItem` can also happen in a slot connected to `destroyed` signal.

There is a qt connection to `_Tool.control.triggered` signal that is not currently disconnected on destruction because there's no good place to do so without adding a lot of extra code. The slot does work with UI components and doesn't have a `control is None` check so it has some potential of causing problems. But I think a better (lower risk) approach would be to add this check to the slot rather than adding additional structure in order to disconnect it.

**Note1:** There are also 2 qt signal-slot connections that are connected to `destroyed` signal. This PR doesn't disconnect them because they shouldn't be disconnected.

**Note2:** `_PaletteTool` in `action_item.py` hasn't been touched because the support for it has not been implemented in qt.

### DIFF
--- a/pyface/ui/qt4/action/menu_manager.py
+++ b/pyface/ui/qt4/action/menu_manager.py
@@ -42,6 +42,11 @@ class MenuManager(ActionManager, ActionManagerItem):
     # The default action for tool button when shown in a toolbar (Qt only)
     action = Instance(Action)
 
+    # Private interface ---------------------------------------------------#
+
+    #: Keep track of all created menus in order to properly dispose of them
+    _menus = []
+
     # ------------------------------------------------------------------------
     # 'MenuManager' interface.
     # ------------------------------------------------------------------------
@@ -56,7 +61,19 @@ class MenuManager(ActionManager, ActionManagerItem):
         if controller is None:
             controller = self.controller
 
-        return _Menu(self, parent, controller)
+        menu = _Menu(self, parent, controller)
+        self._menus.append(menu)
+
+        return menu
+
+    # ------------------------------------------------------------------------
+    # 'ActionManager' interface.
+    # ------------------------------------------------------------------------
+
+    def destroy(self):
+        while self._menus:
+            menu = self._menus.pop()
+            menu._remove_event_listeners()
 
     # ------------------------------------------------------------------------
     # 'ActionManagerItem' interface.
@@ -131,11 +148,7 @@ class _Menu(QtGui.QMenu):
 
         return
 
-    # ------------------------------------------------------------------------
-    # '_Menu' interface.
-    # ------------------------------------------------------------------------
-
-    def disconnect_event_listeners(self):
+    def _remove_event_listeners(self):
         self._manager.on_trait_change(self.refresh, "changed", remove=True)
         self._manager.on_trait_change(
             self._on_enabled_changed, "enabled", remove=True
@@ -151,6 +164,10 @@ class _Menu(QtGui.QMenu):
         )
         # Removes event listeners from downstream menu items
         self.clear()
+
+    # ------------------------------------------------------------------------
+    # '_Menu' interface.
+    # ------------------------------------------------------------------------
 
     def clear(self):
         """ Clears the items from the menu. """
--- a/pyface/ui/qt4/action/menu_manager.py
+++ b/pyface/ui/qt4/action/menu_manager.py
@@ -75,6 +75,8 @@ class MenuManager(ActionManager, ActionManagerItem):
             menu = self._menus.pop()
             menu._remove_event_listeners()
 
+        super(MenuManager, self).destroy()
+
     # ------------------------------------------------------------------------
     # 'ActionManagerItem' interface.
     # ------------------------------------------------------------------------
--- a/pyface/ui/qt4/action/menu_manager.py
+++ b/pyface/ui/qt4/action/menu_manager.py
@@ -73,7 +73,7 @@ class MenuManager(ActionManager, ActionManagerItem):
     def destroy(self):
         while self._menus:
             menu = self._menus.pop()
-            menu._remove_event_listeners()
+            menu.dispose()
 
         super(MenuManager, self).destroy()
 
@@ -150,7 +150,7 @@ class _Menu(QtGui.QMenu):
 
         return
 
-    def _remove_event_listeners(self):
+    def dispose(self):
         self._manager.on_trait_change(self.refresh, "changed", remove=True)
         self._manager.on_trait_change(
             self._on_enabled_changed, "enabled", remove=True
--- a/pyface/ui/qt4/action/menu_manager.py
+++ b/pyface/ui/qt4/action/menu_manager.py
@@ -135,6 +135,23 @@ class _Menu(QtGui.QMenu):
     # '_Menu' interface.
     # ------------------------------------------------------------------------
 
+    def disconnect_event_listeners(self):
+        self._manager.on_trait_change(self.refresh, "changed", remove=True)
+        self._manager.on_trait_change(
+            self._on_enabled_changed, "enabled", remove=True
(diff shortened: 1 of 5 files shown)


ievacerny:
> This is risky because there might be other `QMenu` and `QToolBar` objects that do not have expected `disconnect_event_listeners` method. One solution would be to wrap these in a `try-except`.

And tests are failing on windows because of this... Added `try-except` but it still doesn't feel like a very good solution.

pyface/ui/qt4/application_window.py
```diff
@@ -115,6 +115,28 @@ def _set_window_icon(self):
...
+
+            toolbars = self.control.findChildren(QtGui.QToolBar)
+            for toolbar in toolbars:
+                try:
+                    toolbar.disconnect_event_listeners()
+                except AttributeError:
+                    continue
+
+        super(ApplicationWindow, self).destroy()
+
```
kitchoi, line 139: I agree this is ugly and is not very scalable. If we want to do this, we should do this right.
`pyface.ui.qt4.action.MenuManager.destroy` is called when the application window is destroyed. Currently that method is not implemented and calls the default implementation. I think we should implement that `destroy` method to do the clean up there.  The menu manager might need to hold references to the widget in order to clean up in `destroy`, but those references can be removed in the same `destroy` method. Likewise I think we can add clean up to the `destroy` method of the toolbar manager.

In this `ApplicationWindow`, there is a change handler for `@on_trait_change("tool_bar_manager, tool_bar_managers")`. I think the destroy method applies there too. That may be out-of-scope for this task, but we should open an issue for that. This perhaps supports the idea of implementing the `destroy` method as it would be more reusable in a different context.

ievacerny:
Thank you for the review @kitchoi 

Following the suggestions I changed the approach:
____
**`_Menu` in `menu_manager.py` and `_ToolBar` in `tool_bar_manager.py`**

`MenuManager` and `ToolBarManager` now keep a list of created `_Menu` and `_ToolBar` objects and use that list to call their `_remove_event_listeners` methods (name changed).

One concern I have with this is context and temporary menus (e.g. in [fields.py](https://github.com/enthought/pyface/blob/08ba6c2b156ca8b007214559bb5f9c8160e2c4c5/pyface/ui/qt4/fields/field.py#L62-L66)) might stick around because of those references, but I've done some testing and I haven't seen any changes in behaviour. I would still labels this as a high risk change as I'm not convinced that it's not going to introduce bugs.

____
**`_MenuItem` and `_Tool` in `action_item.py`**

I've implemented a structure similar to the one in `_MenuItem` with `_Tool` but without going through `clear` method. Again, I've done some testing and everything seems to work fine but I'm not convinced that keeping these references won't cause issues, so I'm labelling this as high risk as well.

pyface/ui/qt4/action/menu_manager.py
```diff
@@ -131,6 +150,23 @@ def __init__(self, manager, parent, controller):
...
+            self._on_visible_changed, "visible", remove=True
+        )
+        self._manager.on_trait_change(
+            self._on_name_changed, "name", remove=True
+        )
+        self._manager.on_trait_change(
+            self._on_image_changed, "image", remove=True
+        )
+        # Removes event listeners from downstream menu items
+        self.clear()
```
kitchoi, line 168: Based on the name of this method, I would not expect `clear` to live here. Should this method be called `dispose` instead?

pyface/ui/qt4/action/tool_bar_manager.py
```diff
@@ -168,6 +186,20 @@ def __init__(self, tool_bar_manager, parent):
 
         return
 
+    def _remove_event_listeners(self):
```
kitchoi, line 189: Same here...from the name I'd expect only calls to remove the handlers but this method is doing more than that. Should this be named `dispose` as well?

pyface/ui/qt4/action/menu_manager.py
ievacerny: Should it stay private, i.e. `_dispose`? The comment re privacy was made on another PR but I think it applies here as well.

kitchoi: Right. This class already has a private name so I think we are alright to give this a public name, it would be consistent with a couple of these classes, e.g. `pyface.ui.qt4.qction.action_item._MenuItem.dispose`.

pyface/ui/qt4/action/menu_manager.py
```diff
@@ -42,6 +42,11 @@ class MenuManager(ActionManager, ActionManagerItem):
     # The default action for tool button when shown in a toolbar (Qt only)
     action = Instance(Action)
 
+    # Private interface ---------------------------------------------------#
+
+    #: Keep track of all created menus in order to properly dispose of them
+    _menus = []
```
kitchoi, line 48: I overlooked this. My apologies. This should be `List()`, otherwise we only get one list that all instances of `MenuManager` share.

pyface/ui/qt4/action/tool_bar_manager.py
```diff
@@ -55,6 +55,9 @@ class ToolBarManager(ActionManager):
     # Cache of tool images (scaled to the appropriate size).
     _image_cache = Instance(ImageCache)
 
+    #: Keep track of all created toolbars in order to properly dispose of them
+    _toolbars = []
```
kitchoi, line 59: Same here. I am sorry for overlooking this.

pyface/ui/qt4/action/menu_manager.py
ievacerny: Opened #548

pyface/ui/qt4/action/tool_bar_manager.py
ievacerny: Opened #548

PR #542, opened 2020-06-12
Disconnect qt signals in qt implementation of pyface.tasks

ievacerny:
Part of #258 

Distinct things this PR does:

1. `DockPane` - disconnects signals from its control. Low risk - low reward kind of change. Doesn't hurt to add this.

2. Deletes connections to shortcuts via `_connections_to_remove` in classes providing `IEditorAreaPane`. It is good to disconnect these shortcuts because if something goes wrong and the objects are not deleted, the handlers can still be activated via keyboard events. So the reward is fairly high. The risk with current code is low because it is only disconnecting what is added to the list and there are no other disconnects. The problems might arise if someone decides to disconnect the shortcuts elsewhere and doesn't remove them from the list. But maybe that should be considered a case of "Don't do that".

3. Removes reference to `active_tabwidget` from `SplitEditorAreaPane`. This was preventing the widget from being deleted together with the main control. Low risk - high reward kind of change. Although not directly related to signals (and thus can be separated from this).

4. `EditorAreaPane` - disconnects signals from its control. The handlers do work with UI elements accessible via control but then the signals are emitted from the same control. If control is set to `None` between the signal was emitted and slot was executed, there could be problems. So I would say that this is a low risk - low medium reward kind of change.

5. Removes reference to `active_editor` from `AdvancedEditorAreaPane`. The editor here is a python object so holding a reference to it doesn't really do anything (its control is destroyed separately). But other classes implementing `IEditorAreaPane` are removing this reference via `remove_editor` method, so I added that line of code here as well. No risk - no reward kind of change. Done for consistency.

6. `AdvancedEditorAreaPane` - tells its control to disconnect what it had connected. In this case it had connected a slot to `focusChanged` signal of the main application, so this is important to disconnect. Low risk - high reward kind of change.

5. When `EditorAreaWidget` destroys and `editor_widget`, it also tells the widget to disconnect what it had connected. The signals and slots belong to the same object but in the slot the objects tries to access an attribute on the parent, which might cause problems. Low risk - medium reward.

6. After disconnecting its own signals, `EditorWidget` tells its `titlebarWidget` (`EditorTitleBarWidget`) to disconnect what it had connected. The issue here is that `EditorTitleBarWidget` doesn't keep a reference to the slot it connected its signal to. The slot can be accessed from `EditorWidget` so it is easy to pass that in. However, an assumption has been made here that the `editor` of a certain `editor_widget` doesn't change. Making assumptions here is quite risky and this disconnect doesn't cover the case where new `EditorTitleBarWidget` is set as the `TitleBarWidget` and the old one is discarded. I'm not sure how qt handles disposal in that case but I assume that it destroys the old object and by doing so destroys the connections as well. So given that this added `disconnect` covers only a small case that is probably handled as expected by qt, this is a high risk - low benefit kind of change.

Connections that this PR doesn't remove:

1. In `childEvent` method of `EditorAreaWidget` (control of `Advanced EditorAreaPane`) a couple of slots are connected to a couple of signals of the child. I couldn't find an easy way to disconnect these signals, so I didn't add these changes. But given that this is a child signal connected to a parent slot and both are qt objects, qt should handle this situation well on its own.

2. `DraggableTabWidget` that are children of `SplitEditorAreaPane` control make a couple of connections in their `__init__`. The control can have a tree hierarchy structure with `DraggableTabWidget`s so there's no easy way to remove these connections. But given that both signals and slots belong to the same object, they shouldn't cause any problems.

3. In `ceate_empty_widget` of `DraggableTabWidget` buttons are created and their signals are connected to callbacks provided in top `SplitEditorAreaPane`. Again, there's no easy way to remove these connections. I think these connection can be problematic and would say that removing them would be a medium reward. But I think the code needed to disconnect them would be much riskier. I would like to hear more opinions on this.

### DIFF
--- a/pyface/ui/qt4/tasks/advanced_editor_area_pane.py
+++ b/pyface/ui/qt4/tasks/advanced_editor_area_pane.py
@@ -14,7 +14,10 @@ import sys
 from pyface.qt import QtCore, QtGui
 
 
-from traits.api import DelegatesTo, Instance, on_trait_change, provides
+from traits.api import (
+    Any, Callable, DelegatesTo, Instance, List, on_trait_change, provides,
+    Tuple
+)
 
 
 from pyface.tasks.i_advanced_editor_area_pane import IAdvancedEditorAreaPane
@@ -40,6 +43,11 @@ class AdvancedEditorAreaPane(TaskPane, MEditorAreaPane):
 
     _main_window_layout = Instance(MainWindowLayout)
 
+    #: A list of connected Qt signals to be removed before destruction.
+    #: First item in the tuple is the Qt signal. The second item is the event
+    #: handler.
+    _connections_to_remove = List(Tuple(Any, Callable))
+
     # ------------------------------------------------------------------------
     # 'TaskPane' interface.
     # ------------------------------------------------------------------------
@@ -61,29 +69,48 @@ class AdvancedEditorAreaPane(TaskPane, MEditorAreaPane):
             prev_seq = "Ctrl+PgUp"
         shortcut = QtGui.QShortcut(QtGui.QKeySequence(next_seq), self.control)
         shortcut.activated.connect(self._next_tab)
+        self._connections_to_remove.append(
+            (shortcut.activated, self._next_tab)
+        )
         shortcut = QtGui.QShortcut(QtGui.QKeySequence(prev_seq), self.control)
         shortcut.activated.connect(self._previous_tab)
+        self._connections_to_remove.append(
+            (shortcut.activated, self._previous_tab)
+        )
 
         # Add shortcuts for switching to a specific tab.
         mod = "Ctrl+" if sys.platform == "darwin" else "Alt+"
         mapper = QtCore.QSignalMapper(self.control)
         mapper.mapped.connect(self._activate_tab)
+        self._connections_to_remove.append(
+            (mapper.mapped, self._activate_tab)
+        )
         for i in range(1, 10):
             sequence = QtGui.QKeySequence(mod + str(i))
             shortcut = QtGui.QShortcut(sequence, self.control)
             shortcut.activated.connect(mapper.map)
+            self._connections_to_remove.append(
+                (shortcut.activated, mapper.map)
+            )
             mapper.setMapping(shortcut, i - 1)
 
     def destroy(self):
         """ Destroy the toolkit-specific control that represents the pane.
         """
         self.control.removeEventFilter(self._filter)
+        self.control.disconnect_event_listeners()
         self._filter = None
 
         for editor in self.editors:
             editor_widget = editor.control.parent()
             self.control.destroy_editor_widget(editor_widget)
             editor.editor_area = None
+        self.active_editor = None
+
+        if self.control is not None:
+            while self._connections_to_remove:
+                signal, handler = self._connections_to_remove.pop()
+                signal.disconnect(handler)
 
         super(AdvancedEditorAreaPane, self).destroy()
 
@@ -266,6 +293,11 @@ class EditorAreaWidget(QtGui.QMainWindow):
             QtCore.Qt.AllDockWidgetAreas, QtGui.QTabWidget.North
         )
 
+    def disconnect_event_listeners(self):
+        """ Disconnects focusChanged signal of the application """
+        app = QtGui.QApplication.instance()
+        app.focusChanged.disconnect(self._focus_changed)
+
     def add_editor_widget(self, editor_widget):
         """ Adds a dock widget to the editor area.
         """
@@ -292,6 +324,7 @@ class EditorAreaWidget(QtGui.QMainWindow):
         """
         editor_widget.hide()
         editor_widget.removeEventFilter(self)
+        editor_widget.disconnect_event_listeners()
         editor_widget.editor.destroy()
         self.removeDockWidget(editor_widget)
 
@@ -598,6 +631,13 @@ class EditorWidget(QtGui.QDockWidget):
         self.dockLocationChanged.connect(self.update_title_bar)
         self.visibilityChanged.connect(self.update_title_bar)
 
+    def disconnect_event_listeners(self):
+        self.dockLocationChanged.disconnect(self.update_title_bar)
+        self.visibilityChanged.disconnect(self.update_title_bar)
+        title_bar = self.titleBarWidget()
+        if isinstance(title_bar, EditorTitleBarWidget):
+            title_bar.disconnect_event_listeners(self)
+
     def update_title(self):
         title = self.editor.editor_area._get_label(self.editor)
         self.setWindowTitle(title)
@@ -643,6 +683,10 @@ class EditorTitleBarWidget(QtGui.QTabBar):
         self.setTabsClosable(True)
 
         self.tabCloseRequested.connect(editor_widget.editor.close)
+
+    def disconnect_event_listeners(self, editor_widget):
+        # Assumes that the editor of editor_widget never changes
+        self.tabCloseRequested.disconnect(editor_widget.editor.close)
 
     def mousePressEvent(self, event):
         self.parent().parent()._drag_widget = self.parent()
--- a/pyface/ui/qt4/tasks/advanced_editor_area_pane.py
+++ b/pyface/ui/qt4/tasks/advanced_editor_area_pane.py
@@ -98,7 +98,7 @@ class AdvancedEditorAreaPane(TaskPane, MEditorAreaPane):
         """ Destroy the toolkit-specific control that represents the pane.
         """
         self.control.removeEventFilter(self._filter)
-        self.control.disconnect_event_listeners()
+        self.control._remove_event_listeners()
         self._filter = None
 
         for editor in self.editors:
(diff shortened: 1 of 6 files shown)


pyface/ui/qt4/tasks/advanced_editor_area_pane.py
```diff
@@ -644,6 +684,10 @@ def __init__(self, editor_widget):
 
         self.tabCloseRequested.connect(editor_widget.editor.close)
 
+    def disconnect_event_listeners(self, editor_widget):
+        # Assumes that the editor of editor_widget never changes
+        self.tabCloseRequested.disconnect(editor_widget.editor.close)
```
kitchoi, line 689: I agree this one is riskier: Disconnecting something that isn't connected will cause Qt to raise an error. I'd be in favour of leaving this one out.

pyface/ui/qt4/tasks/advanced_editor_area_pane.py
```diff
@@ -266,6 +293,11 @@ def __init__(self, editor_area, parent=None):
             QtCore.Qt.AllDockWidgetAreas, QtGui.QTabWidget.North
         )
 
+    def disconnect_event_listeners(self):
```
kitchoi, line 296: I saw elsewhere there is this `_remove_event_listeners` which is defined in *other* interfaces, e.g.:
https://github.com/enthought/pyface/blob/f60905affb95cc0024c1a0514da4f4d389e5a533/pyface/i_widget.py#L89-L90

Looks like the intended effects would be the same, should we use the same name? We might want to declare this in the interface of `ITaskPane`.

ievacerny: That interface is used in higher level objects, which would be `AdvancedEditorAreaPane` in this case. I chose a different name in downstream widgets on purpose. Also, the method has to be public so that it can be called from upstream.

pyface/ui/qt4/tasks/editor_area_pane.py
```diff
@@ -82,6 +100,14 @@ def destroy(self):
         for editor in self.editors:
             self.remove_editor(editor)
 
+        if self.control is not None:
+            self.control.currentChanged.disconnect(self._update_active_editor)
+            self.control.tabCloseRequested.disconnect(self._close_requested)
```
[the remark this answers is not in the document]
ievacerny: They can be easily disconnected without adding to the list. And since in other panes the list was used for shortcut connections only, I left the list here to be for shortcut connections only as well.

pyface/ui/qt4/tasks/advanced_editor_area_pane.py
kitchoi: I think we have different levels of restrictions we want to impose: If the method does not have a preceding underscore, and the class name does not have a preceding underscore, and the module name does not have a preceding underscore, that is at the level of accessibility that we need to maintain backward compatibility for, and therefore it would impose limitations on how the code can change. Python is not like Java where there are more than one modifiers for imposing access control enforced by the compiler, so we have to resort to naming hints, comments etc.  I am not sure this `disconnect_event_listeners` is intended for consumption outside of pyface.

My main concern is the confusion this could cause to have two methods that have the same meaning but named differently.

kitchoi: > My main concern is the confusion this could cause to have two methods that have the same meaning but named differently.

Another equally "main" concern is having a publicly named method that will be harder to change because we cannot be sure no one outside of pyface has used it.

ievacerny: Changed the name of the method to `_remove_event_listeners`

pyface/ui/qt4/tasks/editor_area_pane.py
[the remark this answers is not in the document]
ievacerny: Added them to the list

pyface/ui/qt4/tasks/advanced_editor_area_pane.py
```diff
@@ -61,29 +69,48 @@ def create(self, parent):
...
+        self.control._remove_event_listeners()
         self._filter = None
 
         for editor in self.editors:
             editor_widget = editor.control.parent()
             self.control.destroy_editor_widget(editor_widget)
             editor.editor_area = None
+        self.active_editor = None
+
+        if self.control is not None:
```
kitchoi, line 110: The first line in this function already accesses `control.*` so maybe this `if self.control is not None` is not needed? (See another comment too).

pyface/ui/qt4/tasks/editor_area_pane.py
```diff
@@ -82,6 +106,11 @@ def destroy(self):
         for editor in self.editors:
             self.remove_editor(editor)
 
+        if self.control is not None:
```
kitchoi, line 109: Same here, `self.control.removeEventFilter` is already accessing `control`'s attribute. On the other hand, for the `_connections_to_remove` pattern, I would think `if self.control is not None` is the sort of defensive programming that could backfire...e.g. if the block is not run because of an error that should not have occurred, the `_connections_to_remove` could be holding references that prevent the objects from being garbage collected, leading to even more obscure errors.

pyface/ui/qt4/tasks/split_editor_area_pane.py
```diff
@@ -127,6 +134,15 @@ def destroy(self):
         for editor in self.editors[:]:
             self.remove_editor(editor)
 
+        if self.control is not None:
```
kitchoi, line 137: Same here. The connections actually have nothing to do with `control` so if the list is nonempty, we should disconnect all of them.

kitchoi:
Looks like we have disturbed the balance of the forces again... :(

There are a number of error traceback following some of the tests that have nothing to do with any of the widgets, e.g.:
https://github.com/enthought/pyface/blob/0104cf7c17114b7022566d63fe52b43469b67e1d/pyface/ui/qt4/util/tests/test_gui_test_assistant.py#L87-L89

```
test_assert_eventually_true_in_gui_already_true (pyface.ui.qt4.util.tests.test_gui_test_assistant.TestGuiTestAssistant) ... Traceback (most recent call last):
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface-7.1.0.dev17-py3.6.egg/pyface/ui/qt4/gui.py", line 176, in event
    self._callable(*self._args, **self._kw)
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface-7.1.0.dev17-py3.6.egg/pyface/i_window.py", line 233, in close
    self.destroy()
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface-7.1.0.dev17-py3.6.egg/pyface/tasks/task_window.py", line 95, in destroy
    self._window_backend.destroy()
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface-7.1.0.dev17-py3.6.egg/pyface/ui/qt4/tasks/task_window_backend.py", line 56, in destroy
    app.focusChanged.disconnect(self._focus_changed_signal)
TypeError: 'method' object is not connected
ok
``` 
Those tests did not fail, though. :/

And there are a couple that actually register as test failure and cause the CI build to fail:
```
ERROR: test_react_to_dock_pane_removed (pyface.tasks.tests.test_dock_pane_toggle_group.DockPaneToggleGroupTestCase)
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface-7.1.0.dev17-py3.6.egg/pyface/tasks/tests/test_dock_pane_toggle_group.py", line 92, in tearDown
    self.window.destroy()
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface-7.1.0.dev17-py3.6.egg/pyface/tasks/task_window.py", line 101, in destroy
    self._destroy_state(state)
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface-7.1.0.dev17-py3.6.egg/pyface/tasks/task_window.py", line 364, in _destroy_state
    state.menu_bar_manager.destroy()
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface-7.1.0.dev17-py3.6.egg/pyface/action/action_manager.py", line 178, in destroy
    group.destroy()
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface-7.1.0.dev17-py3.6.egg/pyface/action/group.py", line 130, in destroy
    item.destroy()
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface-7.1.0.dev17-py3.6.egg/pyface/ui/qt4/action/menu_manager.py", line 76, in destroy
    menu.dispose()
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface-7.1.0.dev17-py3.6.egg/pyface/ui/qt4/action/menu_manager.py", line 168, in dispose
    self.clear()
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface-7.1.0.dev17-py3.6.egg/pyface/ui/qt4/action/menu_manager.py", line 182, in clear
    super(_Menu, self).clear()
RuntimeError: wrapped C/C++ object of type _Menu has been deleted
```

ievacerny:
I think both of the tracebacks are caused by `MenuManager` and `ToolBarManager` sharing the same list of menus and toolbars (https://github.com/enthought/pyface/pull/541#discussion_r442990452). This should be fixed by #548, so this PR should wait until #548 is merged.

ievacerny:
Closing and reopening to restart CI jobs

PR #549, opened 2020-06-22
Add cron job to install/test against packages from source

rahulporuri:
At the moment, traits is the only package installed from source as pyface is always installed from source on CI. The changes are almost the same as that in traits PR enthought/traits#479.

Exactly the same as the changes in https://github.com/enthought/traitsui/pull/914 except for replacing the word `"pyface"` with `"traitsui"`

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -92,6 +92,11 @@ supported_combinations = {
 
 dependencies = {"traits", "numpy", "pygments", "coverage", "mock"}
 
+# NOTE : traitsui is always installed from source
+source_dependencies = {
+    "traits",
+}
+
 extra_dependencies = {
     "pyside": {"pyside"},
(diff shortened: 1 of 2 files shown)


rahulporuri:
Note that I added a cron job from travis.org settings page for pyface - the cron job runs weekly against master.

etstool.py
```diff
@@ -134,6 +139,8 @@
     envvar="ETSTOOL_EDM",
 )
 
+github_url_fmt = "git+http://github.com/enthought/{0}.git#egg={0}"
```
corranwebster, line 142: I would make `source_dependencies` a dict and specify the complete pip/github url as the value.  That is simpler and more general.

rahulporuri: addressed.

rahulporuri: note that i made the same change in traitsui as well

PR #552, opened 2020-06-26
TST: Test Qt dock widget has the correct size for dock panes

kitchoi:
[Targeting #545]

This PR adds tests to ensure the dock widget size is as expected.

To ensure the tests are indeed meaningful, the dock widget is added to a main window that also has a central widget, and that the main window should have a larger dimensions.
To further that assurance, setting `is_qt4 = True` in a Qt5 environment results in test failures and setting `is_qt4 = False` in a Qt4 environment results in test failures.

### DIFF
--- a/pyface/ui/qt4/tasks/tests/test_main_window_layout.py
+++ b/pyface/ui/qt4/tasks/tests/test_main_window_layout.py
@@ -48,7 +48,7 @@ def create_dummy_dock_widget(parent):
 @unittest.skipIf(
     ETSConfig.toolkit != "qt4",
     "This test targets Qt specific MainWindowLayout. "
-    "Currently toolkit is not Qt."
+    "Current toolkit is not Qt."
 )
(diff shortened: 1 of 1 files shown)


kitchoi:
The error in https://github.com/enthought/pyface/issues/517 happened twice in a row, and it should have nothing to do with dock pane. Very mysterious.

kitchoi:
The build failure is unrelated and is sporadic: https://github.com/enthought/pyface/issues/516

@pedrorivotti @rahulporuri Could you have a look to see if this can be merged to #545 please?

rahulporuri:
reviewing now.

rahulporuri:
thanks for this @kitchoi !

kitchoi:
> I don't think #516 is the same thing; we need a new issue for this sporadic failure (and we should look into the cause).

OK!

kitchoi:
> I don't think #516 is the same thing; we need a new issue for this sporadic failure (and we should look into the cause).
>
> OK!

Opened #556

PR #559, opened 2020-06-30
# not $

rahulporuri:
This was messing with github automatically linking to the PRs. Note that the release page has been manually fixed (https://github.com/enthought/pyface/releases/tag/7.0.0). See below and after below.

before

after


rahulporuri:
> LGTM. This might be the shortest PR title ever... Imagine a line in the changelog that reads * # not $ (#559). This PR probably gets included into the misc line, so I'd not worry about it.

I was laughing the whole time I was making these changes :D

PR #566, opened 2020-07-13
Fix CodeWidget crashing due to KeyError from pygments Style.style_for_token

kitchoi:
This PR fixes the issue seen in https://github.com/enthought/traitsui/issues/982

pygments tokens are hierarchical: styles can define the styling for groups of tokens rather than an individual token type. The `Style.style_for_token` function is a lower level function that does not resolve the closest token group for a given leaf token. One needs to resolve the token group supported by a given style before calling `style_for_token`. 

This is consistent with how pygments handles the same situation: 
https://github.com/pygments/pygments/blob/4cd83f9e5e16e4bafa208e5be5bb3a9dc1ed813a/pygments/formatters/svg.py#L176-L178

### DIFF
--- a/pyface/ui/qt4/code_editor/tests/test_code_widget.py
+++ b/pyface/ui/qt4/code_editor/tests/test_code_widget.py
@@ -32,6 +32,8 @@ class TestCodeWidget(unittest.TestCase):
         self.qapp.processEvents()
 
     def test_different_lexer(self):
+        # Setting a different lexer should not fail.
+        # See enthought/traitsui#982
         cw = CodeWidget(None, lexer="yaml")
         text = "number: 1"
         cw.setPlainText(text)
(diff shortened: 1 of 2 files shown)


pyface/ui/qt4/code_editor/tests/test_code_widget.py
```diff
@@ -31,6 +31,11 @@ def setUpClass(cls):
     def tearDown(self):
         self.qapp.processEvents()
 
+    def test_different_lexer(self):
+        cw = CodeWidget(None, lexer="yaml")
```
rahulporuri, line 35: maybe add a comment about this being a test for the traitsui bug. For now, a test which just sets the correct lexer is better than no test.

kitchoi: For future reference, this test failed before the fix was made, and passed after the fix.

PR #576, opened 2020-07-15
Fix the index page of docs

rahulporuri:
- Remove references to "Traits visualization" - it should be TraitsUI
- Remove references to TraitsGUI, it should also be TraitsUI

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -21,9 +21,8 @@ Toolkit Backends
 TraitsUI and Pyface define APIs that are independent of any GUI toolkit.
 However, in order to actually produce user interfaces with them, you must
 install a supported Python-based GUI toolkit and the appropriate
-toolkit-specific backend project. Conversely, if you wish to use TraitsUI
-without a GUI, a "null" backend is automatically used in the absence of
-a real backend.
+toolkit-specific backend project. Conversely, a "null" backend is
+automatically used in the absence of a real backend.
 
 Currently, the supported GUI toolkits are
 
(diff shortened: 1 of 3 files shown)


docs/source/index.rst
```diff
@@ -43,13 +44,13 @@ that signifies the GUI toolkit.
...
 
-* 'qt4' or 'qt': Pyside2 or `PyQt <http://riverbankcomputing.co.uk/pyqt/>`_,
+* 'qt4' or 'qt': PySide2 or `PyQt <http://riverbankcomputing.co.uk/pyqt/>`_,
   which provide Python bindings for the `Qt <http://www.qt.io>`_ framework
   version 4 and 5.
 * 'wx': `wxPython 4 <http://www.wxpython.org>`_, which provides Python bindings
   for the `wxWidgets <http://wxwidgets.org>`_ toolkit.
 * 'null': A do-nothing toolkit, for situations where neither of the other
-  toolkits is installed, but Traits is needed for non-UI purposes.
+  toolkits is installed, but TraitsUI is needed for non-UI purposes.
```
kitchoi, line 53: I guess the original sentence was targeting dependencies on things that are in the `traits` package.
But it is _probably_ true that traitsui is needed _now_ for non-UI purposes, solely because of the circular dependencies we are still living with. That wasn't the original intent of this sentence though.

Should we remove "but Traits/TraitsUI is needed ..." part of this line?

rahulporuri: i'm fine with that. people might have different reasons why they want to install TraitsUI and Pyface without a toolkit. I think there's one more place where we mention something of this sort which i will remove as well.

Issue #557, opened 2020-06-30
Pyface Bugfix Release 7.0.1


kitchoi:
Target date: To be confirmed, will edit here once confirmed.

This is a bug fix release and therefore the procedure will be slightly different compared to a major/minor release.
See https://github.com/enthought/traitsui/issues/933 for TraitsUI bugfix release 7.0.1.

Pre-release
---
- [x] Verify PRs that need to be merged to the maintanence branch are included in the Release 7.0.1 milestone
- [x] Create a PR with cherry picked changes and target the `maint/7.0` branch, review and merge (#562)
- [x] Test against other ETS packages and other pyface-using projects
- [x] Update changelog and open PR targeting `maint/7.0` branch  (#562)
- [x] Check MANIFEST and requirements are still up to date.
- [x] Test building the documentation

Release
--------
- [x] Create PR targeting `maint/7.0` branch. In this PR: (#580  )
        - Set release to `7.0.1`, and set `IS_RELEASED` is true; commit
        - Bump the version to `7.0.2` and set `IS_RELEASED` to false; commit
- [x] From the commit at which `IS_RELEASED` is true and version is `7.0.1`, create a release branch `release/7.0.1` from the `maint/7.0` branch; push (Edited post-release: in retrospect, this step may not be necessary)
- [x] On the release commit, add an annotated tag, and push the tag
- [x] Make PR targeting `gh-pages` branch: Generate documentation and copy the content to the branch. Verify that the resulting docs/index.html looks good. See https://github.com/enthought/pyface/pull/503 for a similar PR (#583)
- [x] Install from source distribution and run tests again
- [x] Upload to PyPI
- [x] Test the PyPI package
- [x] Announcement

Post-release
-------------
- [x] Package update for `enthought/free` repository (for EDM)
- [x] Backport release note and change log to master, and possibly `maint/7.0` branch (#578 )

kitchoi:
Updated steps to be consistent with past practices in this project.

rahulporuri:
closing as all of the tasks in the release checklist are now complete.

PR #579, opened 2020-07-15
Bump micro version on maint 7.0 branch

rahulporuri:
This PR does has two commits : 

- The first sets IS_RELEASED to True, making the 7.0.1 release of pyface
- The second flips IS_RELEASED back to False and updates the verison to 7.0.2 to continue development on the maintenance branch

ref : #557 for the release issue

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,9 +16,9 @@ from setuptools import setup, find_packages
 
 MAJOR = 7
 MINOR = 0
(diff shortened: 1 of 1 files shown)


kitchoi:
The commit message said "7.0.2". It should be "7.0.1". To avoid confusion in the future, we do this again. Superseded by #580

Issue #597, opened 2020-07-17
Remove the import of pyface.resource for core functionality


aaronayres35:
in resource_manager.py and i_image_resource.py

kitchoi:
For context, this came from this TODO text file, on this commit: https://github.com/enthought/pyface/commit/349e55d3280cc51149d7b456d1dd7dbb6c150797

Issue #596, opened 2020-07-17
make use of drag and drop features optional


aaronayres35:
Need to make use of drag and drop features optional in this project's code (see dock, grid, sheet, tree, and viewer/tree_viewer) or help make traits.util's drag_and_drop module backend agnostic.  Right now, it has a runtime dependency that wx be installed.

kitchoi:
For context, this came from a very old TODO text file:
https://github.com/enthought/pyface/commit/2e7349793f0c390db4bb330e89d8234ab2acb412

Issue #595, opened 2020-07-17
Remove the dock dependence on TraitsBackendWX


aaronayres35:
Remove the direct imports of traitsui.wx.

kitchoi:
For context, this came from this TODO text file on this commit: https://github.com/enthought/pyface/commit/be504511eb728496e72b92bc78b8c74f9ec10699

Issue #594, opened 2020-07-17
Remove the dependency on etsdevtools.developer AND apptools.io by refactoring dock features modules.


aaronayres35:

kitchoi:
For context, this came from a TODO file on this commit: https://github.com/enthought/pyface/commit/02116ea0ac1f46db269f2a4ea2f3a73a34e7292d

The `etsdevtools` is probably not relevant in this repository anymore? (I could not find it).
Dependencies on `apptools.io` do exist, though.

PR #598, opened 2020-07-17
removed top-level TODO.txt file

### LINKED ISSUE #589: Remove top-level TODO.txt
There is a top-level TODO.txt file which is sitting in the github repository, for reasons i am not aware of. It looks like it was possibly used as a basic issue tracker/todo list for the package. Now that we use GitHub to track issues, we can remove the file, after opening issues for each the todo item in the file. 

Ref : https://github.com/enthought/pyface/blob/master/TODO.txt

aaronayres35:
fixes #589 by removing un-needed TODO.txt file.  All items in TODO.txt file have been added as Github issues (see issues #594 #595 #596 #597)


rahulporuri:
can you list all of the issues that you opened from the contents of this todo file in this PR? That way, when someone opens the issues, they have more context on the issues

PR #534, opened 2020-06-06
Pyface Color Class and Trait Type

corranwebster:
This builds on #520 to create a `Color` class and corresponding `PyfaceColor` trait to represent a color in a toolkit independent way.  It also includes a `ColorDialog` class built on top of these classes.

Together with #520 this resolves #476.

This should not be merged until #520 is merged.  This requires features from Traits 6.1.

PR #608, opened 2020-07-18
Toolkit independent color class for Pyface

corranwebster:
This is the color class from #534 pulled out into a separate PR.

This does not include any of the string parsing, trait type, or dialog classes.

### DIFF
--- a/pyface/color.py
+++ b/pyface/color.py
@@ -1,0 +1,260 @@
+# (C) Copyright 2005-2020 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+""" Color classes and corresponding trait types for Pyface.
+
+The base Color class holds red, green, blue and alpha channel values as
+a tuple of normalized values from 0.0 to 1.0.  Various property traits
+pull out the individual channel values and supply values for the HSV
+and HSL colour spaces (with and without alpha).
+
+The ``from_toolkit`` and ``to_toolkit`` methods allow conversion to and
+from native toolkit color objects.
+"""
+
+import colorsys
+
+from traits.api import (
+    HasStrictTraits, Property, Range, Tuple, cached_property
+)
+
+
+def channels_to_ints(channels, maximum=255):
+    """ Convert an iterable of floating point channel values to integers.
+
+    Values are rounded to the nearest integer, rather than truncated.
+
+    Parameters
+    ----------
+    channels : iterable of float
+        An iterable of channel values, each value between 0.0 and 1.0,
+        inclusive.
+    maximum : int
+        The maximum value of the integer range.  Common values are 15,
+        65535 or 255, which is the default.
+
+    Returns
+    -------
+    values : tuple of int
+        A tuple of values as integers between 0 and max, inclusive.
+    """
+    return tuple(int(round(channel * maximum)) for channel in channels)
+
+
+def ints_to_channels(values, maximum=255):
+    """ Convert an iterable of integers to floating point channel values.
+
+    Parameters
+    ----------
+    values : tuple of int
+        An iterable of values as integers between 0 and max, inclusive.
+    maximum : int
+        The maximum value of the integer range.  Common values are 15,
+        65535 or 255, which is the default.
+
+    Returns
+    -------
+    channels : iterable of float
+        A tuple of channel values, each value between 0.0 and 1.0,
+        inclusive.
+    """
+    return tuple(value / maximum for value in values)
+
+
+#: A trait holding a single channel value.
+Channel = Range(0.0, 1.0, value=1.0, channel=True)
+
+#: A trait holding three channel values.
+ChannelTuple = Tuple(Channel, Channel, Channel)
+
+#: A trait holding four channel values.
+AlphaChannelTuple = Tuple(Channel, Channel, Channel, Channel)
+
+
+class Color(HasStrictTraits):
+    """ A mutable specification of a color with alpha.
+
+    This is a class designed to be used by user interface elements which
+    need to color some or all of the interface element.  Each color has a
+    number of different representations as channel tuples, each channel
+    holding a value between 0.0 and 1.0, inclusive.  The standard red,
+    green, blue and alpha channels are also provided as convenience
+    properties.
+
+    Methods are provided to convert to and from toolkit-specific color
+    objects.
+
+    Colors implement equality testing, but are not hashable as they are
+    mutable, and so are not suitable for use as dictionary keys.  If you
+    need a dictionary key, use an appropriate channel tuple from the
+    object.
+    """
+
+    #: A tuple holding the red, green, blue, and alpha channels.
+    rgba = AlphaChannelTuple()
+
+    #: A tuple holding the red, green, and blue channels.
+    rgb = Property(ChannelTuple(), depends_on='rgba')
+
+    #: The red channel.
+    red = Property(Channel, depends_on='rgba')
+
+    #: The green channel.
+    green = Property(Channel, depends_on='rgba')
+
+    #: The blue channel.
+    blue = Property(Channel, depends_on='rgba')
+
+    #: The alpha channel.
+    alpha = Property(Channel, depends_on='rgba')
+
+    #: A tuple holding the hue, saturation, value, and alpha channels.
+    hsva = Property(AlphaChannelTuple, depends_on=['rgba'])
+
+    #: A tuple holding the hue, saturation, and value channels.
+    hsv = Property(ChannelTuple, depends_on=['rgb'])
+
+    #: A tuple holding the hue, lightness, saturation, and alpha channels.
+    hlsa = Property(AlphaChannelTuple, depends_on=['rgba'])
+
+    #: A tuple holding the hue, lightness, and saturation channels.
+    hls = Property(ChannelTuple, depends_on=['rgb'])
+
+    @classmethod
+    def from_toolkit(cls, toolkit_color, **traits):
+        """ Create a new Color object from a toolkit color object.
+
+        Parameters
+        ----------
+        toolkit_color : toolkit object
+            A toolkit color object, such as a Qt QColor or a Wx wx.Colour.
+        **traits
+            Any additional trait values to be passed as keyword arguments.
+        """
+        from pyface.toolkit import toolkit_object
+        toolkit_color_to_rgba = toolkit_object('color:toolkit_color_to_rgba')
+        rgba = toolkit_color_to_rgba(toolkit_color)
+        return cls(rgba=rgba, **traits)
+
+    def to_toolkit(self):
+        """ Create a new toolkit color object from a Color object.
+
+        Returns
+        -------
+        toolkit_color : toolkit object
+            A toolkit color object, such as a Qt QColor or a Wx wx.Colour.
+        """
+        from pyface.toolkit import toolkit_object
+        rgba_to_toolkit_color = toolkit_object('color:rgba_to_toolkit_color')
+        return rgba_to_toolkit_color(self.rgba)
+
+    def hex(self):
+        """ Provide a hex representation of the Color object.
+
+        Note that because the hex value is restricted to 0-255 integer values
+        for each channel, the representation is not exact.
+
+        Returns
+        -------
+        hex : str
+            A hex string in standard ``#RRGGBBAA`` format that represents
+            the color.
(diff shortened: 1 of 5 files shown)


pyface/ui/null/color.py
```diff
@@ -0,0 +1,51 @@
...
+    Returns
+    -------
+    rgba : tuple
+        A tuple of 4 floating point values between 0.0 and 1.0 inclusive.
+    """
+    return ints_to_channels(color)
+
+
+def rgba_to_toolkit_color(rgba):
+    """ Convertan RGBA tuple to a hex tuple.
```
rahulporuri, line 39: nit : space in "convertan"

pyface/color.py
```diff
@@ -0,0 +1,260 @@
...
+    green = Property(Channel, depends_on='rgba')
+
+    #: The blue channel.
+    blue = Property(Channel, depends_on='rgba')
+
+    #: The alpha channel.
+    alpha = Property(Channel, depends_on='rgba')
+
+    #: A tuple holding the hue, saturation, value, and alpha channels.
+    hsva = Property(AlphaChannelTuple, depends_on=['rgba'])
```
rahulporuri, line 119: style nit : i'm not sure why `depends_on=['rgba']` here and `depends_on='rgba'` earlier.

pyface/color.py
```diff
@@ -0,0 +1,260 @@
...
+        ----------
+        toolkit_color : toolkit object
+            A toolkit color object, such as a Qt QColor or a Wx wx.Colour.
+        **traits
+            Any additional trait values to be passed as keyword arguments.
+        """
+        from pyface.toolkit import toolkit_object
+        toolkit_color_to_rgba = toolkit_object('color:toolkit_color_to_rgba')
+        rgba = toolkit_color_to_rgba(toolkit_color)
+        return cls(rgba=rgba, **traits)
```
rahulporuri, line 144: Q: I'm not sure why we're allowing users to pass `**traits` here.

pyface/color.py
```diff
@@ -0,0 +1,260 @@
...
+        Returns
+        -------
+        toolkit_color : toolkit object
+            A toolkit color object, such as a Qt QColor or a Wx wx.Colour.
+        """
+        from pyface.toolkit import toolkit_object
+        rgba_to_toolkit_color = toolkit_object('color:rgba_to_toolkit_color')
+        return rgba_to_toolkit_color(self.rgba)
+
+    def hex(self):
```
rahulporuri, line 158: Q : Why not have this as a property as well?

pyface/tests/test_color.py
```diff
@@ -0,0 +1,246 @@
...
+    def test_init_hsv(self):
+        color = Color(hsv=(0.4, 0.2, 0.6))
+        self.assertEqual(color.rgba, (0.48, 0.6, 0.528, 1.0))
+
+    def test_init_hlsa(self):
+        color = Color(hlsa=(0.4, 0.2, 0.6, 0.8))
+        self.assertEqual(
+            color.rgba,
+            (0.07999999999999996, 0.32000000000000006, 0.17600000000000007, 0.8)  # noqa: E501
+        )
```
rahulporuri, line 97: nit : it might be cleaner if we can use `assertAlmostEqual` and set a tolerance?

pyface/tests/test_color.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def test_init_hls(self):
+        color = Color(hls=(0.4, 0.2, 0.6))
+        self.assertEqual(
+            color.rgba,
+            (0.07999999999999996, 0.32000000000000006, 0.17600000000000007, 1.0)  # noqa: E501
```
rahulporuri, line 103: same comment as above i.e. `assertAlmostEqual` and tolerance.

pyface/tests/test_color.py
```diff
@@ -0,0 +1,246 @@
...
+        hex_value = color.hex()
+        self.assertEqual(hex_value, "#000000FF")
+
+    def test_eq(self):
+        color_1 = Color(rgba=(0.4, 0.2, 0.6, 0.8))
+        color_2 = Color(rgba=(0.4, 0.2, 0.6, 0.8))
+        self.assertTrue(color_1 == color_2)
+        self.assertFalse(color_1 != color_2)
+
+    def test_eq_not_eequal(self):
```
rahulporuri, line 128: nit - spelling `eequal`

pyface/tests/test_color.py
```diff
@@ -0,0 +1,246 @@
...
+        color_1 = Color(rgba=(0.4, 0.2, 0.6, 0.8))
+        color_2 = Color(rgba=(0.4, 0.4, 0.6, 0.8))
+        self.assertTrue(color_1 != color_2)
+        self.assertFalse(color_1 == color_2)
+
+    def test_eq_other(self):
+        color = Color(rgba=(0.4, 0.2, 0.6, 0.8))
+        self.assertFalse(color == 1)
```
rahulporuri, line 136: This is a bit confusing. Should this comparison return True/False? I was under the impression that this would return an Exception.

pyface/color.py
corranwebster: This is basic defensive programming: if we were ever to subclass `Color` in a way that added new traits (or worse, a subclass which multiply inherits), we want the ability to pass values for those additional traits into any factory class methods we might have.

It's the same reason that you should always have `**traits` args in `__init__` methods of `HasTraits` subclasses

pyface/tests/test_color.py
corranwebster: Equality comparisons should almost always return either True or False for almost any pair of Python objects - the the comparison doesn't make sense, then they aren't equal.  There are exceptions (numpy arrays!), but those exceptions make code much more difficult to write when they occur.

pyface/color.py
corranwebster: Mainly because it isn't settable.  It could be, but that would probably be better with the parsing code.

pyface/tests/test_color.py
corranwebster: `AssertAlmostEqual` doesn't support tuples of floats, only floats themselves.

I'll write a method to do the almost-equals comparison.

rahulporuri:
FTR, I went through the older/larger PR - https://github.com/enthought/pyface/pull/534 - and there aren't any comments on that PR which are relevant here that hadn't been addressed.

pyface/color.py
rahulporuri: that's what i was wondering. resolving.

pyface/color.py
rahulporuri: fair enough. resolving.

pyface/tests/test_color.py
rahulporuri: > Objects of different types, except different numeric types, never compare equal.

from the Python docs - https://docs.python.org/3/library/stdtypes.html#comparisons. I didn't recall this. resolving.

rahulporuri:
PR still LGTM

rahulporuri:
@kitchoi can you make time to review this PR over the next couple of days?

kitchoi:
> can you make time to review this PR over the next couple of days?

Yep can do.

pyface/color.py
```diff
@@ -0,0 +1,260 @@
...
+        values = channels_to_ints(self.rgba)
+        return "#{:02X}{:02X}{:02X}{:02X}".format(*values)
+
+    def __eq__(self, other):
+        if isinstance(other, Color):
+            return self.rgba == other.rgba
+        return NotImplemented
+
+    def __str__(self):
+        return "({:0.5}, {:0.5}, {:0.5}, {:0.5})".format(*self.rgba)
```
kitchoi, line 179: What is this `__str__` used for? 

This reminds me of Python 2 behaviour where `str(float)` does not roundstrip, and it has caused issues because it was not at all obvious (e.g. marshmallow Float uses `str`, and it causes issue for people who try to use marshmallow for serializing data in JSON and deserialized back, and expecting the data to stay the same... and then be surprised when they don't).

That Python 2 has since changed in Python 3 such that `str(float)` does round trip.  I am wondering whatever that had bitten people before with `str(float)` not roundtripping might bite us here.

To be safe, could we use `{!r}` here?

pyface/tests/test_color.py
```diff
@@ -0,0 +1,234 @@
...
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+from unittest import TestCase
+
+from traits.testing.unittest_tools import UnittestTools
```
kitchoi, line 13: Nit: It is available from `traits.testing.api`, it would be preferred to import from `api`.

pyface/color.py
```diff
@@ -0,0 +1,260 @@
...
+"""
+
+import colorsys
+
+from traits.api import (
+    HasStrictTraits, Property, Range, Tuple, cached_property
+)
+
+
+def channels_to_ints(channels, maximum=255):
```
kitchoi, line 29: Looks like we never use a different value for `maximum`. When do we need to set a different value?

pyface/color.py
```diff
@@ -0,0 +1,260 @@
...
+    number of different representations as channel tuples, each channel
+    holding a value between 0.0 and 1.0, inclusive.  The standard red,
+    green, blue and alpha channels are also provided as convenience
+    properties.
+
+    Methods are provided to convert to and from toolkit-specific color
+    objects.
+
+    Colors implement equality testing, but are not hashable as they are
+    mutable, and so are not suitable for use as dictionary keys.  If you
```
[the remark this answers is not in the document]
corranwebster: The objects that this is intended to replace (`QColor` and `wx.Colour` values of the `Color` trait) are mutable.

If we were to make this immutable, the design would be a lot different (probably a `namedtuple` rather than a `HasTraits`)

pyface/color.py
corranwebster: The point of `__str__` is always a human-friendly representation of an object, suitable for use in a `print` statement or, in the case of TraitsUI, the default representation in a UI.  Round-tripping is always a secondary consideration.

In particular, the motivation for this is what will appear in a TraitsUI colour editor's text field.

Given that in practice the colours are going to be converted to bytes, they key thing is that the round-trip byte -> float -> str -> float -> byte works (and it does) so that the actual _color_ doesn't change.

Actually, we can even get away with `{0:3}` here.

If there is a specialist case which requires more precision, the user can override the editor's format function to give full precision.

pyface/color.py
corranwebster: We do in the parsing code in the other PR.

pyface/color.py
corranwebster: Ah - I remember why 5.  The CSS styles allow you to specify with 4-digit hex values if you want.  5 places allows round-tripping all of those successfully.

pyface/color.py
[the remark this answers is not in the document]
kitchoi: I guess it depends on how this object is to be used. If we only ever interface it with the GUI toolkit using `from_toolkit` and `to_toolkit`, then we might not need to mutate the traits on this object at all.

Immutables are much easier to work with (and are much easier to maintain), especially in asynchronous context (GUI is such context). It is probably also easier to start with an immutable, and then turn it into a mutable later. And if later we run into a situation where this object really needs to be mutated, we have a reason to do it. On the other hand, it would be harder to make a mutable an immutable later.

I will leave that up to you whether you want to consider turning this an immutable.

pyface/color.py
kitchoi: OK

pyface/color.py
kitchoi: I am wondering if this is actually a view layer that gets put on the model. Perhaps this conversion could be implemented in the view instead of here?

corranwebster: It could, but you want a default, and since this default also makes sense for other situations where you want a human-friendly representation, I think it makes sense to have a `__str__` method.

Otherwise why does Python have a `__str__` at all?

kitchoi: For a library, it allows more room for modifications later if we are more conservative about the functionality exposed. The tools are there but we may not choose to use it, e.g. namedtuple is there but we would be exposing more capability than we need if we use namedtuple for something where we only needed named attribute access, making it hard to back out later. Anyway, it was just a thought.

pyface/color.py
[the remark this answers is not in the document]
corranwebster: No, I expect that the to/from toolkit functions will be used at the point of need internally by toolkit-specific code and not exposed to typical users.  One goal of this exercise is that eventually we should never have someone have to use a `QColour` or `wx.Colour` unless they are writing toolkit-specific code.

corranwebster:
I think everything is addressed.  Will merge in a little while.

PR #609, opened 2020-07-18
Tooolkit-independent Font class

corranwebster:
This is the color class from #520 pulled out into a separate PR.

This does not include any of the string parsing, trait type, or dialog classes.

### DIFF
--- a/pyface/font.py
+++ b/pyface/font.py
@@ -1,0 +1,316 @@
+# (C) Copyright 2005-2020 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+""" Toolkit-independent font utilities.
+
+Pyface fonts are intended to be generic, but able to be mapped fairly well
+to most backend toolkit font descriptions.  In most cases we can describe
+fonts along the common dimensions that are used by CSS, Wx, and Qt.  However
+when it comes to actually working with a font, the toolkit needs to take the
+description and produce something that is as close as possible to the
+specification, but within the constraints of the toolkit, operating system
+and available fonts on the machine where this is being executed.
+
+Font Properties
+---------------
+
+The properties that fonts have are:
+
+Font Family
+    A list of font family names in order of preference, such as "Helvetica"
+    or "Comic Sans".  There are several generic font family names that can
+    be used as fall-backs in case all preferred fonts are unavailable.  In
+    the case of a font that has been selected by the toolkit this list will
+    have one value which is the actual font family name.
+
+Weight
+    How thick or dark the font glyphs are.  These can be given as a number
+    from 1 (lightest) to 999 (darkest), but are typically specified by a
+    multiple of 100 from 100 to 900, with a number of synonyms such as 'light'
+    and 'bold' available for those values.
+
+Stretch
+    The amount of horizontal compression or expansion to apply to the glyphs.
+    These can be given as a percentage between 50% and 200%, or by strings
+    such as as 'condensed' and 'expanded' that correspond to those values.
+
+Style
+    This selects either 'oblique' or 'italic' variants typefaces of the given
+    font family.  If neither is wanted, the value is 'normal'.
+
+Size
+    The overall size of the glyphs. This can be expressed either as the
+    numeric size in points, or as a string such as "small" or "large".
+
+Variants
+    A set of additional font style specifiers, such as "small-caps",
+    "strikethrough", "underline" or "overline", where supported by the
+    underlying toolkit.
+
+Font Specificiation Class
+-------------------------
+
+The Pyface Font class is a HasStrictTraits class which specifies a requested
+font.  It has traits for all of the font properties, plus additional utility
+methods that produce modifed versions of the font.
+
+It also has methods that convert the Font class to and from a toolkit Font
+class.
+
+"""
+import re
+
+from traits.api import (
+    BaseCFloat, CList, CSet, DefaultValue, Enum, HasStrictTraits, Map, Range,
+    Str, TraitError, TraitType
+)
+from traits.trait_type import NoDefaultSpecified
+
+WEIGHTS = {str(i): i for i in range(100, 1001, 100)}
+
+# Note: we don't support 'medium' as an alias for weight 500 because it
+# conflicts with the usage of 'medium' as an alias for a 12pt font in the CSS
+# specification for font attributes.
+WEIGHTS.update({
+    'thin': 100,
+    'extra-light': 200,
+    'ultra-light': 200,
+    'light': 300,
+    'normal': 400,
+    'regular': 400,
+    'book': 400,
+    'semibold': 600,
+    'demibold': 600,
+    'demi': 600,
+    'bold': 700,
+    'extra-bold': 800,
+    'ultra-bold': 800,
+    'black': 900,
+    'heavy': 900,
+    'extra-heavy': 1000,
+})
+
+STRETCHES = {
+    'ultra-condensed': 50,
+    'ultracondensed': 62.5,
+    'extra-condensed': 62.5,
+    'extracondensed': 62.5,
+    'condensed': 75,
+    'semi-condensed': 87.5,
+    'semicondensed': 87.5,
+    'normal': 100,
+    'semi-expanded': 112.5,
+    'semiexpanded': 112.5,
+    'expanded': 125,
+    'extra-expanded': 150,
+    'extraexpanded': 150,
+    'ultra-expanded': 200,
+    'ultraexpanded': 200,
+}
+
+SIZES = {
+    'xx-small': 7.0,
+    'x-small': 9.0,
+    'small': 10.0,
+    'medium': 12.0,
+    'large': 14.0,
+    'x-large': 18.0,
+    'xx-large': 20.0,
+}
+
+STYLES = ('normal', 'italic', 'oblique')
+
+VARIANTS = ['small-caps', 'underline', 'strikethrough', 'overline']
+
+#: A trait for font families.
+FontFamily = CList(Str, ['default'])
+
+#: A trait for font weights.
+FontWeight = Map(WEIGHTS, default_value='normal')
+
+#: A trait for font stretch values.
+#FontStretch = Range(50.0, 200.0, 100.0)
+
+#: A trait for font styles.
+FontStyle = Enum(STYLES)
+
+#: A trait for font variant properties.
+FontVariants = CSet(Enum(VARIANTS))
+
+
+class FontStretch(BaseCFloat):
+    """ Trait type for font stretches.
+
+    The is a CFloat trait which also allows values which are keys of the
+    stretch dictionary  Values must be floats between 50 and 200, inclusive.
+    """
+
+    #: The default value for the trait.
+    default_value = 100.0
+
+    def __init__(self, default_value=NoDefaultSpecified, **metadata):
+        if default_value != NoDefaultSpecified:
+            default_value = self.validate(None, None, default_value)
+        super().__init__(default_value, **metadata)
+
+    def validate(self, object, name, value):
+        if isinstance(value, str) and value.endswith('%'):
+            value = value[:-1]
+        value = STRETCHES.get(value, value)
+        value = super().validate(object, name, value)
+        if not 50 <= value <= 200:
+            self.error(object, name, value)
+        return value
+
+    def info(self):
+        info = (
+            "a float from 50 to 200, "
+            "a value that can convert to a float from 50 to 200, "
+        )
+        info += ', '.join(repr(key) for key in SIZES)
+        info += (
+            " or a string with a float value from 50 to 200 followed by '%'"
+        )
+        return info
+
+
+class FontSize(BaseCFloat):
+    """ Trait type for font sizes.
+
+    The is a CFloat trait which also allows values which are keys of the
+    size dictionary, and also ignores trailing 'pt' ot 'px' annotation in
+    string values.  The value stored is a float.
+    """
+
+    #: The default value for the trait.
+    default_value = 12.0
+
+    def __init__(self, default_value=NoDefaultSpecified, **metadata):
+        if default_value != NoDefaultSpecified:
+            default_value = self.validate(None, None, default_value)
+        super().__init__(default_value, **metadata)
+
+    def validate(self, object, name, value):
+        if isinstance(value, str) and (
+                    value.endswith('pt') or value.endswith('px')):
+            value = value[:-2]
+        value = SIZES.get(value, value)
+        value = super().validate(object, name, value)
+        if value <= 0:
+            self.error(object, name, value)
+        return value
+
+    def info(self):
+        info = (
+            "a positive float, a value that can convert to a positive float, "
+        )
+        info += ', '.join(repr(key) for key in SIZES)
+        info += (
+            " or a string with a positive float value followed by 'pt' or 'px'"
+        )
+        return info
+
+
+class Font(HasStrictTraits):
+    """ A toolkit-independent font specification. """
+
+    #: The preferred font families.
+    family = FontFamily()
+
+    #: The weight of the font.
+    weight = FontWeight()
+
+    #: How much the font is expanded or compressed.
+    stretch = FontStretch()
+
+    #: The style of the font.
+    style = FontStyle()
+
+    #: The size of the font.
+    size = FontSize()
+
+    #: The font variants.
+    variants = FontVariants()
+
+    @classmethod
+    def from_toolkit(cls, toolkit_font):
+        """ Create a Font from a toolkit font object.
+
+        Parameters
+        ----------
+        toolkit_font : any
+            A toolkit font to be converted to a corresponding class instance,
+            within the limitations of the options supported by the class.
+        """
+        from pyface.toolkit import toolkit_object
+        toolkit_font_to_properties = toolkit_object(
+            'font:toolkit_font_to_properties')
+
+        return cls(**toolkit_font_to_properties(toolkit_font))
+
+    def to_toolkit(self):
+        """ Create a toolkit font object from the Font instance.
+
+        Returns
+        -------
+        toolkit_font : any
+            A toolkit font which matches the property of the font as
+            closely as possible given the constraints of the toolkit.
+        """
+        from pyface.toolkit import toolkit_object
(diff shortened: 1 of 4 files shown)


pyface/ui/qt4/font.py
```diff
@@ -0,0 +1,202 @@
...
+This module provides facilities for converting between pyface Font objects
+and Qt QFont objects, trying to keep as much similarity as possible between
+them.
+"""
+
+from pyface.qt import is_qt4
+from pyface.qt.QtGui import QFont
+
+
+qt_family_to_generic_family = {
```
rahulporuri, line 23: Qt docs - https://doc.qt.io/qt-5/qfont.html#StyleHint-enum
PySide2 docs - https://doc.qt.io/qtforpython/PySide2/QtGui/QFont.html#PySide2.QtGui.PySide2.QtGui.QFont.StyleHint

pyface/ui/qt4/font.py
```diff
@@ -0,0 +1,202 @@
...
+    QFont.Normal: 'normal',
+    QFont.Medium: 'medium',
+    QFont.DemiBold: 'demibold',
+    QFont.Bold: 'bold',
+    QFont.ExtraBold: 'extra-bold',
+    QFont.Black: 'black',
+    99: 'extra-heavy',
+}
+
+style_to_qt_style = {
```
rahulporuri, line 73: Qt docs - https://doc.qt.io/qt-5/qfont.html#Style-enum
PySide2 docs - https://doc.qt.io/qtforpython/PySide2/QtGui/QFont.html#PySide2.QtGui.PySide2.QtGui.QFont.Style

pyface/ui/qt4/font.py
```diff
@@ -0,0 +1,202 @@
...
+    if families and hasattr(qt_font, 'setFamilies'):
+        # Qt 5.13 and later
+        qt_font.setFamilies(families)
+    elif families:
+        qt_font.setFamily(families[0])
+        # Note: possibily could use substitutions here,
+        # but not sure if global (which would be bad, so we don't)
+
+    if default_family is not None:
+        qt_font.setStyleHint(generic_family_to_qt_family[default_family])
```
rahulporuri, line 113: We're using `default_family` to set the style hint but we're not using using it for `setFamily`. I'm not sure why.

pyface/ui/qt4/font.py
```diff
@@ -0,0 +1,202 @@
...
+    """
+    family = []
+
+    if hasattr(toolkit_font, 'families'):
+        # Qt 5.13 and later
+        family = list(toolkit_font.families())
+    elif toolkit_font.family():
+        family.append(toolkit_font.family())
+    if toolkit_font.defaultFamily():
+        family.append(toolkit_font.defaultFamily())
```
rahulporuri, line 148: Is it okay to have duplicates in `family`? e.g. toolkit_font.family() returns the same thing as `toolkit_font.defaultFamily()`. This doesn't caught by FontFamily because it's a `CList`, not a `CSet`.

pyface/ui/qt4/font.py
```diff
@@ -0,0 +1,202 @@
...
+        'stretch': stretch,
+        'style': style,
+        'variants': variants,
+    }
+
+
+def map_to_nearest(target, mapping):
+    """ Given mapping with keys from 0 and 99, return closest value.
+
+    This is used to map back from
```
rahulporuri, line 178: Is this an incomplete sentence? Looks like it. "... Qt size to Pyface font size"

pyface/ui/wx/font.py
```diff
@@ -0,0 +1,183 @@
...
+
+from pyface.font import Font
+
+
+# font weight and size features changed in wxPython 4.1/wxWidgets 3.1
+wx_version = tuple(int(x) for x in wx.__version__.split('.')[:2])
+wx_python_4_1 = (wx_version >= (4, 1))
+
+
+wx_family_to_generic_family = {
```
rahulporuri, line 29: WxPython - https://docs.wxpython.org/wx.FontFamily.enumeration.html#wx-fontfamily

pyface/ui/wx/font.py
```diff
@@ -0,0 +1,183 @@
...
+        1000: wx.FONTWEIGHT_MAX,
+    }
+    wx_weight_to_weight = {
+        wx.FONTWEIGHT_LIGHT: 'light',
+        wx.FONTWEIGHT_NORMAL: 'normal',
+        wx.FONTWEIGHT_BOLD: 'bold',
+        wx.FONTWEIGHT_MAX: 'extra-heavy',
+    }
+
+style_to_wx_style = {
```
rahulporuri, line 99: docs - https://docs.wxpython.org/wx.FontStyle.enumeration.html

pyface/ui/wx/font.py
```diff
@@ -0,0 +1,183 @@
...
+    'script': wx.FONTFAMILY_SCRIPT,
+    'sans-serif': wx.FONTFAMILY_SWISS,
+    'swiss': wx.FONTFAMILY_SWISS,
+    'monospace': wx.FONTFAMILY_MODERN,
+    'modern': wx.FONTFAMILY_MODERN,
+    'typewriter': wx.FONTFAMILY_TELETYPE,
+}
+
+if wx_python_4_1:
+    weight_to_wx_weight = {
```
rahulporuri, line 54: https://docs.wxpython.org/wx.FontWeight.enumeration.html

pyface/ui/wx/font.py
```diff
@@ -0,0 +1,183 @@
...
+        wx.FONTWEIGHT_MEDIUM: 'medium',
+        wx.FONTWEIGHT_SEMIBOLD: 'semibold',
+        wx.FONTWEIGHT_BOLD: 'bold',
+        wx.FONTWEIGHT_EXTRABOLD: 'extra-bold',
+        wx.FONTWEIGHT_HEAVY: 'heavy',
+        wx.FONTWEIGHT_EXTRAHEAVY: 'extra-heavy',
+        wx.FONTWEIGHT_MAX: 'extra-heavy',
+    }
+else:
+    weight_to_wx_weight = {
```
rahulporuri, line 80: atm im unable to find an external reference for this.

pyface/ui/wx/font.py
```diff
@@ -0,0 +1,183 @@
...
+
+    Returns
+    -------
+    wx_font : wx.font.Font
+        The best matching wx font.
+    """
+    size = font.size
+    for family in font.family:
+        if family in generic_family_to_wx_family:
+            family = generic_family_to_wx_family[family]
```
rahulporuri, line 126: nit : can we use `default_family` instead of just `family`?

pyface/font.py
```diff
@@ -0,0 +1,313 @@
...
+        toolkit_font : any
+            A toolkit font which matches the property of the font as
+            closely as possible given the constraints of the toolkit.
+        """
+        from pyface.toolkit import toolkit_object
+        font_to_toolkit_font = toolkit_object('font:font_to_toolkit_font')
+
+        return font_to_toolkit_font(self)
+
+    def __str__(self):
```
rahulporuri, line 269: Can we add a doc string here documenting a few examples of the str reprs of Font traits? It's pretty hard to imagine what the str would look like given the conditionals (the test helped to visualize the complexity)

pyface/tests/test_font.py
```diff
@@ -0,0 +1,285 @@
...
+            weight='demibold',
+            stretch='condensed',
+            style='italic',
+            variants={'small-caps', 'underline'},
+        )
+
+        # smoke test
+        toolkit_font = font.to_toolkit()
+
+    def test_from_toolkit(self):
```
rahulporuri, line 268: can we also add roundtrip tests here? the roundtrip tests for Color reminded me of their lack here.

pyface/font.py
```diff
@@ -0,0 +1,313 @@
...
+    Str, TraitError, TraitType
+)
+from traits.trait_type import NoDefaultSpecified
+
+WEIGHTS = {str(i): i for i in range(100, 1001, 100)}
+
+# Note: we don't support 'medium' as an alias for weight 500 because it
+# conflicts with the usage of 'medium' as an alias for a 12pt font in the CSS
+# specification for font attributes.
+WEIGHTS.update({
```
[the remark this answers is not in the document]
corranwebster: These come from the CSS spec, Qt and Wx constant names, true/open type font conventions, and in some cases (originally) from "traditional" values from the typesetting world (eg. the font weights date back to the fonts used in phototypesetting systems).

The CSS spec is probably the simplest comprehensive reference.

pyface/ui/qt4/font.py
```diff
@@ -0,0 +1,202 @@
...
+    'cursive': QFont.Cursive,
+    'script': QFont.Cursive,
+    'sans-serif': QFont.SansSerif,
+    'swiss': QFont.SansSerif,
+    'monospace': QFont.Monospace,
+    'modern': QFont.Monospace,
+    'typewriter': QFont.TypeWriter,
+}
+
+weight_to_qt_weight = {
```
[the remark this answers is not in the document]
corranwebster: I'm not sure that this needs any extra documentation.

I think taking the CSS parsing out of the PR has made these constants less clear than they would be in the context of all of that.

pyface/ui/qt4/font.py
corranwebster: Same here.

pyface/ui/qt4/font.py
corranwebster: The way that Qt font selection works (prior to 5.13, and ignoring font substitutions, which I think are global state so we don't want to mess with them), is that you specify the font that you want (which we are taking as the _first_ font specified) via `setFamily` and then the fall-back _generic_ font type via `setStyleHint` (so we look for one of the special matching generic names, which are a union of Qt, Wx, and CSS special names).

pyface/ui/qt4/font.py
corranwebster: Order matters here, so we don't want a `CSet`, and for Qt <= 5.12 we only take the first, so duplicates don't matter.

No idea what Qt does with duplicates in 5.13+ - I would hope that they are harmless.

pyface/ui/wx/font.py
corranwebster: There isn't one - this is essentially arbitrary, but hopefully sensible.

pyface/font.py
corranwebster: I have added a brief docstring to indicate that it it is CSS style.

Again, this would be clearer with the CSS parsing code available.

pyface/tests/test_font.py
corranwebster: No, unfortunately there isn't 100% fidelity in the round-trips: Qt can lose 2nd choice font families, and Wx loses some font weights, so you aren't guaranteed to get the same thing back.

corranwebster: Although, now that I think about it, it should be round-trip stable toolkit -> pyface -> toolkit.

corranwebster: It turns out if you do pyface -> toolkit -> pyface then you get something that can should be able to do another -> toolkit -> pyface without changing.  I've added tests for this.

corranwebster: Or, alternatively, what is true on my system isn't true on random CI system with who know s what fonts or lack thereof.  I'm going to remove those tests, as they will be essentially impossible to guarantee that they are right on every system.

rahulporuri:
FTR, the roundtrip failures on all three platforms only happens with pyside2 (we didn't get to test wx on linux/macos because the build stopped when it saw the pyside2 failures). Given this, i'd prefer skipping the roundtrip tests on pyside2 instead of not having them at all :  

```
FAIL: test_toolkit_round_trip (pyface.tests.test_font.TestFont)
  File "/home/travis/.edm/envs/pyface-test-3.6-pyside2/lib/python3.6/site-packages/pyface-7.1.0.dev48-py3.6.egg/pyface/tests/test_font.py", line 295, in test_toolkit_round_trip
    self.assertEqual(result, stable_font)
AssertionError: Font([19 chars]s', 'DejaVu Sans', 'default'], size=12.0, stre[65 chars]mal') != Font([19 chars]s', 'default'], size=12.0, stretch=100.0, styl[50 chars]mal')

FAIL: test_toolkit_round_trip_typical (pyface.tests.test_font.TestFont)
  File "/home/travis/.edm/envs/pyface-test-3.6-pyside2/lib/python3.6/site-packages/pyface-7.1.0.dev48-py3.6.egg/pyface/tests/test_font.py", line 312, in test_toolkit_round_trip_typical
    self.assertEqual(result, stable_font)
AssertionError: Font([32 chars]s', 'DejaVu Sans', 'sans-serif'], size=14.0, s[96 chars]old') != Font([32 chars]s', 'sans-serif'], size=14.0, stretch=75.0, st[81 chars]old')

```

pyface/ui/qt4/font.py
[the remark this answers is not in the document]
rahulporuri: > I think taking the CSS parsing out of the PR has made these constants less clear than they would be in the context of all of that.

i'm still happier with the PR as it is. The docs weren't too hard to find so resolving this comment.

pyface/tests/test_font.py
rahulporuri: > ... , as they will be essentially impossible to guarantee that they are right on every system.

hmm. fair enough.

corranwebster:
I'd rather not - the round-tripping test was going for "cheap-and-cheerful", but if it's not generic then it's not particularly useful.

Better in that case is to have toolkit-specific tests that allow more fine-grained tests for what works in each permutation of back-ends.

rahulporuri:
There seem to be a few review comments from #520 for the changes made in this PR which don't seem to be addressed - 

- [ ] https://github.com/enthought/pyface/pull/520/files#r442919620
- [ ] https://github.com/enthought/pyface/pull/520/files#r444788258
- [ ] https://github.com/enthought/pyface/pull/520/files#r442956713
- [ ] https://github.com/enthought/pyface/pull/520/files#r445009225
- [ ] https://github.com/enthought/pyface/pull/520/files#r444984394
- [ ] https://github.com/enthought/pyface/pull/520/files#r445011330

pyface/font.py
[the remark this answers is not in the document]
kitchoi: Looks like we have some extra weights here compared to the CSS spec: https://www.w3.org/TR/css-fonts-3/#propdef-font-weight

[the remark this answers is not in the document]
kitchoi: More specifically, these are the values from CSS: `normal | bold | bolder | lighter | 100 | 200 | 300 | 400 | 500 | 600 | 700 | 800 | 900`
Where do the other values come from?

pyface/font.py
```diff
@@ -0,0 +1,314 @@
...
+    'semicondensed': 87.5,
+    'normal': 100,
+    'semi-expanded': 112.5,
+    'semiexpanded': 112.5,
+    'expanded': 125,
+    'extra-expanded': 150,
+    'extraexpanded': 150,
+    'ultra-expanded': 200,
+    'ultraexpanded': 200,
+}
```
kitchoi, line 116: Same here, CSS has these values: `normal | ultra-condensed | extra-condensed | condensed | semi-condensed | semi-expanded | expanded | extra-expanded | ultra-expanded`

I don't think we need the extra ones here if we document the fact that these values follow CSS convention.

pyface/ui/qt4/font.py
```diff
@@ -0,0 +1,199 @@
...
+    800: QFont.ExtraBold,
+    900: QFont.Black,
+    1000: 99,
+}
+qt_weight_to_weight = {
+    0: 'thin',
+    QFont.ExtraLight: 'extra-light',
+    QFont.Light: 'light',
+    QFont.Normal: 'normal',
+    QFont.Medium: 'medium',
```
kitchoi, line 64: Isn't this `"medium"` not allowed in the `weight` trait? 
In fact, could we map all of these to `"100"`, `"200"` and so on? Then `QFont.Medium` will be mapped to `"500"`, which is allowed.

pyface/ui/qt4/font.py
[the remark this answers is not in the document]
kitchoi: Me too, I am happy that this PR is smaller, which is easier to review. It is also fairly easy to reference a separate branch when the context can be found.

pyface/ui/qt4/font.py
corranwebster: Getting back to this after a year, since this is just the descriptions with no attempt at font-name parsing, `'medium'` should be allowed both here and in the weight trait as there is no conflict with other uses of `medium` at this level.

pyface/font.py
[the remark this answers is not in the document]
corranwebster: As mentioned above, this wasn't just from the CSS spec - in particular it should at least include the mappings that Enable has here https://github.com/enthought/enable/blob/db84efa4d031fd641ebe097301a56764634ce21d/kiva/fonttools/_constants.py#L90-L105
as well as the names that Qt uses for weight constants: https://doc.qt.io/qt-5/qfont.html#Weight-enum
and the names wx uses: https://wxpython.org/Phoenix/docs/html/wx.FontWeight.enumeration.html
and the OpenType font weights: https://docs.microsoft.com/en-us/typography/opentype/spec/os2#usweightclass

More info can be found here: https://bigelowandholmes.typepad.com/bigelow-holmes/2015/07/on-font-weight.html

This is all very fuzzy because the names predate computerized fonts and most font families don't supply all weights.

corranwebster:
> There seem to be a few review comments from #520 for...

Looking through these, I have checked off the ones that have been addressed.  Of the others:

- keeping weights to round numbers for now (we can always allow any number later) because:
  - CSS 2.x (which is roughly what we are modelling right now) only supports round values
  - Qt's numeric weights are non-linear from 0 to 99 rather than 0-1000, which is awkward
  - most of the time it will get rounded to something, and most fonts we will encounter in scientific computing will have round weight values
  - I'd need to write a new trait type to support it, rather than relying on `Map`
  - it feels very YAGNI to me right now
- there is an issue with specifying multiple font families on Qt < 5.13 (we simply take the first specified, plus the first default family as a backup).  Making this work like other systems is complex (eg. Enable, where we have a whole subsystem for this) or require the app to be running.  Again, if we need it we can do it, but this feels like overkill for an old version of Qt when new versions of Qt just do what we want.  It is possible we could pull the Enable font system into Pyface, but that also doesn't feel right.
- I want a simple human-readable string representation (having it CSS-ish is nice, because it is likely to be understood) because the end-goal of this is for things like font editors in TraitsUI or Enable, and we want to be able to display the selected font unambiguously.  It could be a separate method or function, but `__str__` feels OK and I am not terribly worried about subclassing (again, YAGNI here, I think)
- we can't do round-tripping tests for the toolkit objects because of things like the issue with Qt < 5.13, but we have smoke tests and I'm adding a couple more tests for things that should work independent of toolkit.

PR #612, opened 2020-07-20
Use mock from unittest

### LINKED ISSUE #353: Use mock from unittest in Python 3

aaronayres35:
fixes #353

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -90,7 +90,7 @@ supported_combinations = {
     "3.6": {"pyqt", "pyqt5", "pyside2", "wx"},
 }
 
-dependencies = {"traits", "numpy", "pygments", "coverage", "mock"}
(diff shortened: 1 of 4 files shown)


rahulporuri:
This PR needs one additional change to be fully complete. Can you remove `mock` from the set of `dependencies` in etstool here - https://github.com/enthought/pyface/blob/4fda22878e64d930ce0d2bebac69907ea71b31b9/etstool.py#L93 

That way, we can confirm that `mock` package isn't installed in the environment and the testsuite works as expected.

rahulporuri:
Changes still LGTM. Merge when CI is happy.

PR #622, opened 2020-07-22
Add --editable flag to etstool install command

kitchoi:
For developers making frequent changes to the code (e.g. switching between branches), it is more convenient to install the source in editable mode.
This PR adds the `--editable` flag to the `etstool`. This flag can also be found in `traits` and `traitsui` etstool.


kitchoi:
> Should the changes you made to the doctstring be mirrored in the traitsui etstool file?

Yep.

PR #628, opened 2020-07-22
Reduce verbosity when installing package in CI

### LINKED ISSUE #626: Reduce verbosity when installing package in CI
> The job exceeded the maximum log length, and has been terminated.

https://travis-ci.org/github/enthought/pyface/jobs/710848975. The setup step created logs from line #432 to line #9768. The job got terminated when the line # got to ~12000. There is no good reason why we don't reduce verbosity when installing the package - atleast provide the option to reduce verbosity and enable it by default on CI.

This essentially means passing `--quiet` to `python setup.py install` in the etstool.py `install` command
https://github.com/enthought/pyface/blob/8dc1d16c797d56752719060fcab9c601ce12065a/etstool.py#L166

aaronayres35:
fixes #626 
I tried to make passing quiet optional, but the default behavior.  I wasn't sure if you meant to make it the default behavior for the CI (making it optional but then when called in the travis/appveyor files they use quiet) or default within etstool which is what I tried.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -151,6 +151,7 @@ def cli():
 @click.option("--toolkit", default="pyqt", help="Toolkit and API to use")
 @click.option("--environment", default=None, help="EDM environment to use")
 @click.option('--source/--no-source', default=False)
+@click.option('--quiet/--no-quiet', default=True)
 def install(edm, runtime, toolkit, environment, source):
     """ Install project and dependencies into a clean EDM environment.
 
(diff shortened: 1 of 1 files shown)


etstool.py
```diff
@@ -151,6 +151,7 @@ def cli():
 @click.option("--toolkit", default="pyqt", help="Toolkit and API to use")
 @click.option("--environment", default=None, help="EDM environment to use")
 @click.option('--source/--no-source', default=False)
+@click.option('--quiet/--no-quiet', default=True)
 def install(edm, runtime, toolkit, environment, source):
```
rahulporuri, line 155: We're missing a `quiet` argument in the `install` function/command. Ref : https://click.palletsprojects.com/en/7.x/ click is the library that we use to enable parsing the command line arguments.

etstool.py
```diff
@@ -163,8 +164,15 @@ def install(edm, runtime, toolkit, environment, source):
...
-        "{edm} run -e {environment} -- python setup.py install",
     ]
+    if quiet:
+        commands.append(
+            "{edm} run -e {environment} -- python setup.py install --quiet"
+        )
+    else:
+        commands.append(
+            "{edm} run -e {environment} -- python setup.py install"
+        )
```
kitchoi, line 175: Oops, just saw this. I think this now needs to resolve merge conflicts with #622

rahulporuri: #622 didn't remove the original `python setup.py install` (it was doing install twice) which is why i don't think there were merge conflicts.

@aaronayres35 it'll probably be better to integrate with the approach that kit introduced in #622. Like the `if editable:` conditional, we'd just add `--quiet` to `install_pyface` below if `quiet` is set.

Does that make sense?

PR #635, opened 2020-07-24
remove conditionals on python version

### LINKED ISSUE #601: Python 2 cleanup
There are still a few places where we seem to have conditionals that check the python version and behave differently on Python 2 vs 3. We can now throw out these conditionals that behave different on Python 2 because we don't support python 2 anymore.

The conditionals mostly use `sys.version` and `sys.version_info`, which is why i guess they weren't caught earlier

- [ ] `pyface.gui_application`
- [ ] `pyface.qt.__init__`
- [ ] `pyface.timer.i_timer`
- [ ] `pyface.ui.qt4.mimedata`

Note that once the use in `pyface.qt.__init__` is removed, we might also be able to remove this util function `prepare_pyqt4` https://github.com/enthought/pyface/blob/0f3202b4b72d5bc17df741a9c149d1867440fdb0/pyface/qt/__init__.py#L24-L27

Additionally, 

- [x] there are a lot of `u'...'` prefixes which we can now get rid of.

aaronayres35:
fixes #601

### DIFF
--- a/pyface/qt/__init__.py
+++ b/pyface/qt/__init__.py
@@ -58,26 +58,21 @@ qt_api = None
 for api_name, module in QtAPIs:
     if module in sys.modules:
         qt_api = api_name
-        if qt_api == "pyqt" and sys.version_info[0] <= 2:
-             # set the PyQt4 APIs
-             # this is a likely place for failure - pyface really wants to be
-             # imported first, before eg. matplotlib
-             prepare_pyqt4()
         break
 else:
     # does our environment give us a preferred API?
     qt_api = os.environ.get("QT_API")
     if qt_api == "pyqt":
-         # set the PyQt4 APIs
-         prepare_pyqt4()
+        # set the PyQt4 APIs
+        prepare_pyqt4()
 
 # if we have no preference, is a Qt API available? Or fail with ImportError.
 if qt_api is None:
     for api_name, module in QtAPIs:
         try:
             if api_name == "pyqt":
-                 # set the PyQt4 APIs
-                 prepare_pyqt4()
+                # set the PyQt4 APIs
+                prepare_pyqt4()
             importlib.import_module(module)
             importlib.import_module(".QtCore", module)
             qt_api = api_name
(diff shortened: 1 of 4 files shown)


aaronayres35:
> Changes LGTM but i'm a little concerned about the changes in `pyface.qt.__init__`. Can you hold off on merging this until we can get another pair of eyes on this PR?

Yeah I will hold off, tbh I was a bit unsure myself - was curious for any feedback.

pyface/qt/__init__.py
```diff
@@ -21,63 +21,21 @@
...
-    except ImportError:
-        import sip
-    try:
-        sip.setapi("QDate", 2)
-        sip.setapi("QDateTime", 2)
-        sip.setapi("QString", 2)
-        sip.setapi("QTextStream", 2)
-        sip.setapi("QTime", 2)
-        sip.setapi("QUrl", 2)
-        sip.setapi("QVariant", 2)
```
kitchoi, line 40: This code path still gets run in Python 3 with PyQt. The outcome of this import side effect may not have been well tested, and it would be difficult to track down to this change if this removal should cause issues.
Are there other reasons for removing this apart from the comment and message surrounding it being Python 2 specific?

aaronayres35: No, it was simply removed because of the comment and Rahul saying we might be able to remove it on the original issue. To prevent potential additional headaches from its removal, I will add it back in - thanks!

kitchoi: I see. Yes please add it back in. Thank you!
Finger-crossed we won't need to support PyQt4 for much longer so this function is on already the death row - just not quite there yet :)

aaronayres35: Should I add back in the places where it was called?  The one I'm most unsure about is the one below that was on line 61 as that if statement only occurs if python2 is being used explicitly.

kitchoi: You're right - I think the one under `if qt_api == "pyqt" and sys.version_info[0] <= 2:` can be safely removed.

PR #636, opened 2020-07-24
Review comments for Data View selection support

kitchoi:
[Targeting #574]

This PR addresses and/or accompanies my own review comments on #574
As I was reviewing the PR, I find it necessary to test out my comments and suggestions. As I do so, I find more things in the PR that I otherwise would not have noticed just by reading it. I also got to weed out some of my faulty comments so no ones get to be bothered by them :)

This does not need to be merged, but I will let @corranwebster decide what to do. I did not try to break this down into smaller PRs because it is otherwise too hard.

Main changes:
- Validating row and column indexes even when the list of selections is mutated
- Resetting selection when the selection mode changes regardless of toolkit backend
- Handle when changing selection leads to changes to other things that will then change the selection
- Validating index for when selection type is "item"

New tests are added for all of these changes.

There is a new test case that depends on a fake data model.  This is needed to test the many code branches for validating row and column. Otherwise some of the edge cases cannot be covered simply by using the ArrayDataModel (which is a regular data structure).

In order to test row and column index in a toolkit agnostic fashion (the code itself is toolkit agnostic), I also had to expand the null toolkit's `Widget`. That in itself could be a separate PR.

### DIFF
--- a/pyface/data_view/i_data_view_widget.py
+++ b/pyface/data_view/i_data_view_widget.py
@@ -179,9 +179,7 @@ class MDataViewWidget(HasStrictTraits):
         """ Observer for selection_type trait. """
         if self.control is not None:
             self._set_control_selection_type(event.new)
-            with self._selection_updating():
-                self.selection.clear()
-                self._set_control_selection(self.selection)
+            self.selection.clear()
 
     def _get_control_selection_type(self):
         """ Toolkit specific method to get the selection type. """
@@ -195,9 +193,7 @@ class MDataViewWidget(HasStrictTraits):
         """ Observer for selection_mode trait. """
         if self.control is not None:
             self._set_control_selection_mode(event.new)
-            with self._selection_updating():
-                self.selection.clear()
-                self._set_control_selection(self.selection)
+            self.selection.clear()
 
     def _get_control_selection_mode(self):
         """ Toolkit specific method to get the selection mode. """
--- a/pyface/data_view/i_data_view_widget.py
+++ b/pyface/data_view/i_data_view_widget.py
@@ -250,6 +250,7 @@ class MDataViewWidget(HasStrictTraits):
         self._set_control_header_visible(self.header_visible)
         self._set_control_selection_mode(self.selection_mode)
         self._set_control_selection_type(self.selection_type)
+        self._set_control_selection(self.selection)
 
     def _add_event_listeners(self):
         logger.debug('Adding DataViewWidget listeners')
--- a/pyface/data_view/i_data_view_widget.py
+++ b/pyface/data_view/i_data_view_widget.py
@@ -224,7 +224,10 @@ class MDataViewWidget(HasStrictTraits):
     def _update_selection(self, *args, **kwargs):
         if not self._selection_updating_flag:
             with self._selection_updating():
-                self.selection = self._get_control_selection()
+                new = self._get_control_selection()
+                old = self.selection
+                if new != old:
+                    self.selection = new
 
     # ------------------------------------------------------------------------
     # Widget Interface
--- a/pyface/data_view/i_data_view_widget.py
+++ b/pyface/data_view/i_data_view_widget.py
@@ -12,8 +12,9 @@ from contextlib import contextmanager
 import logging
 
 from traits.api import (
-    Bool, ComparisonMode, Enum, HasStrictTraits, Instance, List, Property,
-    TraitError, Tuple, cached_property,
+    BaseTuple,
+    Bool, ComparisonMode, Enum, HasStrictTraits, Instance, List,
+    TraitError, Tuple,
 )
 
 from pyface.data_view.abstract_data_model import AbstractDataModel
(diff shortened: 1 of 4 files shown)


pyface/data_view/tests/test_data_view_widget.py
```diff
@@ -185,6 +502,52 @@ def test_selection_mode_extended(self):
...
+        # switch from "extended" to "single"
+        self.widget.selection = [((0,), ()), ((1,), ())]
+        self.gui.process_events()
+        self.assertEqual(self.widget._get_control_selection_mode(), "single")
+
+        if is_wx:
+            # Not sure why the selection is not updated.
+            self.assertEqual(
+                self.widget._get_control_selection(), [((0,), ()), ((1,), ())]
+            )
```
kitchoi, line 533: The expected value is an empty list. I tried to understand why this did not work for wx, but I couldn't.

pyface/data_view/i_data_view_widget.py
```diff
@@ -116,7 +222,7 @@ def _set_control_selection_mode(self, selection_mode):
 
     def _selection_updated(self, event):
         """ Observer for selection trait. """
-        if self.control is not None and not self._selection_updating_flag:
```
kitchoi, line 119: I removed this to test if it is necessary at all. I tried to test but I could not see why this is needed.

pyface/data_view/i_data_view_widget.py
```diff
@@ -24,6 +27,108 @@
...
+            raise TraitError("Invalid row index {!r}".format(row))
+
+        if object.selection_type == 'item':
+            can_have_children = object.data_model.can_have_children(row)
+            have_rows = object.data_model.get_row_count(row) > 0
+            if can_have_children and have_rows:
+                raise TraitError(
+                    "Row values must have no children when selection_type "
+                    "is 'item', got {!r}".format(row)
+                )
```
kitchoi, line 52: I added the validation for 'item' following the rest of the validation logic. But I have not added the code in the toolkit specific widget to actually avoid setting an invalid selection. e.g. if the user tries to click on row and column, the toolkit specific widget at the moment does not filter the event. Then we will get a trait error, and in Qt case, a hard crash. That filtering will need to be done if we include this validation.

pyface/data_view/tests/test_data_view_widget.py
kitchoi: Weirdly, this assertion fails on Travis CI OSX build, but passes on Travis CI Linux build.
The assertion passes (i.e. expected failure) on my local machine too... It'd be good to get an independent confirmation. :puzzled:

kitchoi: On Travis CI linux build, it printed this error message and the test continued to "pass"(?):
```
test_selection_mode_change_when_selection_change (pyface.data_view.tests.test_data_view_widget.TestWidget) ... Traceback (most recent call last):
  File "/Users/travis/.edm/envs/pyface-test-3.6-wx/lib/python3.6/site-packages/pyface-7.1.0.dev110-py3.6.egg/pyface/data_view/i_data_view_widget.py", line 247, in _update_selection
    self.selection = new
  File "/Users/travis/.edm/envs/pyface-test-3.6-wx/lib/python3.6/site-packages/pyface-7.1.0.dev110-py3.6.egg/pyface/data_view/i_data_view_widget.py", line 127, in validate
    "is 'single', got {!r}".format(value)
traits.trait_errors.TraitError: Selection must have at most one element when selection_mode is 'single', got [((0,), ()), ((1,), ())]
  File "/Users/travis/.edm/envs/pyface-test-3.6-wx/lib/python3.6/site-packages/pyface-7.1.0.dev110-py3.6.egg/pyface/data_view/i_data_view_widget.py", line 247, in _update_selection
    self.selection = new
  File "/Users/travis/.edm/envs/pyface-test-3.6-wx/lib/python3.6/site-packages/pyface-7.1.0.dev110-py3.6.egg/pyface/data_view/i_data_view_widget.py", line 127, in validate
    "is 'single', got {!r}".format(value)
traits.trait_errors.TraitError: Selection must have at most one element when selection_mode is 'single', got [((0,), ()), ((1,), ())]
ok
```

pyface/data_view/i_data_view_widget.py
kitchoi: Or I might have misunderstood what "item" means. The above code tries to exclude row and column headers, but maybe "item" is meant to include them? Headers do tend to represent different data though. 🤫

pyface/data_view/tests/test_data_view_widget.py
```diff
@@ -185,6 +502,52 @@ def test_selection_mode_extended(self):
...
+        self.addCleanup(
+            self.widget.observe,
+            change_selection_mode, "selection.items", dispatch="ui",
+            remove=True,
+        )
+
+        # switch from "extended" to "single"
+        self.widget.selection = [((0,), ()), ((1,), ())]
+        self.gui.process_events()
+        self.assertEqual(self.widget._get_control_selection_mode(), "single")
```
corranwebster, line 527: Just to clarify, this is not how selection mode works! Selection mode tells you what sorts of things can be selected, and does not change based on what the actual selection _is_.

When selection mode is `single` then the selection is always a list with one item.

When selection mode is `extened` then the selection is a list with any number of items, _including_ possibly just one item.

kitchoi: I am confused. From reading the code, the widget resets the selection to an empty list whenever selection mode is changed.
The line you are commenting on is asserting the selection *mode*. When the selection mode is changed on the `self.widget`, the selection mode on the GUI control should be synchronized to the new value, right?

kitchoi: To clarify, this test is testing the logic in this method:
https://github.com/enthought/pyface/blob/9daeb87e832e85d4ce0fbd29a5dc1a700bdec3bb/pyface/ui/qt4/data_view/data_view_widget.py#L91-L95

corranwebster:
Closing, as the original PR is merged.

corranwebster:
Ooops - underlying PR hasn't been merged, but still closing as things have been fixed independently in that PR.

kitchoi:
Yep. Thank you!

PR #641, opened 2020-07-28
fixed a syntax error encountered when running enaml example

aaronayres35:
This is what the final window looks like after the fix.
The example is found in https://github.com/enthought/pyface/tree/master/examples/tasks/enaml

Screen Shot 2020-07-28 at 9 01 22 AM

### DIFF
--- a/examples/tasks/enaml/employee_view.enaml
+++ b/examples/tasks/enaml/employee_view.enaml
@@ -30,7 +30,7 @@ enamldef EmployeeForm(Form):
         text = 'Password'
     Field:
         echo_mode << 'password' if not pw_cb.checked else 'normal'
-        text :: print('Password:'), text
+        text :: print("Password:{}".format(text))
     Label:
         text = 'Show Password:'
     CheckBox: pw_cb:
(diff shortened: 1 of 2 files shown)


examples/tasks/enaml/employee_view.enaml
```diff
@@ -30,7 +30,7 @@ enamldef EmployeeForm(Form):
         text = 'Password'
     Field:
         echo_mode << 'password' if not pw_cb.checked else 'normal'
-        text :: print 'Password:', text
+        text :: print('Password:'), text
```
rahulporuri, line 33: On python 2, if i remember correctly, `print 'Password:', text` will print the value of the variable `text` after the string `"Password:"`. So, the right fix here is something like `print("Password: {}".format(text))` or something of that sort.

rahulporuri:
also, as long as we're on the topic, can we also modify the `empty_form.enaml` to remove the duplicate `Confirm email:` entries - https://github.com/enthought/pyface/blob/f27c847a091d84be2dc90a53abeb067411241330/examples/tasks/enaml/empty_form.enaml#L48-L71 

I'm not sure why there are 6 fields for `"Confirm email:"`. One should suffice.

PR #650, opened 2020-07-29
changing the SKIP message to specify traits_enaml

aaronayres35:


pyface/tasks/tests/test_enaml_task_pane.py
```diff
@@ -20,7 +20,7 @@
         from enaml.widgets.api import Label
         from traits_enaml.testing.gui_test_assistant import GuiTestAssistant
     except ImportError:
-        SKIP_REASON = "Enaml not installed"
+        SKIP_REASON = "traits_enaml not installed"
```
rahulporuri, line 23: nit-picky comment : can we add an "is" to be grammatically correct? i.e. "traits_enaml is not installed"

PR #654, opened 2020-08-03
replacing collections with collections.abc

aaronayres35:
As mentioned in https://github.com/enthought/traitsui/pull/1103, replace collections imports with collections.abc because of deprecation.


kitchoi:
Do you know why `pyface/image/library/icons.zip` and `pyface/image/library/std.zip` are modified here?

aaronayres35:
> Do you know why `pyface/image/library/icons.zip` and `pyface/image/library/std.zip` are modified here?

Oh, that is unintentional, those shouldn't have been added to the commit.  I don't know what causes them to change I never open the files, but they sometimes show up as modified on my machine when I am making other changes.

kitchoi:
No problem. I think I have seen this behaviour before but I have not spent the time to investigate why. For this PR, would you revert the changes to the binary files please since they are not required for the fix?

PR #657, opened 2020-08-04
Maint: Remove a print call in test

kitchoi:
This PR removes an invocation of `print` in a test. Probably left behind from debugging.


rahulporuri:
_i wonder how many more debugging-related print statements are sitting in the codebase_

PR #661, opened 2020-08-10
DataView row model

corranwebster:
This PR creates a row-based table (ie. not hierarchical) model suitable for displaying things like lists of objects where columns are homogenous.  Data access is abstracted out via `DataAccessors` to allow handling of lists of dictionaries, lists of lists, or potentially (untested!) things like Pandas DataFrames in a declarative way.

No assumption is made about the data being viewed, so beyond complete replacement of the underlying data object, it is up to using code to notify the model of changes to the data structure.


pyface/data_view/data_models/row_table_data_model.py
```diff
@@ -0,0 +1,126 @@
...
+        else:
+            column_data = self.column_data[column[0]]
+        if len(row) == 0:
+            return column_data.title_type
+        return column_data.value_type
+
+    def _data_default(self):
+        return []
+
+    @observe('data')
```
kitchoi, line 99: Without assuming a trait container type such as `List`, `data` can be any object conforming to the Sequence interface, such as a numpy array.  However the model won't have builtin support for value updates triggered by mutation on `data` if `data` is actually a TraitList.

Actually, I think it would be okay to replace this `@observe("data")` with `@observe(trait("data").list_items(optional=True))`. The `optional=True` will excuse the situation when `data` isn't a TraitList. But if `data` is a TraitList and is mutated, this model will fire `structure_changed` event. I think this is EEP2 and EEP3 proving their values here :)

Note: It would be naughty to use `"data.items"` because `"items"` is equivalent to `list_items(optional=True) | dict_items(optional=True) | set_items(optional=True) | trait("items", optional=True)`. We can't be sure if `data` is an instance of `HasTraits` that happens to have a trait named `"items"`; we don't want to be accidentally observing that trait for this update. And then dict and set aren't sequence so we are not expecting these being used as `data`.

Alternatively, we may need to instruct users to dynamically attach an observer if they need to observe mutation on `data` that is a trait list.

pyface/data_view/data_models/row_table_data_model.py
```diff
@@ -0,0 +1,126 @@
...
+from pyface.data_view.index_manager import IntIndexManager
+from pyface.data_view.data_models.data_accessors import AbstractDataAccessor
+
+
+class RowTableDataModel(AbstractDataModel):
+    """ A data model that presents a list of objects as rows.
+    """
+
+    #: A sequence of objects to display in columns.
+    data = Any()
```
kitchoi, line 29: Related to the comment about observing mutation on a trait container: It is important that this is not a container trait, otherwise the data would be copied, rather than referenced. We want the container to be referenced. Perhaps we should add a comment as a reminder, because this is quite tempting otherwise to turn this trait into `Union(List, Set, Dict)` in the future.

examples/data_view/row_table_example.py
```diff
@@ -0,0 +1,123 @@
...
+        value_type=TextValue(),
+    ),
+]
+
+
+class MainWindow(ApplicationWindow):
+    """ The main application window. """
+
+    #: A collection of People.
+    data = List(Instance(Person))
```
kitchoi, line 75: It is not a use case for this example, but if this data is to be mutated, we will need to observe mutations on this list as well.

pyface/data_view/data_models/row_table_data_model.py
```diff
@@ -0,0 +1,126 @@
...
+
+    #: An object which describes how to map data for the row headers.
+    row_header_data = Instance(AbstractDataAccessor)
+
+    #: An object which describes how to map data for each column.
+    column_data = List(
+        Instance(AbstractDataAccessor),
+        comparison_mode=ComparisonMode.identity,
```
kitchoi, line 37: Just a comment: This comparison mode setting won't be necessary on the next release of traits (if all goes to plan, that is.)

pyface/data_view/data_models/row_table_data_model.py
kitchoi: Actually, could we use `Instance(collections.abc.Sequence)` here? 🤔

corranwebster: I haven't tested it, but I think a `DataFrame` out to be a valid thing to go here, for example (uncertainty is because I'm not sure if they've changed the indexing API in the last year or so).  Not sure if it is a `Sequence`.

pyface/data_view/data_models/row_table_data_model.py
corranwebster: I like `@observe(trait("data").list_items(optional=True))`, and it is indeed proving the value of the design when it allows you to do things you didn't know you needed to be able to do!

pyface/data_view/data_models/tests/test_row_table_data_model.py
```diff
@@ -0,0 +1,247 @@
...
+        with self.assertTraitChanges(self.model, "values_changed"):
+            self.model.row_header_data = AttributeDataAccessor(attr='b')
+        self.assertEqual(
+            self.values_changed_event.new,
+            ((), (), (), ())
+        )
+
+    def test_row_header_data_values_updated(self):
+        with self.assertTraitChanges(self.model, "values_changed"):
+            self.model.row_header_data.updated = (self.model.row_header_data, 'value')
```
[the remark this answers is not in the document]
corranwebster: No, the table model only cases about the event being fired with a valid payload, in this test we don't care about what caused the event to be fired, only that it was.  The `updated` trait is a specified part of the ABC interface.

pyface/data_view/data_models/row_table_data_model.py
corranwebster: But yes, it is part of the design that users may need to manually fire updates if they aren't working with Trait-y things.

pyface/data_view/data_models/data_accessors.py
```diff
@@ -0,0 +1,168 @@
...
+    index = Int()
+
+    def get_value(self, obj):
+        try:
+            return obj[self.index]
+        except IndexError:
+            return None
+
+    def can_set_value(self, obj):
+        return 0 <= self.index < len(obj)
```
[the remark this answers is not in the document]
corranwebster: I'd suggest that we say non-negative for now, and if someone comes up with a valid use case for negative indices we would re-assess.

pyface/data_view/data_models/row_table_data_model.py
corranwebster: *sigh* Of course it doesn't work out of the box if you use a regular list, because it doesn't get copied into a `TraitsList`.

We need something better than the `List` trait in Traits.

pyface/data_view/data_models/row_table_data_model.py
corranwebster: And for the record `isinstance(DataFrame, Sequence)` is false.

kitchoi: I checked, pandas DataFrame is not a sequence, e.g. it does not implement `__reversed__`.
Looking at this model, looks like all this model needs are `__len__` and `__getitem__`.

Checking for `collections.abc.Sized` gives us `__len__`, and `DataFrame` does pass this test.
However for `__getitem__`, the next available thing is `collections.abc.Sequence`, which captures more than we need and DataFrame does not fulfil the check.

It might be worth documenting that `data` is required to support `__len__` and `__getitem__(int)`

kitchoi: > And for the record isinstance(DataFrame, Sequence) is false.

You beat me to it :)

corranwebster: I played around a bit with getting `DataFrame` working, and it really requires use of `iloc` rather than direct indexing, and given that `DataFrame` can be hierarchical that is pushing me towards `DataFrame` not being supported here.

So targets are `Sequence`-y sorts of things (although as you note, we don't need `__reversed__`) like lists, tuples, 1D or 2D arrays (and "array-like").  I can't think of anything which has `__len__` and `__getitem__` but not `__reversed__`, so may be `Instance(Sequence)` _is_ good enough?

kitchoi: Yea, sounds good to me!

pyface/data_view/data_models/row_table_data_model.py
kitchoi: Ah you are referring to the `_data_default` returning a Python list. Maybe the default has to be an immutable sequence then? This would avoid causing confusion. If one needs a TraitList for `data`, assign one to it.

pyface/data_view/data_models/row_table_data_model.py
corranwebster: Arg.  `Array` isn't a `Sequence` either.

kitchoi: OK, let's do duck typing here and just specify what methods are required for this model then.
(Just found an interesting issue on numpy's repo about this, it is still an open issue.)

pyface/data_view/data_models/data_accessors.py
```diff
@@ -0,0 +1,168 @@
...
+
+    This is suitable for use with Python objects, including HasTraits
+    classes.
+    """
+
+    #: The extended attribute name of the trait holding the value.
+    attr = Str()
+
+    def get_value(self, obj):
+        return xgetattr(obj, self.attr, None)
```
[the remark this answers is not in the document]
corranwebster: I will think about this.  The current expectation of the API is that there will be no errors returned by get methods, because early on I ran into issues with performance using exception handling higher up the stack.  This may have been due to relative likelihoods of exceptions.

But I think I'd like to defer that investigation to a separate issue.

I will look into some sort of logged message or warning if there is a missing attribute, so at least it isn't silenced.

pyface/data_view/data_models/data_accessors.py
```diff
@@ -0,0 +1,168 @@
...
+
+    @observe('attr')
+    def attr_updated(self, event):
+        self.updated = (self, 'value')
+
+    def _title_default(self):
+        # create human-friendly version of extended attribute
+        attr = self.attr.split('.')[-1]
+        title = attr.replace('_', ' ').title()
+        return title
```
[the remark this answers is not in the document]
corranwebster: Empty string is handled correctly (and has to be - empty string is a valid value).

pyface/data_view/data_models/data_accessors.py
```diff
@@ -0,0 +1,168 @@
...
+    #: header.
+    title_type = Instance(AbstractValueType, factory=TextValue)
+
+    #: The value type of the data accessed.
+    value_type = Instance(AbstractValueType)
+
+    #: An event fired when accessor is updated update.  The payload is
+    #: a tuple of the accessor info and whether the title or value changed
+    #: (or both).
+    updated = Event
```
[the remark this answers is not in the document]
corranwebster: I thought about this somewhat, and did specifically consider what you are siggesting.  There are two reasons for the current design:
- the data accessor is inspired by the `RowInfo` class in the column example, and when the column model is implemented not as a library, I'd like it to use data accessors as well (although there'll probably be some need for modification to handle hierarchical accessors): https://github.com/enthought/pyface/blob/master/examples/data_view/column_data_model.py#L25
- the potential for hierarchical accessors in the future is why I am returning the object; that was needed to properly handle updates in the column model. That could be dropped for now, but I think I'd still want to return a tuple.
- What I don't want is a proliferation of event traits, just generally as a design, I'd prefer a single "updated" event with a payload.

I expect that the data accessor class will change a bit as more and more models are implemented and we converge on what the "right" abstraction is for declaring how to get things from underlying data objects.

pyface/data_view/data_models/data_accessors.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+
+    #: The extended trait name of the trait holding the value.
+    index = Int()
+
+    def get_value(self, obj):
+        try:
+            return obj[self.index]
+        except IndexError:
+            return None
```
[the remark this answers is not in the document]
corranwebster: Same comment here - I'd like to investigate as a separate issue.  I will make sure that it isn't silent, however.

pyface/data_view/data_models/data_accessors.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def set_value(self, obj, value):
+        if not self.attr:
+            raise DataViewSetError(
+                "Attribute is not specified for {!r}".format(self)
+            )
+        xsetattr(obj, self.attr, value)
+
+    @observe('attr')
+    def attr_updated(self, event):
```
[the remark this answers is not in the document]
corranwebster: This wasn't intentionally public, and doesn't need to be.

pyface/data_view/data_models/row_table_data_model.py
```diff
@@ -0,0 +1,126 @@
...
+        self.structure_changed = True
+
+    @observe('row_header_data')
+    def _update_row_header_data(self, event):
+        self.values_changed = ((), (), (), ())
+
+    @observe('row_header_data:updated')
+    def _update_row_header_data_event(self, event):
+        if event.new[1] == 'value':
+            self.values_changed = ((0,), (), (len(self.data) - 1,), ())
```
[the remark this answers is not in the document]
corranwebster: Yep - it's a bug.  It just needs a guard for empty data.  If there are no rows, then no values are changed, so the event does not need to fire.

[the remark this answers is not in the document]
corranwebster: I've opened an issue fo the `ArrayDataModel`.

pyface/data_view/data_models/row_table_data_model.py
corranwebster: No, I was referring to my attempts to write a test for the modified behaviour which weren't working because I was assigning a regular list expecting Traits to magically turn it into a trait list - just a brain-fart on my part.

However, returning an empty `TraitList` as the default may not be a bad idea.

pyface/data_view/data_models/row_table_data_model.py
corranwebster: For now, due to https://github.com/enthought/traits/issues/1277 I'm leaning towards `Instance(Sequence)` and we can expand as needed (or people can subclass).

pyface/data_view/data_models/data_accessors.py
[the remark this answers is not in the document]
corranwebster: On customizing the default value:

- the `value_type` can be customized to return something different (particularly for the text value), and that's a better place for it, I think.
- people can always write their own subclasses

So I'm not too concerned about that right now.

pyface/data_view/data_models/row_table_data_model.py
kitchoi: > However, returning an empty TraitList as the default may not be a bad idea.

Yes that would work too. What makes me very slightly worried is that we could be calling the constructor of `TraitList` directly, which to me is closer to playing with implementation details of Traits. The constructor of `TraitList` is not yet a published public API (I think). But maybe it was intended to be a published public API at the end, then this could be a way to test the API within ETS before making it more public?

pyface/data_view/data_models/row_table_data_model.py
kitchoi: SGTM!

pyface/data_view/data_models/data_accessors.py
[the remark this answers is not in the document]
kitchoi: > I will think about this. The current expectation of the API is that there will be no errors returned by get methods, because early on I ran into issues with performance using exception handling higher up the stack. This may have been due to relative likelihoods of exceptions.
>
> But I think I'd like to defer that investigation to a separate issue.

Yep. Let's open a separate issue then. It's probably a good idea to consider this across all the objects we have.

> I will look into some sort of logged message or warning if there is a missing attribute, so at least it isn't silenced.

Yes please.

pyface/data_view/data_models/data_accessors.py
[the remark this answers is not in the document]
kitchoi: > the data accessor is inspired by the RowInfo class in the column example, and when the column model is implemented not as a library, I'd like it to use data accessors as well

Because it was written as an example that might need to be updated later, I did not review it at the same level as I would for library code :)  The example is preliminary, we should not need to veto a design based on that.

> the potential for hierarchical accessors in the future is why I am returning the object

The object is available in the `event.object` attribute, so we still have access to it.

> What I don't want is a proliferation of event traits, just generally as a design, I'd prefer a single "updated" event with a payload.

Actually..., why not? :) If we really need a catch-all, general "updated" event, we can add one that observes the more specific events. The use case here seems to suggest that we need the more specific events, rather than a general one and then differentiate them again.

On the other hand, this event firing logic is hidden from the public API (I think?). So I think we can do with deferring this. There are currently only two places doing this `if event.new[1] == 'value'` check. If more of this pattern occurs, then we can consider splitting the event up.

pyface/data_view/data_models/row_table_data_model.py
corranwebster: `TraitList` won't be part of the public API until `TraitHandler`s are fully removed, due to the name collision with the `TraitList` trait handler.  However it is something that I would like to eventually be public.

It may be better to spend the effort prototyping the "something better than the `List` trait," however.

corranwebster:
I think everything is addressed, if not completely resolved, and tests are passing.

Error handling for `get_value` and friends will be in a separate PR.

corranwebster:
The two issues above are addressed.

rahulporuri (approved):

PR #666, opened 2020-08-12
Use established mechanisms to skip tests incompatible with Traits 6.0

### LINKED ISSUE #662: Revert #656 on master branch or improve skipping of tests
Having a look at the awkwardness of avoiding problems by skipping tests with Traits 6.0, I would strongly recommend either we:

- revert #656 on master
- use the unittest test loader mechanisms to detect Traits 6.0 and skip over the tests in data_view modules (eg. inject something here https://github.com/enthought/pyface/blob/master/pyface/__init__.py#L33)

I'm generally in favour of the first as long as we are good about maintenance releases, as I think it just adds complexity and makes implicit promises that I'm not sure are worth the extra effort.

corranwebster:
This PR adds the `data_view` packages to the excluded tests regex in the case of a Traits 6.0-based environment, using the same mechanisms as for skipping tests for wx and qt when inappropriate.

It also cleans up some dead code left-over from nose-based testing.

This doesn't revert any changes that were made to the test code for the data views - that can be in a separate PR.

Fixes #662

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -107,7 +107,7 @@ extra_dependencies = {
     "pyside": {"pyside"},
     # XXX once pyside2 is available in EDM, we will want it here
     "pyside2": set(),
-    "pyqt": {"pyqt<4.12"},  # FIXME: build of 4.12-1 appears to be bad 
+    "pyqt": {"pyqt<4.12"},  # FIXME: build of 4.12-1 appears to be bad
     "pyqt5": {"pyqt5"},
     # XXX once wxPython 4 is available in EDM, we will want it here
     "wx": set(),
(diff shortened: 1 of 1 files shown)


kitchoi:
@corranwebster Is this intending to supersede https://github.com/enthought/pyface/pull/665?

kitchoi:
The tight coupling between pyface load_tests and `etstool.py` might be making the repository unfriendly to contributors who don't necessarily use EDM for development purposes.

corranwebster:
@kitchoi oops no - I didn't see your PR, I just quickly pushed what I threw together to get the tests to run for my other PR over coffee this morning.  I'll close this as yours is more comprehensive.

kitchoi:
No problem. Thank you!

PR #667, opened 2020-08-13
Mention in docstrings that examples aren't implemented on qt backend 

### LINKED ISSUE #647: NotImplementedError in explorer.py and multi_toolbar_window.py examples on qt
When running the explorer.py example located in (https://github.com/enthought/pyface/tree/master/examples ) I see 

<details> 

```
(qt-env-pyface)~/Desktop/pyface/examples aayres  (master) $ python explorer.py 
  File "explorer.py", line 189, in <module>
    window.open()
  File "/Users/aayres/.edm/envs/qt-env-pyface/lib/python3.6/site-packages/pyface/i_window.py", line 203, in open
    self._create()
  File "/Users/aayres/.edm/envs/qt-env-pyface/lib/python3.6/site-packages/pyface/ui/qt4/application_window.py", line 134, in _create
    contents = self._create_contents(self.control)
  File "/Users/aayres/.edm/envs/qt-env-pyface/lib/python3.6/site-packages/pyface/split_application_window.py", line 39, in _create_contents
    return self._create_splitter(parent)
  File "/Users/aayres/.edm/envs/qt-env-pyface/lib/python3.6/site-packages/pyface/ui/qt4/split_widget.py", line 60, in _create_splitter
    splitter.addWidget(self._create_lhs(splitter))
  File "explorer.py", line 74, in _create_lhs
    return self._create_file_tree(parent, os.path.abspath(os.curdir))
  File "explorer.py", line 143, in _create_file_tree
    filters=[AllowOnlyFolders()],
  File "/Users/aayres/.edm/envs/qt-env-pyface/lib/python3.6/site-packages/pyface/base_toolkit.py", line 184, in __init__
    raise NotImplementedError(msg % (toolkit, package, name))
NotImplementedError: the qt4 pyface.ui.qt4 backend doesn't implement viewer.tree_viewer:TreeViewer
```

</details>

If this is expected, then the doc string should mention that this example is only meant to be run on wx

Effectively the same thing occurs for the multi_toolbar_window.py example as well

aaronayres35:
fixes #647 

Also, there was a deprecation warning when running explorer.py on wx that I tried to fix here.  It probably could have been in its own PR, but it was so small I just added it here.  

Also as a side note I don't really understand what exactly the multi_toolbar_window.py is really supposed to demonstrate.  It also seems like the window isn't displaying correctly.  

Screen Shot 2020-08-13 at 4 12 47 PM

Notice the "Foo" and "Bus" seem to overlap.  I will open a separate issue for this.

### DIFF
--- a/examples/explorer.py
+++ b/examples/explorer.py
@@ -8,8 +8,9 @@
 #
 # Thanks for using Enthought open source!
 
-""" A file explorer example. 
-    Note: This demo only works on the wx backend. """
+""" A file explorer example.
+    Note: This demo only works on the wx backend.
+"""
 
 
 import os
--- a/examples/explorer.py
+++ b/examples/explorer.py
@@ -8,7 +8,8 @@
 #
 # Thanks for using Enthought open source!
 
-""" A file explorer example. """
+""" A file explorer example. 
+    Note: This demo only works on the wx backend. """
 
 
 import os
--- a/examples/multi_toolbar_window.py
+++ b/examples/multi_toolbar_window.py
@@ -8,8 +8,9 @@
 #
 # Thanks for using Enthought open source!
 
-""" Mulit-tool bar example. 
-    Note: This demo only works on the wx backend. """
+""" Mulit-tool bar example.
+    Note: This demo only works on the wx backend.
+"""
 
 
 import os, sys
--- a/examples/multi_toolbar_window.py
+++ b/examples/multi_toolbar_window.py
@@ -8,7 +8,8 @@
 #
 # Thanks for using Enthought open source!
 
-""" Mulit-tool bar example. """
+""" Mulit-tool bar example. 
+    Note: This demo only works on the wx backend. """
 
 
 import os, sys
--- a/pyface/ui/wx/viewer/table_viewer.py
+++ b/pyface/ui/wx/viewer/table_viewer.py
@@ -310,12 +310,12 @@ class _Table(wx.ListCtrl):  # (ULC.UltimateListCtrl):#
 
         # Set up attributes to show alternate rows with a different background
         # colour.
-        self._even_row_attribute = wx.ItemAttr()
+        self._even_row_attribute = wx.ListItemAttr()
         self._even_row_attribute.SetBackgroundColour(
             self._viewer.even_row_background
         )
 
-        self._odd_row_attribute = wx.ItemAttr()
+        self._odd_row_attribute = wx.ListItemAttr()
         self._odd_row_attribute.SetBackgroundColour(
             self._viewer.odd_row_background
         )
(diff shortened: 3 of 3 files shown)


examples/explorer.py
```diff
@@ -8,7 +8,8 @@
 #
 # Thanks for using Enthought open source!
 
-""" A file explorer example. """
+""" A file explorer example. 
+    Note: This demo only works on the wx backend. """
```
kitchoi, line 12: Ultra-Nit that can be ignored: For consistency in this code base, if there are more than one line, the closing `"""` should be in the next line on its own.

aaronayres35:
> Agreed that these are all small changes.
> 
> However in this case I'd probably still separate the two for the following reasons:
> (1) These fixes touch different files.
> (2) The fix in wx `table_viewer.py` is a bug fix: It is a deprecation warning now, which means it will become an error in the future.
> 
> If the `examples` files are moved on master, and if we want to backport the fix for table_viewer.py to the long-running maintenance branch where the examples have not been moved, we are going to have to manually create a separate commit in order to just backport the fix in table_viewer.py without touching the examples.
> 
> (Is there a log on the deprecation warning message? It would be useful for future reference and for search.)

I have removed the change from this PR.  Also, looking back on it even with the change undone I am not seeing the deprecation warning.  I realized yesterday I was working in a wx environment that was created using the etstool for traitsui rather than pyface while working in pyface.   When I run explorer.py in that environment I see:
```
/Users/aayres/Desktop/pyface/pyface/ui/wx/viewer/table_viewer.py:313: wxPyDeprecationWarning: Using deprecated class. Use ItemAttr instead
  self._even_row_attribute = wx.ListItemAttr()
/Users/aayres/Desktop/pyface/pyface/ui/wx/viewer/table_viewer.py:318: wxPyDeprecationWarning: Using deprecated class. Use ItemAttr instead
  self._odd_row_attribute = wx.ListItemAttr()
```
However, in the correct pyface wx environment there is now warning.

kitchoi:
May I ask how you ran the tests? I am wondering if it is the difference caused by pyface `etstool.py` not having `-W default` set for running its tests. We should probably set it.  It helps capturing deprecation warnings early (though the test suite would be noisier, but for good reasons I hope).

kitchoi:
> May I ask how you ran the tests? I am wondering if it is the difference caused by pyface etstool.py not having -W default set for running its tests. We should probably set it. It helps capturing deprecation warnings early (though the test suite would be noisier, but for good reasons I hope).

Sorry, please ignore the above. That's probably not it.
Maybe this is why:
https://github.com/enthought/pyface/blob/facbc9d51979b4cfbbcc682c2aeb1a74e2205636/etstool.py#L202

aaronayres35:
> May I ask how you ran the tests?

I ran `python etstool.py install --toolkit=wx --environment=wx-env-_____` and then simply `python explorer.py` after shelling into that environment.  However in one case the env was created by that command within traitsui and the other within pyface.

kitchoi:
In that case it might worth opening issue or PR for the deprecation warning: it is probably legit, we just haven't run into it yet as the development environment has pinned an older version of wxPython.

aaronayres35:
> the development environment has pinned an older version of wxPython.

Just opened a PR #670

PR #669, opened 2020-08-14
Flake8 etstool.py

kitchoi:
There was a trailing whitespace that my IDE keeps trying to fix it for me whenever I save the file.
This PR lints the entire file for flake8 compliance.

I added `noqa` to silence flake8 in cases where it is better/readable to keep the integrity of a command than trying to break it up.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -107,7 +107,7 @@ extra_dependencies = {
     "pyside": {"pyside"},
     # XXX once pyside2 is available in EDM, we will want it here
     "pyside2": set(),
-    "pyqt": {"pyqt<4.12"},  # FIXME: build of 4.12-1 appears to be bad 
+    "pyqt": {"pyqt<4.12"},  # FIXME: build of 4.12-1 appears to be bad
     "pyqt5": {"pyqt5"},
     # XXX once wxPython 4 is available in EDM, we will want it here
     "wx": set(),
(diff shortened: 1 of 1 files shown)


etstool.py
```diff
@@ -199,12 +203,12 @@ def install(edm, runtime, toolkit, environment, editable, source):
...
         if sys.platform == "darwin":
             commands.append(
-                "{edm} run -e {environment} -- python -m pip install wxPython<4.1"
+                "{edm} run -e {environment} -- python -m pip install wxPython<4.1"  # noqa: E501
             )
         elif sys.platform == "linux":
             # XXX this is mainly for TravisCI workers; need a generic solution
             commands.append(
-                "{edm} run -e {environment} -- pip install -f https://extras.wxpython.org/wxPython4/extras/linux/gtk3/ubuntu-14.04/ wxPython<4.1"
+                "{edm} run -e {environment} -- pip install -f https://extras.wxpython.org/wxPython4/extras/linux/gtk3/ubuntu-14.04/ wxPython<4.1"  # noqa: E501
```
rahulporuri, line 211: this is tangential but we're not using the ubuntu 14.04 images on travis (i think xenial is 16.04) at the moment so i don't know if we should be using this url to download the expython wheels.

rahulporuri: yep xenial is 16.04 - https://docs.travis-ci.com/user/reference/linux#overview

kitchoi: Yep, probably needs a different image.  Not in this PR though.

PR #670, opened 2020-08-14
replacing ListItemAttr with ItemAttr to avoid deprecation warning

aaronayres35:
See comments on #667.

We do not currently see the deprecation warning when using a wx environment created by the etstool in pyface as the development environment has pinned an older version of wxPython.
However, when using an environment with newer wx, there was the following deprecation warning (found when running the explorer.py example):
```
/Users/aayres/Desktop/pyface/pyface/ui/wx/viewer/table_viewer.py:313: wxPyDeprecationWarning: Using deprecated class. Use ItemAttr instead
  self._even_row_attribute = wx.ListItemAttr()
/Users/aayres/Desktop/pyface/pyface/ui/wx/viewer/table_viewer.py:318: wxPyDeprecationWarning: Using deprecated class. Use ItemAttr instead
  self._odd_row_attribute = wx.ListItemAttr()
```

### DIFF
--- a/pyface/ui/wx/viewer/table_viewer.py
+++ b/pyface/ui/wx/viewer/table_viewer.py
@@ -310,12 +310,12 @@ class _Table(wx.ListCtrl):  # (ULC.UltimateListCtrl):#
 
         # Set up attributes to show alternate rows with a different background
         # colour.
-        self._even_row_attribute = wx.ListItemAttr()
+        self._even_row_attribute = wx.ItemAttr()
         self._even_row_attribute.SetBackgroundColour(
             self._viewer.even_row_background
         )
 
-        self._odd_row_attribute = wx.ListItemAttr()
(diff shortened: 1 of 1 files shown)


aaronayres35:
CI is failing, I'm not sure why.  I will try to look into this later today

kitchoi:
Thank you @aaronayres35!

corranwebster:
Wx failure appears unrelated, but we've been getting a lot of linux/wx failures on multiple PRs.  I've set the test to re-run, but if this continues, we may need to dig in a bit (or mark linux/wx test suites as being allowed to fail for CI to pass).

kitchoi:
@rahulporuri spotted a version mismatch between Ubuntu version on Travis and the one targeted by the wxPython wheel used for CI. It might be related (or it might not be...)
https://github.com/enthought/pyface/pull/669#discussion_r470762909

PR #675, opened 2020-08-17
Fix Wx/linux clipboard test failures

### LINKED ISSUE #674: Segmentation faults in wxPython/Linux CI
This is happening about 50% of the time.  The point of failure is consistent:
```
test_set_object_data (pyface.tests.test_clipboard.TestClipboard) ... Command "coverage run -p -m unittest discover -v pyface" exited with segmentation fault (-11)
```
But I suspect that the actual cause is unrelated.

corranwebster:
Fixes #674 

Under X11-based systems,  Wx can pull values to be pasted either from the clipboard, or from the current ("primary") selection.  This change ensures that the Pyface clipboard is always getting data from the clipboard, and never from the current selection.

This is implemented via a context manager which ensures that the wider application state controlling the source of paste data is unmodified, minimizing the likelihood of problems in the wider application (particularly if Pyface is being used as a library).  The context manager will do nothing if called from a non-X11 system.


corranwebster:
Arg - branched from wron place; should. have done. it from master.

PR #677, opened 2020-08-17
Update Wx CI

corranwebster:
This updates the Wx CI to use the wxPython package for Ubuntu 16.04, which is now the default version of Ubuntu used in Travis CI (updated from 14.04 when wxPython testing was last updated).

It also turns on Python's faulthandler when running tests via etstool, so segfaults and other hard crashes should give better feedback.

This needs #676 to be merged first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ addons:
     - pulseaudio
     - libpulse-mainloop-glib0
     # Wx dependencies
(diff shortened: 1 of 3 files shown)


corranwebster:
Oops - looks like I missed a cherry-picked commit, but that's fine, as it probably should be in a separate PR anyway.

Issue #556, opened 2020-06-29
Sporadic test failure with test_timer/test_repeat on Mac OSX and Qt


kitchoi:
The following test failure occurs **occasionally** on the Mac OSX build:

```
======================================================================
FAIL: test_repeat (pyface.timer.tests.test_timer.TestTimer)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/travis/.edm/envs/pyface-test-3.6-pyside2/lib/python3.6/site-packages/pyface-7.1.0.dev22-py3.6.egg/pyface/timer/tests/test_timer.py", line 425, in test_repeat
    self.assertEqual(handler.count, 4)
AssertionError: 5 != 4
```

corranwebster:
Looking in to this a bit, the logic being tested is here: https://github.com/enthought/pyface/blob/4b3b15f95fdb100e8ec2c3860f259c6997c6cb93/pyface/timer/i_timer.py#L186-L210

It looks like the `_perform` callback can be called when the number of remaining repeats is 0 or less.  I'm guessing that we're getting some sort of race condition between the `stop` taking effect and the next iteration of the timer.

Potential fixes to increase reliability:

- check the `_active` flag at the start of the `perform` method and exit early if it is false.
- putting a check for remaining repeats to guard the call of the `_perform` method

Some care needs to be taken with the logic, as we want to call the `stop()` on the last iteration _after_ the `_perform` call, and we want to try to be careful to only call the toolkit `stop` once if we can - I don't know that we can assume calls to toolkit stop are idempotent.

kitchoi:
For reference, the test that is failing has the Timer's `expire` and `repeat` set to None. So I am not sure why `repeats` can be 0 or less, maybe I was looking at the wrong test?  I am looking at this one:
`pyface.timer.tests.test_timer.TestTimer.test_repeat`

If I just run the test many times, each as a separate process, I could not reproduce it.

But if I rewrite the test like this :

```python
    def test_repeat(self):
        for i in range(40):
            handler = ConditionHandler()

            start_time = perf_counter()
            timer = Timer(250, handler.callback)
            try:
                self.assertTrue(timer.IsRunning())
                self.event_loop_helper.event_loop_until_condition(
                    handler.called_n(4),
                )
                self.assertTrue(timer.IsRunning())
            finally:
                timer.Stop()

            self.assertFalse(timer.IsRunning())
```

[...]

PR #678, opened 2020-08-18
Drive-by clean-up of wx clipboard code

corranwebster:
Adds a closing context manager, plus adds logging for load failures and a few additional clean-ups.  Purely opportunistic since we have been in the code for this.


kitchoi:
CI is failing due to #556 again, i.e. orthogonal...

PR #680, opened 2020-08-20
Better error handling for DataViews

### LINKED ISSUE #679: Handling Failure to Get Data in the DataView Framework
This is a follow-up to comments in various PRs about what to do if there is an error in a data model for a data view:

- currently there is no error-checking: values _must_ return valid values, or you are likely to get a crash.
- raw data values will likely be filtered through multiple code paths (eg. to get text, colors, edit values, etc.), so a failure at the lowest level will likely cause failure across all data channels.
- any error which occurs is likely to be repeated across multiple cells (eg. all cells in a column)
- the combination of the above means that when a failure occurs, it will be multiplied if handled at the highest level
- effectively handling exceptions is expensive, logging involves writing to screen or disk which is significantly slower than simply letting things pass.  Coupled with the multiplicative nature of errors, this can mean that the application performance can severely degrade
- writing essentially the same superficial error message to the logs hundreds or thousands of times may obscure less frequent but more important root problems

Proposal:

- introduce a new error type that can be used to signal "expected failure in getting" which does not require logging (or where the logging can be controlled at the point of generation); this should be used sparingly
- for any other type of error we log, to ensure things are properly recorded for the developer, and then raise.  This is likely to crash the app, but this is no different than if the same programming error occurred elsewhere in the app.
- data models and value types included in Pyface will not try to handle every possible problem robustly, rather they will work on the presumption of developer correctness within the parameters of their design (eg. if a developer says that you want to get an attribute, and there isn't one, then don't try to second-guess).
- custom data models by developers can do whatever they want, but should expect poor performance if they do a lot of exception logging.

corranwebster:
This adds a `DataViewGetError` exception and better handling of errors in the data view infrastructure when getting a value as discussed in #679.

Fixes #679 .


pyface/data_view/data_view_errors.py
```diff
@@ -0,0 +1,27 @@
...
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+""" Provide an Exception classes for the the DataView code.
+"""
+
+
+class DataViewError(ValueError):
```
kitchoi, line 15: Why does this subclass `ValueError`?  This means if we do `except ValueError`, all errors related to DataView will be captured. This is a design choice that would be difficult to revert.  Is there a particular reason why we should categorize all data view errors as a type of ValueError? If not, it would be more flexible to defer this by not subclassing `ValueError`. e.g. `TraitError` just subclasses `Exception`.

kitchoi (commented):
LGTM. One question on the exception type all the data view exceptions subclass from. I think I overlooked this earlier in the bigger PR that introduces it.

(The code in the Qt and Wx implementations are untested, but those modules are uncovered anyway because they are still meant to be scaffolding code to be replaced.)

pyface/data_view/data_view_errors.py
corranwebster: No strong reason: `ValueError` is described as an error type to return when you have an inappropriate value, but no more specific exception applies (eg. `IndexError`) which fits this situation, and `ValueError` is not a leaf of the `Exception` hierarchy.  In particular, the `DataViewSetError` is often a proxy for a different `ValueError` generated by eg. a failure to evaluate text to an appropriate data value, or setting a data value of the wrong type.

So that's why the original choice was made.

By having a base class for all the `DataViewErrors`, it makes it easy to filter out from other `ValueError`s if someone really cares about it.

And if it is wrong, I don't see why it is a particularly difficult to change.

Let's keep it as-is for now.

kitchoi: I don't think I follow that, e.g. why `DataViewSetError` can't be defined as `DataViewSetError(DataViewError, ValueError)` whereas `DataViewError` is simply `DataViewError(Exception)`. This will allow the flexibility to have specialized exception that is not related to `ValueError`, e.g. `DataViewGetError(DataViewError, LookupError)`.  Up to you if you don't want this. 

>  I don't see why it is a particularly difficult to change.

I guess I am thinking about these exceptions being used in try-except in code outside of pyface. Then it would be backward incompatible to change it one way or another.

kitchoi (approved):

PR #687, opened 2020-08-28
Update copyright header

rahulporuri:
`pyface.util.guisupport` is taken from the IPython codebase, which has copyright till 2010 and we can only claim copyright for changes during/after 2010.

(It might be worth while to see what changes have happened to the original `guisupport` module to see if we can incorporate any meaningful changes back into the module.)

### DIFF
--- a/pyface/util/guisupport.py
+++ b/pyface/util/guisupport.py
@@ -1,4 +1,4 @@
-# (C) Copyright 2005-2020 Enthought, Inc., Austin, TX
+# (C) Copyright 2010-2020 Enthought, Inc., Austin, TX
 # All rights reserved.
 #
 # This software is provided without warranty under the terms of the BSD


corranwebster:
We are putting the same copyright header on all Enthought-origin files: 2005-2020 is for the codebase as a whole.

rahulporuri:
> We are putting the same copyright header on all Enthought-origin files: 2005-2020 is for the codebase as a whole.

ohh. my understanding was that we can only claim copyright for vendored code from the date when it was vendored.

corranwebster:
For vendorized code, we have too include the original copyright as well, as an additional copyright line.

corranwebster:
So this line https://github.com/enthought/pyface/blob/0cdceef835aa4d30a3bcb288a3e4e35150f2c8ad/pyface/util/guisupport.py#L69 should be in the copyright header and we should remove the secondary header.

PR #688, opened 2020-09-01
Add DataWrapper classes to support abstracted DnD and clipboard

corranwebster:
To support drag and drop for data views, we need an abstraction that wraps Qt `QMimeData` and Wx `DataObject` classes, both of which store serialized bytes tagged by mimetype.

This PR provides a generic wrapper for both, along with a simple `DataFormat` class for handling different mimetypes.  Planned design for drag and drop support is that when a drag is started, the data view will create a wrapper object and supply data in appropriate formats; and when an object is dropped the data view will supply a list of data formats that are supported and they will be tried in order.  The deserialized data will then be supplied to `set_data` or similar methods on the class.

Cut, Copy and Paste will work similarly, when supported, but that is low priority.

Note that both Wx and Qt provide special ways of getting text, HTML, image, etc. data.  We don't attempt to wrap those, but it is possible that we may choose to in the future; similarly we may need to tweak these classes a little depending on what real-world data we actually see being dragged and dropped (eg. precise mimetypes, etc.).

This resolves part of #699.

### DIFF
--- a/pyface/data_view/data_wrapper.py
+++ b/pyface/data_view/data_wrapper.py
@@ -1,0 +1,13 @@
+# (C) Copyright 2005-2020 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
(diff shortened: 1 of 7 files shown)


corranwebster:
BTW: will accept suggestions for better names than "DataWrapper"

rahulporuri (approved):
LGTM

PR #690, opened 2020-09-19
Add a new shell click command

rahulporuri:
This PR adds a new `shell` click command - which when run activates the development environment for traits. The usual command line options are available to choose the environment name and the runtime/toolkit version of the development environment. Underneath the hood, the click command just uses the `edm shell` command with the appropriate environment name, which is why the docstring for the click command is _almost_ the same as that of `edm shell`.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -244,6 +244,23 @@ def install(edm, runtime, toolkit, environment, editable, source):
 @click.option("--runtime", default="3.6", help="Python version to use")
 @click.option("--toolkit", default="pyqt", help="Toolkit and API to use")
 @click.option("--environment", default=None, help="EDM environment to use")
+def shell(edm, runtime, toolkit, environment):
+    """ Create a shell into the EDM development environment
+    (aka 'activate' it).
+
+    """
+    parameters = get_parameters(edm ,runtime, toolkit, environment)
+    commands = [
+        "{edm} shell -e {environment}",
+    ]
+    execute(commands, parameters)
+
+
+@cli.command()
+@edm_option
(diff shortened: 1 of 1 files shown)


etstool.py
```diff
@@ -239,6 +239,23 @@ def install(edm, runtime, toolkit, environment, editable, source):
...
+def shell(edm, runtime, toolkit, environment):
+    """ Create a shell into the EDM development environment
+    (aka 'activate' it).
+
+    """
+    parameters = get_parameters(edm ,runtime, toolkit, environment)
+    commands = [
+        "{edm} shell -e {environment}",
+    ]
+    execute(commands, parameters)
```
corranwebster, line 256: As a convenience, perhaps it should check to see if the environment exists, and if it doesn't then it should create it?  This would be particularly useful when using the `--environment` option because `edm shell -e ...` is otherwise more simpler to use.

corranwebster (approved):

etstool.py
rahulporuri: >  because edm shell -e ... is otherwise more simpler to use.

I disagree because the default env itself has a very verbose name `pyface-test-3.6-null` and every so often I make mistakes and get atleast one character wrong when typing the env name.

> As a convenience, perhaps it should check to see if the environment exists, and if it doesn't then it should create it?

I'd prefer leaving it as is because edm is helpful in pointing out that the end doesn't exist. And it's not too much of a hastle changing `etstool shell` to `etstool install`.

That way, we are also consistent with how edm behaves.

PR #692, opened 2020-09-21
Remove border from Qt HeadingText.

### LINKED ISSUE #664: HeadingText has strange border around it on Qt
HeadingText has a border around it on Qt, which seems inconsistent with native UIs. Here is an example:
```Python
from traits.api import HasStrictTraits, Str
from traitsui.api import Heading, Label, Item, VGroup, View


class Model(HasStrictTraits):
    item_1 = Str()
    item_2 = Str()
    item_3 = Str()

    traits_view = View(
        VGroup(
            Heading('Heading'),
            VGroup(
                Item('item_1'),
                Item('item_2'),
                Item('item_3'),
                label='Borderless Group',
                show_border=False,
            ),
            VGroup(
                Item('item_1'),
                Item('item_2'),
                Item('item_3'),
                label='Border Group',
                show_border=True,
            ),
        )
    )


if __name__ == "__main__":
    model = Model()
    model.configure_traits()
```

On Ubuntu the result is:
heading-border

The border is a result of the following line. It looks to be an explicit choice made here rather than something Qt is automatically doing.
https://github.com/enthought/pyface/blob/68e694214a6d09c3d51da3e6488f824be4b67319/pyface/ui/qt4/heading_text.py#L60

Removing that line gives a more expected result:

heading-border-2

corranwebster:
Fixes #664

### DIFF
--- a/pyface/ui/qt4/heading_text.py
+++ b/pyface/ui/qt4/heading_text.py
@@ -57,8 +57,6 @@ class HeadingText(MHeadingText, Widget):
         self.control = QtGui.QLabel(parent)
         self._set_text(self.text)
 
-        self.control.setFrameShape(QtGui.QFrame.StyledPanel)
-        self.control.setFrameShadow(QtGui.QFrame.Raised)
         self.control.setSizePolicy(
             QtGui.QSizePolicy.Preferred, QtGui.QSizePolicy.Fixed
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM

PR #693, opened 2020-09-22
Replace usage of Instance(ImageResource) with Image()

### LINKED ISSUE #691: SplashScreen image should be an Image not an Instance(ImageResource)
See https://github.com/enthought/pyface/blob/33542a4afd6452f08c18756229253b3d384e0da1/pyface/i_splash_screen.py#L31

Using an `Image` trait here would mean that users could pass image names or paths instead of `ImageResource` objects.

corranwebster:
The `Image` trait provides a strict superset of the functionality of `Instance(ImageResource)` and related constructs like `Instance(IImageResource)` - in addition to accepting `None` and `ImageResource` instances it accepts strings that define the location of an image.

This will potentially make it easier to configure application splash screens and similar.

In terms of changes, this PR:

- replaces `Instance(ImageResource)` with `Image`
- leaves default `ImageResource()` values as-is even though they could likely be replaced with the string name of the resource directly.
- cleans up imports appropriately

Fixes #691.


rahulporuri:
Just to confirm - these changes are strictly backwards compatible. Right?

rahulporuri:
Looks like CI is not at all happy and i'm not sure if it's actually because of changes you introduced in this PR

corranwebster:
I was too vigorous about removing imports - hopefully fixed.  The error was weird, though - for whatever reason we weren't seeing the actual cause reported at all and only the failure of the task window to clean itself up.

pyface/dock/dock_sizer.py
```diff
@@ -259,10 +260,10 @@ class DockImages(HasPrivateTraits):
     # ---------------------------------------------------------------------------
 
     # Image for closing a tab:
-    close_tab = Instance(ImageResource, ImageResource("close_tab"))
+    close_tab = Image(ImageResource("close_tab"))
```
rahulporuri, line 263: Is there a specific reason why we can't just use `Image("close_tab")` which IIUC is the same as `Image(ImageResource("close_tab"))`?

corranwebster: The lookup rules for `Image` and `ImageResource` are subtly different (they go up different numbers of containing directories when looking for image directories, IIRC) and `Image` also does caching.

I left them as `ImageResource` instances since then I know the behaviour is the same, to be conservative.

kitchoi: So on the one hand we want people to use `Image()` instead of `Instance(ImageResource)` but then on the other hand we want to discourage setting a non-ImageResource value to it because `Image` is too magical?

corranwebster: No.  It's no more magical than `ImageResource`, it's just _different_.

I would encourage people to use strings directly in new code if it makes sense for them, or to replace old `ImageResource` instances with strings when they migrate, as long as they check that it works correctly for them.

In this PR, I didn't want to check two behaviours at once.

kitchoi: Orthogonal but related in spirit: Yesterday we discussed how unintuitive to read code like this:
```
from library import BaseCalss

class Foo(BaseClass):
    name = "NAME"
```
where the `BaseClass` is defined elsewhere (maybe from a separate library) like this:
```
class BaseClass(HasTraits):
    name = Str()
```
It is not obvious to the developer that `name` there is a trait rather than just a plain class attribute. This increases cognitive load: developers have to spend more time to understand the code and be surprised by the magic.  (Speaking from my own experience, I have been surprised by exactly this behaviour, and that had costed me time trying to understand code that uses this pattern.)

The example above with `Str` is less bad, because `Str` does not transform the given `"NAME"` value.
But for `Image` that does transform the value, the code is going to be even less straightforward to read and understand.

I'd personally in favour of typing a bit more such that the code is easier to read and to understand. Code is written not just for the machine but for the human as well.

kitchoi: To put it differently: The developer who sets the trait to a string _knows_ it gets transformed to something else because of the trait type. But another developer who reads the code may not know that.

kitchoi: This also came up from offline discussion, for the record:

Often when I want to change library code, I want to gather and discover usages of that library feature to understand how it has been used. It is particularly important to decide if a change is backward compatible, and to decide whether a change is worth it. So for example, I can search the enthought GitHub organization for all occurrences of `ImageResource` to see how it has been used.  That's very quick task. But if project codes are not instantiating an instance of `ImageResource` but simply set a string, such usage will escape my search, i.e. I get less information. If I want to gather the same information, I will need to spend a (hell) lot more time to find what objects have declared the `Image` image, where these objects are used (and subclassed) and where the corresponding trait is changed.

rahulporuri:
@corranwebster do you want to resurrect this PR?

rahulporuri (approved):

Issue #533, opened 2020-06-04
New Tree/Table Editor: other data channels


corranwebster:
Implement a way for models to provide colour (eg. colour mapped), image, checkbox, tooltip, etc. information that may be used by tables to display the data.  Since the way to adapt raw data to these other channels is likely to be fairly standard across underlying data models, this should ideally be done in a way that can be composed into the model or on top of it.

Channels that have been added:

- [X] Text
- [X] Editor value
- [X] Image
- [X] Checkbox
- [X] Tooltip
- [ ] Color

PR #694, opened 2020-09-22
DataView checkboxes

corranwebster:
This PR adds support for checkboxes in the data view; and adds a bool value type that uses them.

The bool value type uses a different code path than the editable values because the checkboxes are always interactable and don't require a click to activate an editor.

### DIFF
--- a/pyface/data_view/abstract_value_type.py
+++ b/pyface/data_view/abstract_value_type.py
@@ -28,7 +28,7 @@ from .data_view_errors import DataViewSetError
 
 class CheckState(IntEnum):
     "Possible checkbox states"
-    # XXX in the future this may need a "partial" state
+    # XXX in the future this may need a "partial" state, see Pyface #695
     UNCHECKED = 0
     CHECKED = 1
 
--- a/pyface/data_view/abstract_value_type.py
+++ b/pyface/data_view/abstract_value_type.py
@@ -19,9 +19,18 @@ It is up to the data view to take this standardized data and determine what
 and how to actually display it.
 """
 
+from enum import IntEnum
+
 from traits.api import ABCHasStrictTraits, Event, observe
 
 from .data_view_errors import DataViewSetError
+
+
+class CheckState(IntEnum):
+    "Possible checkbox states"
+    # XXX in the future this may need a "partial" state
+    UNCHECKED = 0
+    CHECKED = 1
 
 
 class AbstractValueType(ABCHasStrictTraits):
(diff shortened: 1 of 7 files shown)


pyface/data_view/abstract_value_type.py
```diff
@@ -19,11 +19,20 @@
...
+from enum import IntEnum
+
 from traits.api import ABCHasStrictTraits, Event, observe
 
 from .data_view_errors import DataViewSetError
 
 
+class CheckState(IntEnum):
+    "Possible checkbox states"
+    # XXX in the future this may need a "partial" state
```
rahulporuri, line 31: I'm not sure what a partial state can mean for a checkbox.

pyface/data_view/abstract_value_type.py
```diff
@@ -226,6 +235,76 @@ def get_image(self, model, row, column):
...
+
+        Returns
+        -------
+        has_check_state : bool
+            Whether or not the value has a checked state.
+        """
+        return False
+
+    def get_check_state(self, model, row, column):
+        """ The textual representation of the underlying value.
```
rahulporuri, line 260: I can't find a better docstring but `The textual representation of the underlying value.` doesn't feel appropriate for a `get_check_state`.

pyface/data_view/abstract_value_type.py
```diff
@@ -226,6 +235,76 @@ def get_image(self, model, row, column):
...
+        Returns
+        -------
+        check_state : CheckState
+            The current checked state.
+        """
+        return (
+            CheckState.CHECKED
+            if model.get_value(row, column)
+            else CheckState.UNCHECKED
+        )
```
rahulporuri, line 283: I'm not sure why the default `get_check_state` doesn't return a `DataViewGetError` like `set_check_state` does.

pyface/data_view/tests/test_abstract_value_type.py
```diff
@@ -85,6 +85,16 @@ def test_get_image(self):
...
 
+    def test_has_check_state(self):
+        value_type = ValueType()
+        result = value_type.has_check_state(self.model, [0], [0])
+        self.assertFalse(result)
+
+    def test_get_check_state(self):
+        value_type = ValueType()
+        result = value_type.get_check_state(self.model, [0], [0])
+        self.assertEqual(result, CheckState.CHECKED)
```
rahulporuri, line 96: can we also add a new trivial test here that test that `set_check_state` raises the expected error?

pyface/data_view/value_types/bool_value.py
```diff
@@ -0,0 +1,109 @@
...
+        -------
+        has_editor_value : bool
+            Whether or not the value is editable.
+        """
+        return False
+
+    def get_text(self, model, row, column):
+        """ The textual representation of the underlying value.
+
+        The default implementation calls str() on the underlying value.
```
rahulporuri, line 48:
```suggestion
```

pyface/data_view/value_types/bool_value.py
```diff
@@ -0,0 +1,109 @@
...
+            The textual representation of the underlying value.
+        """
+        return (
+            self.true_text if model.get_value(row, column) else self.false_text
+        )
+
+    def has_check_state(self, model, row, column):
+        """ Whether or not the value has checked state.
+
+        The default implementation returns False.
```
rahulporuri, line 71:
```suggestion
        The default implementation returns True.
```

pyface/data_view/abstract_value_type.py
corranwebster: When you have a tree view a parent may show a 3rd "partially selected" state when some of the children are checked and some are not.  Visually it usually looks something like this:
image

This is why I am using an enum here rather than simple bool.

pyface/data_view/abstract_value_type.py
corranwebster: This is a copy-pasta.  I will fix.

pyface/data_view/abstract_value_type.py
corranwebster: This is a default implementation that is almost always what you want: if the value has a check state, then it should be checked if the underlying value is truthy and unchecked if it is falsey.

By implementing it on the ABC we save repetition elsewhere (we do similar things for get text and get image, try to have a sane default for the getter).

pyface/data_view/abstract_value_type.py
rahulporuri: Can we put this enhancement into an issue and xref it in the code?

corranwebster:
I think that's everything.

rahulporuri (approved):
LGTM if CI is also happy.

corranwebster:
Partiall resolves #533

PR #703, opened 2020-09-29
Add a PyfaceColor trait to pyface.ui_traits

corranwebster:
This provides a convenience trait `PyfaceColor` that casts tuples, strings, etc to `Color`.

This trait is intended to eventually be a replacement for the existing `Color` trait from TraitsUI.


rahulporuri (approved):
LGTM

Issue #699, opened 2020-09-25
Drag, Drop, Copy and Paste support for Data Views


corranwebster:
We need the ability for users to drag and copy selections out of the data view; and for users to drop and paste values from elsewhere into the data view.

These are all related as they all involve serialization of data to standard MIME formats (eg. text/plain, text/csv, application/json, etc.) as this is what Qt and Wx use to talk to the underlying OS clipboard, etc.

This will require:

- [ ] serialization/deserialzation support for MIME types (#688)
- [ ] hooks to support dragging out values (#689)
- [ ] hooks to allow dropping values into the data view
- [ ] cut, copy, paste, etc. methods

Full implementation of this may require the ability to insert/delete on the _structure_ of the data model, rather than simply changing the values of individual items.  Some of this is encapsulated by #681 which discusses the events that will need to be generated.

PR #704, opened 2020-09-29
DataView: Drag support

corranwebster:
This PR builds on #688 to provide generic support for mime-based data export from DataViews.

These classes and objects are designed to support Drag and Drop, Clipboard operations, and possibly saving of data to files. The AbstractDataExporter class is designed to be subclasses by users who need to support particular formats, selection types and data types within their application. This PR provides two generically useful implementations, one which allows dragging the contents of a single cell (designed to work well with single item selections), the other which provides row-oriented dragging (designed to work well with row-based selections).

The PR also adds some clean-up and generalization of code initially committed in the selection support PRs, which turned out to be useful for more than just selections.

It also adds drag support to the Qt implementations to prove capabilities. Drag support for Wx will be forthcoming in a separate PR.

#688 should be approved and merged before this PR.

This resolves part of #699.

This replaces #689 (it is a rebased version of the same code).


rahulporuri:
@corranwebster can you take a look at the conflicts?

corranwebster:
Arg - I had assumed it was good to g because it looks good on the main Issues page.

pyface/data_view/data_formats.py
```diff
@@ -0,0 +1,293 @@
...
+    ----------
+    data : array-like
+        The array to be serialized.
+
+    Returns
+    -------
+    raw_data : bytes
+        The serialized data as a bytestring.
+    """
+    from numpy import save, atleast_2d
```
rahulporuri, line 260: It looks like we are making numpy an optional dependency of pyface here - and we don't have a quiet way of falling back when numpy isn't available in the environment.

rahulporuri, line 260: Can we follow the usual way we do/use numpy imports here i.e. `import numpy as np; np.save/np.atleast_2d` ?

pyface/data_view/data_formats.py
```diff
@@ -0,0 +1,293 @@
...
+    ----------
+    raw_data : bytes
+        The serialized CSV data as a byte string.
+
+    Returns
+    -------
+    data : list of list of str
+        The data extracted as a list of lists of strings.
+    """
+    from numpy import load
```
rahulporuri, line 281: same comment as above regarding `import numpy as np; np.load`.

pyface/data_view/exporters/item_exporter.py
```diff
@@ -0,0 +1,47 @@
+
```
rahulporuri, line 1: missing copyright header.

pyface/data_view/exporters/item_exporter.py
```diff
@@ -0,0 +1,47 @@
...
+from pyface.data_view.abstract_data_exporter import AbstractDataExporter
+from pyface.data_view.data_view_errors import DataViewGetError
+
+
+class ItemExporter(AbstractDataExporter):
+    """ Export a single item from a data view.
+
+    This is suitable for drag and drop or copying of the content of a single
+    item in a data view.  If passed an multiple items it will fail by
```
rahulporuri, line 10:
```suggestion
    item in a data view.  If passed multiple items, it will fail by
```

pyface/data_view/exporters/item_exporter.py
```diff
@@ -0,0 +1,47 @@
...
+        indices : list of (row, column) index pairs
+            The indices where the data is to be stored.
+
+        Returns
+        -------
+        data : any
+            The data, of a type that can be serialized by the format.
+        """
+        row, column = indices[0]
+        return self.get_value(model, row, column)
```
rahulporuri, line 47: i'm not sure how this is going to raise a `DataViewGetError` as mentioned in the class docstring.

pyface/data_view/exporters/row_exporter.py
```diff
@@ -0,0 +1,50 @@
+from traits.api import Bool
```
rahulporuri, line 1: missing copyright header.

pyface/data_view/exporters/row_exporter.py
```diff
@@ -0,0 +1,50 @@
...
+
+from pyface.data_view.abstract_data_exporter import AbstractDataExporter
+
+
+class RowExporter(AbstractDataExporter):
+
```
rahulporuri, line 7: can you add a class docstring, like with the `ItemExporter`?

rahulporuri: you can move a significant part of the `get_data` docstring here, if you didn't want to write a new docstring.

pyface/data_view/exporters/row_exporter.py
```diff
@@ -0,0 +1,50 @@
...
+            The indices where the data is to be stored.
+
+        Returns
+        -------
+        data : any
+            The data, of a type that can be serialized by the format.
+        """
+        if self.column_headers:
+            indices = [((), ())] + indices
+        rows = sorted({row for row, column in indices})
```
rahulporuri, line 40: not sure why we're handling `self.column_headers` and `self.row_headers` in different ways - if `self.column_headers` is set, we're adding `[((), ())]` to the list of indices - indicating the row and column header - and then we add `[()]` to `columns` if `self.row_headers` is set.

It might be clearer if we just added `[()]` to `columns`/`rows` if the flags are set instead of doing it in two different ways.

pyface/ui/qt4/data_view/data_view_item_model.py
```diff
@@ -12,14 +12,15 @@
...
-from pyface.qt.QtCore import QAbstractItemModel, QModelIndex, Qt
+from pyface.qt.QtCore import QAbstractItemModel, QMimeData, QModelIndex, Qt
 from pyface.qt.QtGui import QColor
 from pyface.data_view.abstract_data_model import AbstractDataModel
 from pyface.data_view.abstract_value_type import CheckState
 from pyface.data_view.data_view_errors import (
     DataViewGetError, DataViewSetError
 )
 from pyface.data_view.index_manager import Root
+from .data_wrapper import DataWrapper
```
rahulporuri, line 23: nit : same comment as earlier about style variations in absolute vs relative imports.

pyface/data_view/abstract_data_exporter.py
```diff
@@ -0,0 +1,78 @@
...
+        model : AbstractDataModel instance
+            The data model holding the data.
+        indices : list of (row, column) index pairs
+            The indices where the data is to be stored.
+        """
+        try:
+            data = self.get_data(model, indices)
+            data_wrapper.set_format(self.format, data)
+        except DataViewGetError:
+            pass
```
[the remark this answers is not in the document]
corranwebster: The convention from discussion on previous PRs is that `DataViewGetError` and `DataViewSetError` are used to flag _expected_ errors which can safely be ignored.  I'm not sure we've documented that anywhere.  I'll open an issue.

pyface/data_view/data_formats.py
```diff
@@ -0,0 +1,293 @@
...
+    default : callable or None
+        Callable that takes a Python object and returns a JSON-serializable
+        data structure.
+
+    Returns
+    -------
+    raw_data : bytes
+        The serialized data as a bytestring.
+    """
+    str_data = json.dumps(data, default=default, separators=(',', ':'))
```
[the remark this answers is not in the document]
corranwebster: Maybe.  Separators here is just to produce the compact form, rather than a human readable one.

Encoding is more difficult, as it may be possible to specify encodings as part of the mimetype but I'm not sure what the toolkits do with that, so for now I'm assuming UTF-8 for everything as a sensible default, and if we run into instances where it is a problem we can revisit.

pyface/data_view/data_formats.py
```diff
@@ -0,0 +1,293 @@
...
+
+    Returns
+    -------
+    raw_data : bytes
+        The serialized data as a bytestring.
+    """
+    with StringIO() as fp:
+        writer = csv.writer(fp, delimiter=delimiter, **kwargs)
+        writer.writerow(data)
+        return fp.getvalue().encode(encoding)
```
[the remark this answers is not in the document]
corranwebster: As with any context manager when you leave the block, it will be in the variable `fp`.  It will have been `closed`, however.

Not sure why I used a context manager here - it may have simply been because of code evolution from an earlier iteration of the API where I was passing in a file-like.

pyface/data_view/data_formats.py
corranwebster: Numpy is already a required dependency for Wx, and is an optional dependency in a number of other places (not least the `ArrayDataModel`).

corranwebster: Also, I'm OK with a loud failure if code tries to use numpy-based serialization but numpy isn't in the environment.  That's a fairly clear programmer error.

pyface/data_view/exporters/row_exporter.py
corranwebster: Yes, that's clearer.

pyface/ui/qt4/data_view/data_view_item_model.py
corranwebster: Leaving this one here because the convention seems to be that within `pyface.ui.*` you do local imports.

I will open an issue more generally.

pyface/data_view/data_formats.py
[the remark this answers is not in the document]
corranwebster: I think I'm leaving this as-is for now.  If this doesn't fit someone's use-case it's supposed to be easy to define your own data formats, and I expect that this module may grow additional formats and arguments to existing formats as time goes on.

pyface/data_view/data_formats.py
```diff
@@ -0,0 +1,293 @@
...
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+import csv
+from functools import partial
+from io import BytesIO, StringIO
+import os
+import json
+
+from .i_data_wrapper import DataFormat, text_format
```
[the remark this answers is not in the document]
corranwebster: I was about to open an issue here, but the current code style guide in EEP 1 says:

> For imports within a sub-package, explicit relative imports are encouraged. For imports outside of a package, use fully-qualified absolute imports. One should not use explicit relative imports to go up a level and over to another package.

I'll ping folks to see if we want to re-think this.

corranwebster:
Assuming PR passes, I think everything is addressed one way or another.

rahulporuri (approved):
LGTM

PR #706, opened 2020-09-30
DataView drop handler

corranwebster:
This is basic drop support using the existing `IDropHandler` interface.  This is designed mainly for supporting entire-table drop actions (eg. dropping a file with alternative data onto a table, or dropping an object that will be inserted into a location automatically).

With some toolkit-specific code could be used for deeper introspection of the drop target and handle dropping to replace, insert, or even move values.  However, at some point we would like to expose an additional API that makes these sorts of operations easier.

This PR includes adding support for dropping `.npy` format files onto the array example to replace the array being viewed with one loaded from the `.npy` file.

### DIFF
--- a/pyface/data_view/i_data_view_widget.py
+++ b/pyface/data_view/i_data_view_widget.py
@@ -17,6 +17,7 @@ from traits.api import (
 )
 
 from pyface.data_view.abstract_data_model import AbstractDataModel
+from pyface.i_drop_handler import IDropHandler
 from pyface.i_widget import IWidget
 
 
@@ -31,6 +32,12 @@ class IDataViewWidget(IWidget):
 
     #: Whether or not the column headers are visible.
     header_visible = Bool(True)
+
+    #: The global drop handlers for the data view.  These are intended to
+    #: handle drop actions which either affect the whole data view, or where
+    #: the data handler can work out how to change the underlying data without
+    #: additional input.
+    drop_handlers = List(Instance(IDropHandler))
 
     #: What can be selected.  Implementations may optionally allow "column"
     #: and "item" selection types.
@@ -55,6 +62,12 @@ class MDataViewWidget(HasStrictTraits):
 
     #: Whether or not the column headers are visible.
     header_visible = Bool(True)
+
+    #: The global drop handlers for the data view.  These are intended to
+    #: handle drop actions which either affect the whole data view, or where
+    #: the data handler can work out how to change the underlying data without
+    #: additional input.
+    drop_handlers = List(Instance(IDropHandler))
 
     #: The selected indices in the view.  This should never be mutated, any
     #: changes should be by replacement of the entire list.
--- a/pyface/data_view/i_data_view_widget.py
+++ b/pyface/data_view/i_data_view_widget.py
@@ -37,7 +37,7 @@ class IDataViewWidget(IWidget):
     #: handle drop actions which either affect the whole data view, or where
     #: the data handler can work out how to change the underlying data without
     #: additional input.
-    drop_handlers = List(Instance(IDropHandler))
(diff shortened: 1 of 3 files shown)


pyface/data_view/i_data_view_widget.py
```diff
@@ -32,6 +33,12 @@ class IDataViewWidget(IWidget):
     #: Whether or not the column headers are visible.
     header_visible = Bool(True)
 
+    #: The global drop handlers for the data view.  These are intended to
```
rahulporuri, line 36: I'm not sure what you mean by "global" here and what the difference between a global vs local drop handler would be. Is it "global" because it can't handle local drops i.e. drops in a specific row/column/item?

pyface/ui/qt4/data_view/data_view_widget.py
```diff
@@ -45,10 +45,56 @@
...
+
+    _widget = None
+
+    def dragEnterEvent(self, event):
+        if self._widget is not None:
+            widget = self._widget
+            for drop_handler in widget.drop_handlers:
+                if drop_handler.can_handle_drop(event, self):
+                    event.acceptProposedAction()
+                    return
```
rahulporuri, line 59: Can we separate this out into a separate private method on the class? This code is repeated in all 4 methods and if I understand correctly, what we are doing here is try handling the event and pass it to the super class if we cannot.

pyface/ui/qt4/data_view/data_view_widget.py
```diff
@@ -45,10 +45,56 @@
...
+        super().dropEvent(event)
+
+
 @provides(IDataViewWidget)
 class DataViewWidget(MDataViewWidget, Widget):
     """ The Qt implementation of the DataViewWidget. """
 
+    #: Factory for the underlying Qt control, to facilitate replacement
```
rahulporuri, line 95: I'm not sure how this is facilitating replacement. Isn't this facilitating handling of the drop events?

rahulporuri (approved):
LGTM with a few questions. I tested the example locally and it works as expected.

pyface/data_view/i_data_view_widget.py
corranwebster: Yes.

If needed it _can_ handle row/column/item-specific drops, but you need to start writing toolkit-specific code to dig that information out of the toolkit event and widget objects.

pyface/ui/qt4/data_view/data_view_widget.py
corranwebster: One of the lessons learned from the TabularEditor, etc. was that you sometimes want to replace the Qt `QTreeView` control with something different but still adhering to the `QAbstractItemView` API; without a hook for a factory for this control, we need to subclass everything and re-implement `_create_control`; by adding a hook for a factory you just need to set a non-default value for the factory.

We were probably going to need this anyway; the subclassing of the `QTreeView` just meant that this was a good time to add the functionality.

pyface/ui/qt4/data_view/data_view_widget.py
corranwebster: Done

pyface/data_view/i_data_view_widget.py
```diff
@@ -37,7 +37,7 @@ class IDataViewWidget(IWidget):
     #: handle drop actions which either affect the whole data view, or where
     #: the data handler can work out how to change the underlying data without
     #: additional input.
-    drop_handlers = List(Instance(IDropHandler))
+    drop_handlers = List(Instance(IDropHandler, allow_none=False))
```
rahulporuri, line 40: IIUC, this ensures that it's an empty list of a list of handlers - not a list of `None`s

rahulporuri:
still LGTM

PR #709, opened 2020-09-30
DataView Color clean-up

### LINKED ISSUE #708: Fix DataView ColorValue expected test failures
There were some expected test failures in the `ColorValue` class for tests that were expecting `from_str` to be available; they are still failing but for unrelated reasons (calling mis-named methods).

corranwebster:
This PR does two things:

- switches from `Instance(Color)` to `PyfaceColor` traits
- fixes #708 (`ColorValue` was using an older iteration of the `ValueType` API)

We will probably revisit `ColorValue` again once there are custom editors available.

### DIFF
--- a/examples/data_view/column_example.py
+++ b/examples/data_view/column_example.py
@@ -10,14 +10,12 @@
 
 """ Example showing DataView for ColumnDataModel using row info. """
 
-from functools import partial
 import logging
-from random import choice, randint, uniform
 
 from traits.api import Bool, Dict, HasStrictTraits, Instance, Int, Str, List
 
 from pyface.api import ApplicationWindow, GUI, Image, ImageResource
-from pyface.color import Color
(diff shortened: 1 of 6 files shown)


rahulporuri (approved):

kitchoi:
Please could we consider the pattern of splitting Str and CStr so that one trait type does strictly validation and the other does casting. I believe the casting capability in the color trait is motivated by migrating legacy code in enable amd chaco. The fact that one can assign primitive types (e.g. str) to an attribute to get an instance back makes tracking such usage in enable/chaco for refactoring difficult. (It is also surprising to read.) By having a separate trait type that doesn't do casting, we can prevent this anti-pattern cropping up in new code. Note that I am only talking about the 'validate' method on the trait type, not the instantiation. Accepting string in the trait type instantiation is fine.

PR #713, opened 2020-10-05
Fix display of flags in examples

### LINKED ISSUE #712: DataView Row and Column examples not displaying flags
The problem is with this line: https://github.com/enthought/pyface/blob/ebe828cec61cb8c93ca718df05c232174098143b/examples/data_view/example_data.py#L93

It should be return ing "UK" and "USA" and not "Canada"

corranwebster:
Decided to add a Canadian flag rather than removing Canada; added flags to row example.

Fixes #712.


rahulporuri (approved):
LGTM assuming CI is happy.

rahulporuri:
Ran the examples locally and they work as expected (the flags show up)

PR #715, opened 2020-10-06
Fix QFontMetrics.width deprecation warning

### LINKED ISSUE #653: QFontMetrics.width deprecation warnings with PySide2 5.15
Related to https://github.com/enthought/traitsui/issues/1097

Running the test suite of TraitsUI with PySide2 5.15 reveals warnings emitted from pyface code:
```
/Users/kchoi/Work/ETS/venvs/traitsui-pyside2/lib/python3.8/site-packages/pyface/ui/qt4/code_editor/gutters.py:100: DeprecationWarning: QFontMetrics.width(const QString & arg__1, int len) const is deprecated
  self.fontMetrics().width(u"0" * ndigits) + 3, self.min_width
/Users/kchoi/Work/ETS/venvs/traitsui-pyside2/lib/python3.8/site-packages/pyface/ui/qt4/code_editor/find_widget.py:22: DeprecationWarning: QFontMetrics.width(const QString & arg__1, int len) const is deprecated
  self.button_size = self.fontMetrics().width(u"Replace All") + 20
/Users/kchoi/Work/ETS/venvs/traitsui-pyside2/lib/python3.8/site-packages/pyface/ui/qt4/code_editor/replace_widget.py:24: DeprecationWarning: QFontMetrics.width(const QString & arg__1, int len) const is deprecated
  self.button_size = self.fontMetrics().width(u"Replace All") + 20
/Users/kchoi/Work/ETS/venvs/traitsui-pyside2/lib/python3.8/site-packages/pyface/ui/qt4/code_editor/code_widget.py:471: DeprecationWarning: QFontMetrics.width(const QString & arg__1, int len) const is deprecated
  width = font_metrics.width(" ") * 80
```

The documentation [here](https://doc.qt.io/qtforpython/PySide2/QtGui/QFontMetrics.html#PySide2.QtGui.PySide2.QtGui.QFontMetrics.width) recommends using `horizontalAdvance` instead. However it is only available since Qt 5.11

aaronayres35:
fixes #653 

This PR simply uses QFontMetrics().horizontalAdvance() instead of QFontMetrics().width()when possible (horizontalAdvance was introduced in Qt 5.11) to avoid a deprecation warning.

Same fix was done in an open PR on traitsui: https://github.com/enthought/traitsui/pull/1315

### DIFF
--- a/pyface/ui/qt4/code_editor/code_widget.py
+++ b/pyface/ui/qt4/code_editor/code_widget.py
@@ -473,7 +473,12 @@ class CodeWidget(QtGui.QPlainTextEdit):
         style = self.style()
         opt = QtGui.QStyleOptionHeader()
         font_metrics = QtGui.QFontMetrics(self.document().defaultFont())
-        width = font_metrics.width(" ") * 80
+        # QFontMetrics.width() is deprecated and Qt docs suggest using
+        # horizontalAdvance() instead, but is only available since Qt 5.11
+        if QtCore.__version_info__ >= (5, 11):
+            width = font_metrics.horizontalAdvance(" ") * 80
+        else:
+            width = font_metrics.width(" ") * 80
         width += self.line_number_widget.sizeHint().width()
         width += self.status_widget.sizeHint().width()
         width += style.pixelMetric(QtGui.QStyle.PM_ScrollBarExtent, opt, self)
--- a/pyface/ui/qt4/code_editor/find_widget.py
+++ b/pyface/ui/qt4/code_editor/find_widget.py
@@ -19,7 +19,12 @@ class FindWidget(QtGui.QWidget):
         super(FindWidget, self).__init__(parent)
         self.adv_code_widget = weakref.ref(parent)
 
-        self.button_size = self.fontMetrics().width("Replace All") + 20
+        # QFontMetrics.width() is deprecated and Qt docs suggest using
+        # horizontalAdvance() instead, but is only available since Qt 5.11
+        if QtCore.__version_info__ >= (5, 11):
+            self.button_size = self.fontMetrics().horizontalAdvance("Replace All") + 20
+        else:
+            self.button_size = self.fontMetrics().width("Replace All") + 20
 
         form_layout = QtGui.QFormLayout()
         form_layout.addRow("Fin&d", self._create_find_control())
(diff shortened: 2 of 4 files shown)


kitchoi:
All the changes are pretty similar and they solve the same compatibility problem. Could we refactor the handling `font_metrics.width(" ")` into a single function?

I am wondering if it is worth having a module for compatibility among various versions of Qt, so that when we drop a specific version of Qt, we can maintain the compatibility layer more easily.  @corranwebster, what do you think?

(I would have wanted to designate such a module to be private to pyface to start with, but I don't think there is a notion of private modules in pyface so far 🤷 .)

corranwebster:
@kit we already do some amount of fix-ups to make the APIs match (mainly imports between Qt4 and Qt5) in the `pyface.qt` code, so we could, potentially, do something like monkeypatching old versions of `QFontMetrics` to add `horizontalAdvance` as a synonym for `width`.

kitchoi:
Hmmm the `pyface.qt` package only normalizes the API namespace differences among different Python/Qt wrapper without changing behaviour. When I import from `pyface.qt`, I consider it being a collection of aliases and I trust what I use from there to be consistent with what I get if I had imported the objects from PySide2 and PyQt5 directly. Modifying behaviour and monkeypatching will absolutely kill that assumption and understanding for that package.

kitchoi:
I was mainly thinking about defining and reusing a function like this which is simple enough and we can find their usage again easily:
```
def font_horizontal_advance(qfont_metrics, text):
    if QtCore.__version_info__ >= (5, 11):  # as in this PR
        return qfont_metrics.horizontal_advance(text)
    return qfont_metrics.width(text)
```
If adding this function is okay, the question is, where should this live?
Ideas: `pyface.ui.qt4.compat`?

kitchoi:
On the other hand, these repetitions are still minor so I'd also be happy to just merge this as it is.

corranwebster:
I think it's fine to merge as-is myself.  If we were to generalize/unify as a separate function, I think it would make sense to do so to Wx as well (similar to what we do with `ISystemMetrics` for screen size).

corranwebster (approved):

corranwebster:
I have opened #716 to remind us of the discussion.

kitchoi:
SGTM.

rahulporuri (commented):
Can we create an issue and link it in the changes being made in this PR? I don't want future us arriving at these pieces of code after a random walk through the pyface codebase.

PR #719, opened 2020-10-07
Fix deprecation warning from Qt SystemMetrics width and height computation

### LINKED ISSUE #718: DeprecationWarning from Qt SystemMetrics width and height computation
Running the test suite with PySide2 5.15 produces the following deprecation warnings:
```
$ python -Wdefault -m unittest pyface.tests.test_system_metrics
./Users/kchoi/Work/ETS/pyface/pyface/ui/qt4/system_metrics.py:47: DeprecationWarning: QDesktopWidget.screenGeometry(int screen) const is deprecated
  QtGui.QApplication.instance().desktop().screenGeometry().height()
./Users/kchoi/Work/ETS/pyface/pyface/ui/qt4/system_metrics.py:43: DeprecationWarning: QDesktopWidget.screenGeometry(int screen) const is deprecated
  return QtGui.QApplication.instance().desktop().screenGeometry().width()
```

aaronayres35:
fixes #718 

This PR replaces `QtGui.QApplication.instance().desktop(`) with `QtGui.QApplication.instance().screens()[0]` when possible (when qt version is recent enough to have `screens()`) so that screen geometry can be obtained on a `QScreen` object (where it is not a deprecated method) rather than on a `QDesktopWidget` object.

### DIFF
--- a/pyface/ui/qt4/system_metrics.py
+++ b/pyface/ui/qt4/system_metrics.py
@@ -12,7 +12,7 @@
 # However, when used with the GPL version of PyQt the additional terms described in the PyQt GPL exception also apply
 
 
-from pyface.qt import QtGui
+from pyface.qt import QtCore, QtGui
 
 
 from traits.api import HasTraits, Int, Property, provides, Tuple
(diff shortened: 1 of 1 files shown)


pyface/ui/qt4/system_metrics.py
```diff
@@ -40,12 +40,24 @@ class SystemMetrics(MSystemMetrics, HasTraits):
     # ------------------------------------------------------------------------
 
     def _get_screen_width(self):
-        return QtGui.QApplication.instance().desktop().screenGeometry().width()
+        # QDesktopWidget.screenGeometry() is deprecated and Qt docs
+        # suggest using screens() instead, but screens in not available in qt4
+        if QtCore.__version_info__ >= (5, 0):
```
corranwebster, line 45: You might use `is_qt5` constant from `pyface.qt` here: https://github.com/enthought/pyface/blob/daa26ff0afbb36831be2febab6ab5fa591a2e47a/pyface/qt/__init__.py#L95

pyface/ui/qt4/system_metrics.py
```diff
@@ -12,7 +12,7 @@
 # However, when used with the GPL version of PyQt the additional terms described in the PyQt GPL exception also apply
 
 
-from pyface.qt import QtGui
+from pyface.qt import QtCore, QtGui
```
corranwebster, line 15:
```suggestion
from pyface.qt import QtCore, QtGui, is_qt5
```

pyface/ui/qt4/system_metrics.py
```diff
@@ -40,12 +40,24 @@ class SystemMetrics(MSystemMetrics, HasTraits):
...
+        else:
+            return QtGui.QApplication.instance().desktop().screenGeometry().width()
 
     def _get_screen_height(self):
-        return (
-            QtGui.QApplication.instance().desktop().screenGeometry().height()
-        )
+        # QDesktopWidget.screenGeometry(int screen) is deprecated and Qt docs
+        # suggest using screens() instead, but screens in not available in qt4
+        if QtCore.__version_info__ >= (5, 0):
```
corranwebster, line 53: Same comment here.

kitchoi (approved):
Just a suggestion: If you are going to open an issue like the ones in TraitsUI for cleaning up these patches when Qt4 is dropped, you might want to consider including the issue number in the comment before merging so that when someone does remove the patch, they can find the issue more easily and close it. Without the back reference, if one tries to find these patches by searching for `is_qt5` in the code base (instead of finding the issues that mentions Qt4), they may not know they can then close the issue.

(Also I am not suggesting a new issue should be open for this, neither am I suggesting we should not open one. I think we all have our own preferences, whatever makes sense ;) )

PR #727, opened 2020-10-09
Use availableGeometry when computing screen size to only get available space on scree

### LINKED ISSUE #724: Change Qt SystemMetrics width and height computation to use availableGeometry instead of screenGeometry
Recently in #719 the Qt ScreenMetrics code was modified to address a deprecation warning, and this was motivated by an desire to use this code in traitsui rather than having to recreate it in traitsui.  However, the SystemMetrics screen width and height computation on Qt currently uses `screenGeometry` and `geometry` rather than `availableGeometry`.  In reality we would typically want the available space on the screen rather than the size of the full screen itself.  See https://github.com/enthought/traitsui/pull/1322#issuecomment-706207595.

aaronayres35:
fixes #724 

For discussion of the motivation see https://github.com/enthought/traitsui/pull/1322

This PR simply replaces calls to `screeGeometry` and `geometry` with calls to `availableGeometry` as we are really interested in the available screen space not the full screen space.

### DIFF
--- a/pyface/ui/qt4/system_metrics.py
+++ b/pyface/ui/qt4/system_metrics.py
@@ -54,11 +54,11 @@ class SystemMetrics(MSystemMetrics, HasTraits):
         # see issue: enthought/pyface#721
         if is_qt5:
             return (
-                QtGui.QApplication.instance().screens()[0].geometry().height()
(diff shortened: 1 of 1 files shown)


corranwebster (approved):
This looks good.

PR #728, opened 2020-10-10
Update Trait Types

corranwebster:
This PR replaces `Either` with `Union` and some uses of `Trait` with `Map`.

This breaks compatibility with Traits 6.0, and should not be merged until after Pyface 7.1 is released.


corranwebster:
Failure is an expected failure when running against Traits 6.0.

pyface/timer/i_timer.py
```diff
@@ -30,6 +30,7 @@
     Property,
     Range,
     Tuple,
+    Union,
```
rahulporuri, line 33:
```suggestion
```

rahulporuri (approved):
Almost LGTM

PR #731, opened 2020-10-10
Have mixins inherit from HasTraits

### LINKED ISSUE #730: Mixins should inherit HasTraits rather than object
It is convenient to define common traits on the mixin classes for widgets (ie. `MWidget`, `MDialog`, etc.)  However many of these classes inherit from `object` rather than `HasTraits`.

We should make all mixin classes inherit from `HasTraits` so traits can be defined on them.  Changing this should not affect the behaviour of any existing `Widget` class that simply defined methods, as the methods that they define will not conflict with any methods on `HasTraits`, since the concrete classes all inherit from `HasTraits`.

corranwebster:
This is a simple search and replace of `class MFoo(object):` with `class MFoo(HasTraits):`, combined with fixing of the imports where necessary.

Fixes #730.

This not intended to be part of the 7.1.0 release.

### DIFF
--- a/pyface/i_about_dialog.py
+++ b/pyface/i_about_dialog.py
@@ -11,7 +11,7 @@
 """ The interface for a simple 'About' dialog. """
 
 
-from traits.api import Instance, List, Str
+from traits.api import HasTraits, Instance, List, Str
 
 
 from pyface.i_dialog import IDialog
@@ -33,7 +33,7 @@ class IAboutDialog(IDialog):
     image = Instance(ImageResource, ImageResource("about"))
 
 
-class MAboutDialog(object):
+class MAboutDialog(HasTraits):
     """ The mixin class that contains common code for toolkit specific
     implementations of the IAboutDialog interface.
     """
--- a/pyface/i_application_window.py
+++ b/pyface/i_application_window.py
@@ -11,7 +11,7 @@
 """ The interface of a top-level application window. """
 
 
-from traits.api import Instance, List
+from traits.api import HasTraits, Instance, List
 
 
 from pyface.action.api import MenuBarManager, StatusBarManager, ToolBarManager
@@ -109,7 +109,7 @@ class IApplicationWindow(IWindow):
         """ Sets the window icon (if required). """
 
 
-class MApplicationWindow(object):
+class MApplicationWindow(HasTraits):
     """ The mixin class that contains common code for toolkit specific
     implementations of the :py:class:`IApplicationWindow` interface.
 
--- a/pyface/i_confirmation_dialog.py
+++ b/pyface/i_confirmation_dialog.py
@@ -11,7 +11,7 @@
 """ The interface for a dialog that prompts the user for confirmation. """
 
 
-from traits.api import Bool, Enum, Instance, Str
+from traits.api import Bool, Enum, HasTraits, Instance, Str
 
 
 from pyface.constant import CANCEL, NO, YES
@@ -50,7 +50,7 @@ class IConfirmationDialog(IDialog):
     yes_label = Str()
 
 
-class MConfirmationDialog(object):
+class MConfirmationDialog(HasTraits):
     """ The mixin class that contains common code for toolkit specific
     implementations of the IConfirmationDialog interface.
     """
--- a/pyface/i_dialog.py
+++ b/pyface/i_dialog.py
@@ -11,7 +11,7 @@
 """ The abstract interface for all pyface dialogs. """
 
 
-from traits.api import Bool, Enum, Int, Str
+from traits.api import Bool, Enum, HasTraits, Int, Str
 
 
 from pyface.constant import OK
@@ -134,7 +134,7 @@ class IDialog(IWindow):
         """
 
 
-class MDialog(object):
+class MDialog(HasTraits):
     """ The mixin class that contains common code for toolkit specific
     implementations of the IDialog interface.
 
--- a/pyface/i_directory_dialog.py
+++ b/pyface/i_directory_dialog.py
@@ -12,7 +12,7 @@
 """
 
 
-from traits.api import Bool, Str
+from traits.api import Bool, HasTraits, Str
 
 
 from pyface.i_dialog import IDialog
@@ -42,7 +42,7 @@ class IDirectoryDialog(IDialog):
     path = Str()
 
 
-class MDirectoryDialog(object):
+class MDirectoryDialog(HasTraits):
     """ The mixin class that contains common code for toolkit specific
     implementations of the IDirectoryDialog interface.
     """
--- a/pyface/i_file_dialog.py
+++ b/pyface/i_file_dialog.py
@@ -14,7 +14,7 @@
 import sys
 
 
-from traits.api import Enum, Str, Int
+from traits.api import Enum, HasTraits, Int, Str
 
 
 from pyface.i_dialog import IDialog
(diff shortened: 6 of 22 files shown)


kitchoi:
See https://github.com/enthought/pyface/issues/730#issuecomment-706969173 for a scenario where this PR may change behaviour.

corranwebster:
See the response on #730 - the example does not occur for any class.

This PR should, however, make the base `Widget` classes inherit from just `MWidget` rather than `MWidget` and `HasTraits`, for clarity and simplicity.

kitchoi:
Another question: Could we use `HasStrictTraits` instead of `HasTraits` please? Honestly the `HasStrictTraits` is safer and more useful.

kitchoi:
To be honest, this switch from not-subclassing a base class with a metaclass to subclassing one makes me nervous. It would be good to have a use case that can motivate this change.

corranwebster:
I would love to use `HasStrictTraits`, but that is a change in behaviour that could break user code.  If we were to do that change it should be in a major release.

The motivating use case is being able to write traits code in the mixins and be confident that they will work as expected.

corranwebster:
Just to be clear, these are the curren MROs for the modified classes:
```
AboutDialog : AboutDialog, MAboutDialog, Dialog, MDialog, Window, MWindow, Widget, MWidget, HasTraits, CHasTraits, object
ApplicationWindow : ApplicationWindow, MApplicationWindow, Window, MWindow, Widget, MWidget, HasTraits, CHasTraits, object
ConfirmationDialog : ConfirmationDialog, MConfirmationDialog, Dialog, MDialog, Window, MWindow, Widget, MWidget, HasTraits, CHasTraits, object
Dialog : Dialog, MDialog, Window, MWindow, Widget, MWidget, HasTraits, CHasTraits, object
DirectoryDialog : DirectoryDialog, MDirectoryDialog, Dialog, MDialog, Window, MWindow, Widget, MWidget, HasTraits, CHasTraits, object
FileDialog : FileDialog, MFileDialog, Dialog, MDialog, Window, MWindow, Widget, MWidget, HasTraits, CHasTraits, object
GUI : GUI, MGUI, HasTraits, CHasTraits, object
HeadingText : HeadingText, MHeadingText, Widget, MWidget, HasTraits, CHasTraits, object
ImageCache : ImageCache, MImageCache, HasTraits, CHasTraits, object
ImageResource : ImageResource, MImageResource, HasTraits, CHasTraits, object
MessageDialog : MessageDialog, MMessageDialog, Dialog, MDialog, Window, MWindow, Widget, MWidget, HasTraits, CHasTraits, object
ProgressDialog : ProgressDialog, MProgressDialog, Window, MWindow, Widget, MWidget, HasTraits, CHasTraits, object
PythonEditor : PythonEditor, MPythonEditor, Widget, MWidget, HasTraits, CHasTraits, object
PythonShell : PythonShell, MPythonShell, Widget, MWidget, HasTraits, CHasTraits, object
SingleChoiceDialog : SingleChoiceDialog, MSingleChoiceDialog, Dialog, MDialog, Window, MWindow, Widget, MWidget, HasTraits, CHasTraits, object
SplashScreen : SplashScreen, MSplashScreen, Window, MWindow, Widget, MWidget, HasTraits, CHasTraits, object
SplitDialog : SplitDialog, Dialog, MDialog, Window, MWindow, Widget, MWidget, SplitWidget, MSplitWidget, HasTraits, CHasTraits, object
SplitPanel : SplitPanel, Widget, MWidget, SplitWidget, MSplitWidget, HasTraits, CHasTraits, object
SystemMetrics : SystemMetrics, MSystemMetrics, HasTraits, CHasTraits, object
Widget : Widget, MWidget, HasTraits, CHasTraits, object
Window : Window, MWindow, Widget, MWidget, HasTraits, CHasTraits, object
Wizard : Wizard, MWizard, Dialog, MDialog, Window, MWindow, Widget, MWidget, HasTraits, CHasTraits, object
WizardPage : WizardPage, MWizardPage, HasTraits, CHasTraits, object
```
None of the MROs are changed by this PR (because they all end with `HasTraits, CHasTraits, object`).

kitchoi:
So it only backward compatible to go from strict to non-strict, but not the other way round.

As you replied in https://github.com/enthought/pyface/issues/730#issuecomment-706993502:

> It is not a problem because everything affected is, for better or worse, using HasTraits, except the MDataViewWidget which is already using HasStrictTraits, which is OK for this change. In particular, all other widgets get the HasTraits by inheritance from Widget with no other.

Did I understand correctly that this change will silently remove the strictness on `MDataViewWidget`?
What about other subclasses outside of pyface that have subclassed `HasStrictTraits` one way or another?

corranwebster:
No, you misunderstood, this change will not affect the strictness of `MDataViewWidget`.  MRO before:
```
DataViewWidget : DataViewWidget, MDataViewWidget, HasStrictTraits, Widget, MWidget, HasTraits, CHasTraits, object
```
MRO after:
```
DataViewWidget : DataViewWidget, MDataViewWidget, HasStrictTraits, Widget, MWidget, HasTraits, CHasTraits, object
```
(ie. unchanged)

I honestly can't see any sane usage where this change would change behaviour.  If someone tried to mixin `HasStrictTraits` in a subclass, they either have to do something like (trying to make a reasonable example):
```
class MyWindow(HasStrictTraits, ApplicationWindow):
```
or
```
class MyWindow(ApplicationWindow, HasStrictTraits):
```
The first has MRO:
```
MyWindow, HasStrictTraits, ApplicationWindow, MApplicationWindow, Window, MWindow, Widget, MWidget, HasTraits, CHasTraits, object
```
which is unchanged by this PR; the second has MRO:
```
MyWindow, ApplicationWindow, MApplicationWindow, Window, MWindow, Widget, MWidget, HasStrictTraits, HasTraits, CHasTraits, object
```
which is also unchanged by this PR.

This change is safe from those concerns as best I can tell.

kitchoi:
I'd be happy for existing classes that are nonstrict to stay that way for backward compatibility sake... it is not something worth breaking code for. But I'd hope new classes can use `HasStrictTraits` where possible and I'd hope that existing classes that are already strict do not lose that strictness from this change.

The `DataViewWidget` is not affected because it's defined as `DataViewWidget(MDataViewWidget, Widget)`. If it was defined as `DataViewWidget(Widget, MDataViewWidget)` then it would have been affected.
e.g. right now the following raises:
```
DataViewWidget().i_am_undefined = 1
# traits.trait_errors.TraitError: Cannot set the undefined 'i_am_undefined' attribute of a 'DataViewWidget' object.
```
This branch does not change that either. But if `DataViewWidget` was defined with parent classes like this: `DataViewWidget(Widget, MDataViewWidget)`, then the above assignment would not raise.

Some other classes have more complicated inheritance tree so it will require quite a bit of care if we don't want to remove strictness that has been in p lace.

kitchoi:
> But if DataViewWidget was defined with parent classes like this: DataViewWidget(Widget, MDataViewWidget), then the above assignment would not raise.

Argh, if `Widget` and `MDataViewWidget` were swapped like that then `DataViewWidget` would not be strict on master either. This is too confusing for me 😢 .  Perhaps the easiest way to ensure this does not bring about unwanted change is to add tests for all the classes here.

corranwebster:
I've written up all the gory details of what's going on in #730, but the summary is that this won't change the `HasStrictTraits` behaviour of any external classes if they inherit from the concrete widget classes: either they worked before or they didn't work before, and this won't change it one way or another.

corranwebster:
From all the discussion, this is not going to make any immediate changes if merged.

There is a possibility of change of strictness because of https://github.com/enthought/traits/issues/1305 but no known class is affected by this, and any change of behaviour because of that is best resolved by fixing https://github.com/enthought/traits/issues/1305

So I think this is OK to merge to master.

kitchoi (approved):

corranwebster:
Closing and re-opening to re-trigger checks.

Issue #729, opened 2020-10-10
The _create method should be public


corranwebster:
The base `Widget._create` method is nominally private, but in practice tends to get used as:
```
w = Widget(parent=parent)
w._create()
```
although a few widgets perform the `_create()` call as part of their `__init__`.

Two-phase creation has an advantage in that it permits programmatic configuration of the widget beyond what is passed in at class creation time before the underlying toolkit control is created.

This issue proposes that:

i) there is a public `create` method, which initially can be implemented to simply call `_create`
ii) any classes which call `_create` or (worse) `_create_control` in their `__init__` be modified to not do that

corranwebster:
Strategy for removing `create` calls from `__init__`:

- [ ] add a keyword-only argument `create` which defaults to `True` initially (since `create` is a method name it will never conflict with a trait name)
- [ ] if `create` is `True` then call the `create` method, but also issue a `PendingDeprecationWarning`
- in the next a major release:
  - [ ]ignore the `create` argument other than indicate that it is deprecated if it is passed
- [ ] at some later release, remove any handling for `create` (ideally, removing the entire `__init__` method).

corranwebster:
I've added this to the 8.0 release - the next few steps in this should be done for that release, if time permits.

corranwebster:
Incremental progress in the 8.0 release, the rest pushed to 9.0

PR #735, opened 2020-10-10
Add a create method for IWidget

corranwebster:
This adds `create` as a synonym for `_create` and uses it for `Widget` class tests.

This is the first part of #729.


rahulporuri (approved):
LGTM

PR #739, opened 2020-10-11
Add a section to the documentation on the basics of Pyface Widgets

corranwebster:
It would be nice to have this merged before we generate the 7.1 release docs.

This is a pure documentation update, so should not affect functionality in any way.


docs/source/widgets.rst
```diff
@@ -0,0 +1,137 @@
...
+
+The actual implementation is provided by the corresponding toolkit-dependent
+:class:`~pyface.widget.Widget` subclass, but with shared functionality split
+into separate mixins: :class:`~pyface.i_widget.MWidget` and so on, following
+the general design described in the :ref:`overview` section.
+
+Lifecycle
+=========
+
+When a :class:`~pyface.widget.Widget` instance is created only the traits
```
kitchoi, line 26: After running `python etstool.py docs --toolkit=pyqt5`, the classes are merely bolded but they are not links to anything.
Screenshot 2020-10-13 at 16 39 46

Just to confirm, I can import `pyface.widget.Widget` from the same environment I used for building the doc.

kitchoi (approved):
LGTM

docs/source/widgets.rst
corranwebster: I'm guessing that the autodoc tools are not seeing the classes being imported there because of toolkit magic.

Fine for now; but should open an issue because there may be other places thiis is happening.

Issue #714, opened 2020-10-05
Pyface minor version release 7.1.0


rahulporuri:
Target date: To be confirmed, will edit here once confirmed.
This is a minor release. See https://github.com/enthought/traitsui/issues/1307 for TraitsUI minor release 7.1.0.

Release candidates
--------------------
- [x] Verify that all relevant PRs have been merged to master
- [x] Bump version number on master for continued development (#746)
- [x] From the commit just before bumping the version, create a new branch `maint/7.1`
- [x] Update `ci-src-requirements.txt` to target TraitsUI `maint/7.1` branch (#754)
- [x] Make sure `maint/7.1` is included (somehow) in the following block in appveyor.yml (#748, but see #751 as well. This step may not be needed in future releases)
```
branches:
  only:
    - ...
```
- [x] Update changelog and open PR targeting a new `maint/7.1` branch (#720)
- [x] Check MANIFEST and requirements are still up to date.
- [x] Update version in setup.py for the prerelease (#771)
       - (FTR: This step was done for 7.1.0rc2 but not 7.1.0rc1 due to #744.)
       - Create a new branch from maint/7.1: `git checkout maint/7.1; git pull; git checkout -b call-it-anything-you-like`
       - Set PRERELEASE to "rc2" and IS_RELEASED to true, commit, e.g.`git commit -m "Set IS_RELEASED to true for prerelease 7.1.0rc2"`
       - Flip IS_RELEASED back to false, commit.
       - Open a PR against `maint/7.1`, merge but DO NOT SQUASH
                No squashing so that the tagged commit in the next step will be on the branch (Kit: It is not a big deal to me even if the tagged commit was not on the branch. But some prefer it that way.) Alternatively, flip the IS_RELEASED flag in a separate PR.
- [x] Tag (annotated!) the release candidate on the commit where IS_RELEASED is set to true, e.g. `git tag -a -m "Release candidate 7.1.0rc2" 7.1.0rc2 <commit-hash>`
        - (FTR: the tag for 7.1.0rc1 was different due to #744)
- [x] Push the tag to GitHub, e.g. `git push origin 7.1.0rc2`
- [x] Upload to PyPI
      - (FTR: This step was done for 7.1.0rc2 but not 7.1.0rc1 due to #744.)
      - `git checkout <tag>`, `git clean -ffxd`, `python setup.py sdist`, `twine check dist/<...>.tar.gz`, `twine upload dist/<...>.tar.gz`
[...]

PR #747, opened 2020-10-13
Update logic for writing _version.py to allow prereleases

### LINKED ISSUE #744: Update version computation logic for release candidates
It would be nice if the release candidate can say in its version that it is a release candidate, e.g. 
```
>>> from pyface import __version__
>>> __version__
"7.1.0rc1"
```
Currently any version apart from the release has a suffix of "dev{num}".

We may be able to bring the `setup.py` in pyface more inline with traits.

kitchoi:
Closes #744

This PR updates the machinery in setup.py so that we can set prerelease version. 
This will allow us to have a commit that set the version and have the commit to be tagged.

The changes here try to bring `setup.py` inline with the one in traits as much as possible.
_Except_, the logic to do with computing git revision has not changed.

Main changes
- The computation of version now includes a prerelease segment.
- `version` is now the same as `full_version`. In all scenarios we have only used `full_version` anyway. See this:
https://github.com/enthought/pyface/blob/dbe6f9aef89f40ad5b8185cac78a8eed5b05ee8d/pyface/__init__.py#L15-L18
and this:
https://github.com/enthought/pyface/blob/dbe6f9aef89f40ad5b8185cac78a8eed5b05ee8d/docs/source/conf.py#L40-L45
- `_version.py` is always updated in the development environment when `setup.py` is run. Previously it is only updated if the file was not found. 

These are the tests I wrote to help me make sure I have not broken things catastrophically. Not sure if I shall commit this and where I should commit this:

<details>
    <summary> Tests for setup.py </summary>

```
import contextlib
import os
import runpy
import shutil
import tempfile
import unittest
from unittest import mock

import packaging.version


class Namespace:
    def __init__(self, variables):
        self.variables = variables.copy()

    def __getattr__(self, name):
        return self.variables[name]


SETUP_PY = Namespace(runpy.run_path("setup.py"))


@contextlib.contextmanager
def change_dir(dir):
    original = os.getcwd()
    os.chdir(dir)
    try:
        yield
    finally:
        os.chdir(original)


class TestSetup(unittest.TestCase):

    def get_version_py_content(self):
        with tempfile.TemporaryDirectory() as tmp_dir:
            filename = os.path.join(tmp_dir, "_version.py")
            SETUP_PY.write_version_file(*SETUP_PY.git_version(), filename)
            return runpy.run_path(filename)

    def test_version_py(self):
        version_file_content = self.get_version_py_content()

        version_str = version_file_content["version"]
        parsed_version = packaging.version.Version(version_str)
        self.assertEqual(
            parsed_version.release,
            (SETUP_PY.MAJOR, SETUP_PY.MINOR, SETUP_PY.MICRO)
        )
        if SETUP_PY.IS_RELEASED:
            self.assertEqual(
                version_str,
                "{}.{}.{}".format(parsed_version.release)
            )

    def test_full_version(self):
        version_file_content = self.get_version_py_content()
        self.assertEqual(
            version_file_content["version"],
            version_file_content["full_version"]
        )

    def test_git_revision(self):
        version_file_content = self.get_version_py_content()
        git_revision = version_file_content["git_revision"]
        self.assertEqual(len(git_revision), 8)
        try:
            int(git_revision, 16)
        except ValueError:
            self.fail(f"Git version should be hex decimal, got {git_revision}")

    def test_is_released(self):
        version_file_content = self.get_version_py_content()
        is_released = version_file_content["is_released"]
        self.assertEqual(is_released, SETUP_PY.IS_RELEASED)

    def test_prerelease(self):
        version_file_content = self.get_version_py_content()
        version_str = version_file_content["version"]
        parsed_version = packaging.version.Version(version_str)
        if SETUP_PY.PRERELEASE:
            self.assertEqual(
                parsed_version.pre, SETUP_PY.PRERELEASE
            )
        else:
            self.assertIsNone(parsed_version.pre)

    def test_resolve_version(self):
        with mock.patch("sys.stdout"):
            version, _ = SETUP_PY.resolve_version()
        expected = "{}.{}.{}{}".format(
            SETUP_PY.MAJOR,
            SETUP_PY.MINOR,
            SETUP_PY.MICRO,
            SETUP_PY.PRERELEASE,
        )
        if not SETUP_PY.IS_RELEASED:
            self.assertTrue(version.startswith(expected + ".dev"))
        else:
            self.assertEqual(version, expected)

    def test_no_git_no_version_py(self):

        with tempfile.TemporaryDirectory() as tmp_dir:
            fake_setup_py = os.path.join(tmp_dir, "setup.py")
            shutil.copy("setup.py", fake_setup_py)

            # this file is not created.
            filename = os.path.join(tmp_dir, "_version.py")

            with change_dir(tmp_dir), \
                    mock.patch("sys.stdout"):
                setup_py = Namespace(
                    runpy.run_path(
                        "setup.py",
                        {
                            "__file__": fake_setup_py,
                        },
                    )
                )
                with self.assertRaises(RuntimeError) as e:
                    setup_py.resolve_version()
        self.assertIn(
            "Unable to determine package version",
            str(e.exception)
        )

    def test_no_git_but_has_version_py(self):

        with tempfile.TemporaryDirectory() as tmp_dir:
            fake_setup_py = os.path.join(tmp_dir, "setup.py")
            shutil.copy("setup.py", fake_setup_py)

            with change_dir(tmp_dir), \
                    mock.patch("sys.stdout"):
                setup_py = Namespace(
                    runpy.run_path(
                        "setup.py",
                        {
                            "__file__": fake_setup_py,
                        },
                    )
                )
                os.makedirs(os.path.dirname(setup_py.VERSION_FILE))
                setup_py.write_version_file("6.1.0.dev0", "cd34fd23")
                version, _ = setup_py.resolve_version()
                actual_content = runpy.run_path(setup_py.VERSION_FILE)
                self.assertEqual(version, actual_content["version"])
                self.assertEqual(version, "6.1.0.dev0")

```

</details>


docs/source/conf.py
```diff
@@ -39,10 +41,10 @@
...
 # The default replacements for |version| and |release|, also used in various
 # other places throughout the built documents.
-d = {}
-with open(os.path.join('..', '..', 'pyface', '_version.py')) as fp:
-    exec (fp.read(), d)
-version = release = d['full_version']
+version_py = os.path.join('..', '..', 'pyface', '_version.py')
+version_content = runpy.run_path(version_py)
+version = ".".join(version_content["version"].split(".", 2)[:2])
+release = version
```
kitchoi, line 47: This is somewhat a drive-by. The new logic here is consistent with the one in TraitsUI. However Pyface documentation has not used `|version|` anywhere.

setup.py
```diff
@@ -73,76 +141,157 @@ def _minimal_ext_cmd(cmd):
     else:
         git_revision, git_count = match.group("hash"), match.group("count")
 
-    return git_revision, git_count
+    return git_count, git_revision
```
kitchoi, line 144: Swapped to be consistent with Traits.

setup.py
```diff
@@ -10,21 +10,71 @@
...
+is_released = {is_released}
+"""
+
+# Git executable to use to get revision information.
+GIT = "git"
 
 
 def read_module(module, package="pyface"):
-    """ Read a simple .py file from pyface in a safe way.
+    """ Read a simple .py file from in a safe way.
```
rahulporuri, line 77:
```suggestion
    """ Read a simple .py file from package in a safe way.
```

setup.py
```diff
@@ -10,21 +10,71 @@
...
 
-VERSION = "%d.%d.%d" % (MAJOR, MINOR, MICRO)
+# If this file is part of a Git export (for example created with "git archive",
+# or downloaded from GitHub), ARCHIVE_COMMIT_HASH gives the full hash of the
+# commit that was exported.
+ARCHIVE_COMMIT_HASH = "$Format:%H$"
+
+# Templates for version strings.
+RELEASED_VERSION = "{major}.{minor}.{micro}{prerelease}"
+UNRELEASED_VERSION = "{major}.{minor}.{micro}{prerelease}.dev{dev}"
```
rahulporuri, line 36: just to confirm, here's what i think the process needs to be based on this PR

- `PRERELEASE = ''` set to empty string up until a release
- before making a prerelease, we flip `IS_RELEASED=True` and set `PRERELEASE='rc0'`
- when we are reading to make the release, we set `PRERELEASE=''` set back to empty string.

Is that correct?

rahulporuri (commented):
i'm not sure where the test file can go but i think it is definitely worth committing the setup.py test you put together.

setup.py
kitchoi: This is also my understanding. That said, this is also taken from Traits as I am trying to make things more consistent.
(For my own personal projects, I compute the version differently 😂  so there isn't a "IS_RELEASED" boolean.)

rahulporuri: the exact same question I had is what Corran also asked in https://github.com/enthought/traits/pull/846/files#r372846588 - to which Mark responsed in the thread.

It would be useful to document this somewhere though (either in the setup.py or in the checklist) instead of in this thread/in our brains.

kitchoi: My thought is that this will take another release to exercise the full release process including the release candidate, and while we do that, we can document the steps like we did in #714, with references to PRs making these steps. Eventually when the same steps have been demonstrated to work a second time, we can write them in a separate document if we like. But repositories are going to have slight variations (like TraitsUI having to update ci-src-requirements.txt while Pyface does not), so the issue from the previous release is always going to the one I would look for when I make a release.

kitchoi: See #714 where I have updated the steps as I went, documenting what happened for 7.1 and the steps that had not happened because of #744.

rahulporuri (approved):
LGTM.

I tested 4 variations locally - `PRERELEASE` set to `""` and `"rc1"` and `IS_RELEASED` set to `True` and `False`. Everything works as expected, although it looks awkward setting `PRERELEASE` when `IS_RELEASED` is `False`.

rahulporuri:
@kitchoi i'm guessing the same/mostly same changes will be moved over to traitsui `setup.py` as well?

kitchoi:
> i'm guessing the same/mostly same changes will be moved over to traitsui setup.py as well?

Yep.

setup.py
kitchoi: > like TraitsUI having to update ci-src-requirements.txt while Pyface does not

My bad of jumping into conclusion before verifying. Turns out pyface installs traitsui master too, so the ci-src-requirements.txt in pyface needs to be updated for the maintenance branch too.

rahulporuri:
@kitchoi i totally forgot about it but `ARCHIVE_COMMIT_HASH` won't work as expected if the `.gitattributes` file doesn't exist in the git source repository/directory - https://github.com/enthought/traits/blob/master/.gitattributes. See original PR that added this code - https://github.com/enthought/traits/pull/526

PR #750, opened 2020-10-14
Add api.py to data view packages

### LINKED ISSUE #740: pyface.data_view is missing an api module
There is an `api.py` for most subpackage of `pyface`, e.g. `pyface.grid.api`, `pyface.fields.api`.

But the new `pyface.data_view` does not have an api module. This means it is not clear what are public API intended to be used by projects outside of pyface. Projects would be forced to browse the package to find API intended for them, and may end up depending on APIs intended for internal use only.

corranwebster:
This PR adds `api.py` modules to the various `data_view` sub-packages.  It also modifies examples and documentation to use the these modules when importing objects.

This fixes #740.

### DIFF
--- a/pyface/data_view/api.py
+++ b/pyface/data_view/api.py
@@ -19,7 +19,7 @@ from pyface.data_view.data_formats import (  # noqa: 401
     csv_column_format, csv_format, csv_row_format, from_csv, from_csv_column,
     from_csv_row, from_json, from_npy, html_format, npy_format,
     standard_text_format, to_csv, to_csv_column, to_csv_row, to_json, to_npy,
-    table_format, text_column_format, text_format, text_row_format
+    table_format, text_column_format, text_row_format
 )
 from pyface.data_view.data_view_errors import (  # noqa: 401
     DataViewError, DataViewGetError, DataViewSetError
@@ -29,7 +29,9 @@ from pyface.data_view.data_wrapper import DataWrapper  # noqa: 401
 from pyface.data_view.i_data_view_widget import (  # noqa: 401
     IDataViewWidget, MDataViewWidget
 )
-from pyface.data_view.i_data_wrapper import IDataWrapper, MDataWrapper  # noqa: 401
+from pyface.data_view.i_data_wrapper import (  # noqa: 401
+    DataFormat, IDataWrapper, MDataWrapper, text_format
+)
 from pyface.data_view.index_manager import (  # noqa: 401
     AbstractIndexManager, IntIndexManager, TupleIndexManager
 )
--- a/pyface/data_view/api.py
+++ b/pyface/data_view/api.py
@@ -1,0 +1,35 @@
+# (C) Copyright 2005-2020 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+
+# NOTE: this public-facing API is provisional and may change in future
+# minor releases through until Pyface 8
+
+from pyface.data_view.abstract_data_exporter import AbstractDataExporter  # noqa: 401
+from pyface.data_view.abstract_data_model import AbstractDataModel  # noqa: 401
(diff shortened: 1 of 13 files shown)


pyface/data_view/data_models/api.py
```diff
@@ -9,3 +9,8 @@
 # Thanks for using Enthought open source!
 
 from .array_data_model import ArrayDataModel  # noqa: F401
```
kitchoi, line 11: This import requires NumPy but projects using DataView may only require the RowTableDataModel. It will be annoying to have to install NumPy in order to access this `api` module for things that don't actually require NumPy.
TraitsUI api does something like this and I think it is worth doing:
https://github.com/enthought/traitsui/blob/7b53731606959dfaa7b2e3e8089d00e7a771e312/traitsui/api.py#L28-L33

(This is similar to this issue: https://github.com/enthought/pyface/issues/655)

kitchoi (commented):
Please could we have some way to exercise the `api` modules in CI please? I believe the examples are NOT exercised in CI is that right?

e.g. Updating the imports in the tests to use the `api` module would exercise them. Even better would be to have a simple test case just for exercising the `api` module.

Otherwise, if there is a trivial mistake in `api` (like a typo, or a renamed module), it won't be detected until it is used downstreams.

corranwebster:
Noticed that the main api is missing `DataFormat`.

pyface/data_view/api.py
```diff
@@ -0,0 +1,35 @@
...
+    standard_text_format, to_csv, to_csv_column, to_csv_row, to_json, to_npy,
+    table_format, text_column_format, text_format, text_row_format
+)
+from pyface.data_view.data_view_errors import (  # noqa: 401
+    DataViewError, DataViewGetError, DataViewSetError
+)
+from pyface.data_view.data_view_widget import DataViewWidget  # noqa: 401
+from pyface.data_view.data_wrapper import DataWrapper  # noqa: 401
+from pyface.data_view.i_data_view_widget import (  # noqa: 401
+    IDataViewWidget, MDataViewWidget
```
kitchoi, line 30: If mixin classes are considered part of the public API, then the evaluation on #731 will have to include downstream subclasses we don't know about, which basically mean we can't change them easily. Should we keep these mixins as implementation details and remove them from the `api.py`?

pyface/data_view/api.py
```diff
@@ -0,0 +1,35 @@
...
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+
+# NOTE: this public-facing API is provisional and may change in future
+# minor releases through until Pyface 8
+
```
kitchoi, line 14: If this is turned into a module level docstring then it can be visible in the generated API documentation. e.g.
https://github.com/enthought/traitsui/blob/b855634ec98690b84228c7c7cdcd66b1ebe8b2ba/traitsui/testing/api.py

kitchoi: (Not a request change, just a comment for consideration.)

pyface/data_view/api.py
corranwebster: Actually, I should remove because they are useless here: no code that actually wants to use them can import them from .api because it will trigger toolkit selection and circular imports.  So having them here will just cause noise.

But they are "public" as far as if anyone were to build a new toolkit implementation then they will probably want to use these, but they are also not going to be used often and not usually by end-users. So we can't complain if we make them "private" (ie. not in `.api`) but people use them (they are _intended_ to be used!).

Also I don't see why this changes anything regarding #731 - all of the mixins are potentially used by 3rd party libraries if they are implementing a new toolkit; it's just that there currently aren't any such libraries that we know of.  But in the discussion re #731, it doesn't change the analysis because these mixins are _always_ being mixed in to HasTraits classes.

corranwebster:
Leaving the tests as is, although I would prefer them to just be:
```
def test_api:
    from ... import api
```
as I don't believe we will derive more value from maintenance than it will cost in terms of keeping the tests up to date.

kitchoi (approved):
LGTM

kitchoi:
FWIW, if the test is proving so inconvenient that it can always be removed. It all depends on how often the api module gets changed and gets changed by multiple PRs concurrently.

Having tests on individual items really have values if we have CI jobs that exercises a minimal set of dependencies, i.e. with or without NumPy.

PR #754, opened 2020-10-15
Use TraitsUI maintenance branch for Pyface 7.1 maintenance branch

kitchoi:
[Targeting maint/7.1]

Little did I know pyface also installs traitsui's master. So we need to set the source dependency to point to TraitsUI maintenance branch for Pyface maintenance branch.

(Added this step in #714 for future reference.)


rahulporuri (approved):
LGTM.

PR #760, opened 2020-10-15
Add the missing gitattributes for git archive commit hash

kitchoi:
https://github.com/enthought/pyface/pull/747 should have included this change such that the archive commit hash is replaced properly.

This PR fixes the omission.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# For archives, substitute the commit hash in the setup.py file.
+/setup.py export-subst


rahulporuri:
verified by doing the following : 

- download a ZIP of the pyface repository from the master branch
- check `ARCHIVE_COMMIT_HASH` string in `setup.py` from the downloaded source. It will be `"$Format:%H$"`
- download a ZIP of the pyface repository from this branch
- check `ARCHIVE_COMMIT_HASH` string in `setup.py` from the downloaded source. It will be `077488592d78ab2e1dbb9ab0df04586886742663` - which is the commit hash.

rahulporuri (approved):
LGTM

kitchoi:
Another way to test locally after the fact:
```
>>> pip install --force-reinstall https://github.com/enthought/pyface/archive/077488592d78ab2e1dbb9ab0df04586886742663.tar.gz
```
Using the commit from master results in an error.

PR #762, opened 2020-10-15
Skip load_tests in API documentation

### LINKED ISSUE #757: Should pyface.load_tests be part of the public API documentation?
If one clicks the pyface API documentation, the first thing they see is this:
Screenshot 2020-10-15 at 11 07 05

Problems:
(1) It says it does filtering with regular expression, but it does not say how to supply the regular expression. Users that have installed `pyface` from PyPI would not know how to use this properly. 
(2) Even if the `load_tests` is good enough for public consumption, it is implemented to satisfy unittest load_tests protocol and is expected by called by unittest when a developer runs the tests with `discover`. It seems redundant to be documenting this function in the API documentation, and in such prominent place.

Possible solution:
(a) Engineer the `load_tests` out of the API documentation so that it is not visible.
(b) Mention the name of the environment variable used for the regular expression.

I would prefer solution (a) to (b) because the (b) means the environment variable setup mechanism will become a public API for anyone installing Pyface (e.g. from PyPI), so that's one more public API to maintain.

However in order to do (a), I would suggest making `load_tests` not having to require an environment variable to achieve its goal so that no additional setup is required to run pyface tests. Checking the current toolkit at runtime is a perfectly good way to filter tests that require certain toolkits, that would do away with the environment variable.

kitchoi:
Closes #757

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -226,9 +226,11 @@ autodoc_mock_imports = [
     'IPython.frontend.wx.wx_frontend',
 ]
 
+
 def autodoc_skip_member(app, what, name, obj, skip, options):
     # Skip load_tests
     return skip or name == "load_tests"
+
 
 def setup(app):
     app.connect('autodoc-skip-member', autodoc_skip_member)
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -225,3 +225,10 @@ autodoc_mock_imports = [
     'IPython.frontend.wx',
     'IPython.frontend.wx.wx_frontend',
 ]
+
+def autodoc_skip_member(app, what, name, obj, skip, options):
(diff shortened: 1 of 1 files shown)


rahulporuri:
I'm unable to test this on windows because the docs building process relies on `make`. @aaronayres35 can you review this PR?

aaronayres35 (approved):
LGTM!  FTR now when I build the docs and go to API documentation I see:
Screen Shot 2020-10-15 at 11 19 35 AM

kitchoi:
Thanks @aaronayres35 

Marking this to be backported to the 7.1 branch. This probably does not warrant a changelog update because the `load_tests` was not published in the last release so it has never been shown on the public API documentation. (No one will know it has ever been around...)

PR #765, opened 2020-10-16
Install pyside2 using edm instead of pip

rahulporuri:
This PR updates `etstool.py` to install `PySide2` using EDM instead of installing it using pip. This is now possible because `PySide2` v 5.14.1 is available via EDM.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -105,8 +105,7 @@ source_dependencies = {
 
 extra_dependencies = {
     "pyside": {"pyside"},
-    # XXX once pyside2 is available in EDM, we will want it here
-    "pyside2": set(),
+    "pyside2": {"pyside2"},
     "pyqt": {"pyqt<4.12"},  # FIXME: build of 4.12-1 appears to be bad
     "pyqt5": {"pyqt5"},
     # XXX once wxPython 4 is available in EDM, we will want it here
@@ -193,15 +192,7 @@ def install(edm, runtime, toolkit, environment, editable, source):
         install_pyface,
     ])
 
-    # pip install pyqt5 and pyside2, because we don't have them in EDM yet
-    if toolkit == "pyside2":
-        commands.extend(
-            [
-                "{edm} run -e {environment} -- pip install shiboken2",
-                "{edm} run -e {environment} -- pip install pyside2",
-            ]
-        )
-    elif toolkit == "wx":
+    if toolkit == "wx":
         if sys.platform == "darwin":
             commands.append(
                 "{edm} run -e {environment} -- python -m pip install wxPython<4.1"  # noqa: E501
--- a/etstool.py
+++ b/etstool.py
@@ -185,8 +185,13 @@ def install(edm, runtime, toolkit, environment, editable, source):
             "edm environments create {environment} --force --version={runtime}"
         ]
 
+    if toolkit == "pyside2":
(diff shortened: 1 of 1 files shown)


rahulporuri:
we see a segfault this time - 

```
test_mimeData (pyface.ui.qt4.data_view.tests.test_data_view_item_model.TestDataViewItemModel) ... Fatal Python error: Segmentation fault

Current thread 0x00007f1f34ad9700 (most recent call first):

  File "/home/travis/.edm/envs/pyface-test-3.6-pyside2/lib/python3.6/site-packages/pyface/ui/qt4/data_view/tests/test_data_view_item_model.py", line 61 in test_mimeData
```

note that I am able to reproduce the segfault locally.

corranwebster:
We weren't seeing this segfault under the pip-installed PySide2, right?

I am recalling possibly having seen something like this before where PySide2 returns a different type of "collection of bytes" object that PyQt does - this may be an actual error, although perhaps only one in the test suite.

kitchoi:
> We weren't seeing this segfault under the pip-installed PySide2, right?

I tested master with the pip-installed PySide2 5.14.1 on OSX and could not reproduce the error.

corranwebster:
Yep, I think this is a bug, and it is also present in actual code.  The line just before the crash should read:
```
        raw_data = mime_data.data('text/plain').data()
```
(note the extra `.data()`).  I think a similar change needs to be made in the data wrapper module.

corranwebster:
I have opened #772 to hopefully fix the test failures.

kitchoi:
>> We weren't seeing this segfault under the pip-installed PySide2, right?
>
> I tested master with the pip-installed PySide2 5.14.1 on OSX and could not reproduce the error.

My apologies. Actually, I can reproduce on OSX and PySide2 5.14.1 from PyPI. I think for my previous comment I was running the test suite in an environment installed from PyPI and without NumPy, that particular test case is skipped if NumPy is not available (for it depends on ArrayDataModel, it would be nicer if the test about mime data could be free of NumPy dependency, but that's a separate topic).

To reproduce:
```
$ pip install pyside2==5.14.1 numpy pygments
$ git checkout master
$ pip install .
$ python -m unittest pyface.ui.qt4.data_view.tests.test_data_view_item_model.TestDataViewItemModel.test_mimeData
Segmentation fault: 11
```

kitchoi:
Also, for the record, after upgrading to PySide2 5.15.1, I no longer see the segmentation fault.
```
$ pip install "pyside2>=5.15"
...
Successfully installed pyside2-5.15.1 shiboken2-5.15.1
(pyface-7.1.0-test) kchoi-mbp2019:pyface kchoi$ python -m unittest pyface.ui.qt4.data_view.tests.test_data_view_item_model.TestDataViewItemModel.test_mimeData
/Users/kchoi/Work/ETS/venvs/pyface-7.1.0-test/lib/python3.8/site-packages/traits/traits.py:330: DeprecationWarning: 'TraitMap' trait handler has been deprecated. Use Map instead.
  other.append(TraitMap(map))
.
Ran 1 test in 0.011s

OK 
```
(The deprecation warning there is an open issue on traits)

rahulporuri:
Closing as the experiment revealed missing webkit submodule from our build of PySide2. We will want to try this again when the next update of PySide2 is available via edm.

PR #768, opened 2020-10-16
Get ModalDialogTester tests working on pyqt5

### LINKED ISSUE #302: ModalDialogTester broken on PyQt5
Tests crash. Not sure of cause yet.  Same tests run fine on other backends.

rahulporuri:
The tests were previously being skipped because they failed mysteriously on pyqt5 - on further investigation, I saw weird errors related to two of the tests - a missing AttributeError for `self.gui`. We shouldn't be seeing the errors because `self.gui` exists on the super class `GUITestAssistant`. I did not investigate further why `self.gui` was missing.

I fixed the issue by simply using `GUI` from `pyface.api`

fixes #302

### DIFF
--- a/pyface/ui/qt4/util/tests/test_modal_dialog_tester.py
+++ b/pyface/ui/qt4/util/tests/test_modal_dialog_tester.py
@@ -16,7 +16,7 @@ from io import StringIO
 import platform
 
 from pyface.qt import QtGui
-from pyface.api import Dialog, GUI, MessageDialog, OK, CANCEL
+from pyface.api import Dialog, MessageDialog, OK, CANCEL
 from pyface.toolkit import toolkit_object
 from traits.api import HasStrictTraits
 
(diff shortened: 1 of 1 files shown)


kitchoi:
I think the `self.gui` is expected to be defined by `GuiTestAssistant.setUp`, but is missing because the test case is defined like this:
```
class TestModalDialogTester(unittest.TestCase, GuiTestAssistant):
```

The `unittest.TestCase` has an empty `setUp`, so that shadowed the `setUp` from `GuiTestAssistant`.
To fix this, one would either change the order:
```
class TestModalDialogTester(unittest.TestCase, GuiTestAssistant):
```

Or redefine the `setUp` and `tearDown`:
```
def setUp(self):
    GuiTestAssistant.setUp(self)

def tearDown(self):
    GuiTestAssistant.tearDown(self)
```

rahulporuri:
duh. dunno why i didn't notice that. reversed order.

kitchoi (approved):
Thank you!

Issue #563, opened 2020-07-08
Warnings from building documentation with sphinx


kitchoi:
While running `python etstool.py api-docs`, sphinx emits the following warnings:

<details>

<summary> sphinx build output </summary>

```
sphinx-build -b html -d build/doctrees   source build/html
Running Sphinx v2.3.1
making output directory... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 332 source files that are out of date
updating environment: [new config] 332 added, 0 changed, 0 removed
reading sources... [100%] toolkits                                                                                                                                                          
/Users/kchoi/.edm/envs/pyface-test-3.6-pyqt5/lib/python3.6/site-packages/pyface-7.0.1.dev4-py3.6.egg/pyface/action/action_controller.py:docstring of pyface.action.action_controller.ActionController.can_add_to_menu:6: WARNING: Inline literal start-string without end-string.
/Users/kchoi/.edm/envs/pyface-test-3.6-pyqt5/lib/python3.6/site-packages/pyface-7.0.1.dev4-py3.6.egg/pyface/action/action_controller.py:docstring of pyface.action.action_controller.ActionController.can_add_to_toolbar:6: WARNING: Inline literal start-string without end-string.
/Users/kchoi/.edm/envs/pyface-test-3.6-pyqt5/lib/python3.6/site-packages/pyface-7.0.1.dev4-py3.6.egg/pyface/application.py:docstring of pyface.application:6: WARNING: Unknown interpreted text role "py:method".
/Users/kchoi/.edm/envs/pyface-test-3.6-pyqt5/lib/python3.6/site-packages/pyface-7.0.1.dev4-py3.6.egg/pyface/application.py:docstring of pyface.application:6: WARNING: Unknown interpreted text role "py:method".
/Users/kchoi/.edm/envs/pyface-test-3.6-pyqt5/lib/python3.6/site-packages/pyface-7.0.1.dev4-py3.6.egg/pyface/application.py:docstring of pyface.application:6: WARNING: Unknown interpreted text role "py:method".
/Users/kchoi/.edm/envs/pyface-test-3.6-pyqt5/lib/python3.6/site-packages/pyface-7.0.1.dev4-py3.6.egg/pyface/base_toolkit.py:docstring of pyface.base_toolkit:42: WARNING: Unknown interpreted text role "py_func".
[...]

PR #769, opened 2020-10-16
Address sphinx warnings when building docs

rahulporuri:
Addresses most of the warnings reported in #563 

This PR addresses most of the sphinx documentation warnings reported when building the documentation.

### DIFF
--- a/pyface/timer/do_later.py
+++ b/pyface/timer/do_later.py
@@ -38,7 +38,7 @@ def do_later(callable, *args, **kwargs):
     ----------
     callable : callable
         The callable to call in 50ms time.
-    *args, **kwargs :
+    args, kwargs : tuple, dict
         Arguments to be passed through to the callable.
     """
     return CallbackTimer.single_shot(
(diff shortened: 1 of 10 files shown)


kitchoi:
These two files look orthogonal:
pyface/image/library/icons.zip 
pyface/image/library/std.zip 

Could you revert the changes to these two files please? I still haven't got round to understand why they keep being changed.

pyface/timer/do_later.py
```diff
@@ -38,7 +38,7 @@ def do_later(callable, *args, **kwargs):
     ----------
     callable : callable
         The callable to call in 50ms time.
-    *args, **kwargs :
+    args, kwargs : tuple, dict
```
kitchoi, line 41: Would this suggest `args` and `kwargs` can be passed as positional arguments?

rahulporuri: reverted the change to use `*args` and `**kwargs` again.

kitchoi (approved):
LGTM

kitchoi:
Build on OSX fails due to the sporadic error noted in #556.

corranwebster:
> Could you revert the changes to these two files please? I still haven't got round to understand why they keep being changed.

They are automatically regenerated based on timestamps (and a new mtime means git wants to update them) in `pyface.image.image`.  I remember fixing a timezone problem in them a few years ago after a plane flight broke images in Pyface (so they now use UTC at least), but it's quite possible that they have other issues around the regeneration of the files.  In particular, if you unzip them I am pretty sure that it will re-zip them at least once which will reset the mtime, but there may be other code which is in play.

PR #772, opened 2020-10-19
Fix PySide2 bytearrays

corranwebster:
This makes sure that data is explicitly extracted from QByteArrays rather than relying on toolkit wrappers to autoconvert to `bytes`-like things.

### DIFF
--- a/pyface/ui/qt4/data_view/data_wrapper.py
+++ b/pyface/ui/qt4/data_view/data_wrapper.py
@@ -47,7 +47,7 @@ class DataWrapper(MDataWrapper):
         mimedata : bytes
             The mime media data as bytes.
         """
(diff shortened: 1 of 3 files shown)


corranwebster:
Test failures seem unrelated to bytes issue - it looks like neither WebEngine nor WebKit are in the build?

rahulporuri:
can you just update the base of this branch to target master instead of #765 ? we can get it merged and backported that way.

kitchoi:
@rahulporuri beats me to it, I was about to comment the same.

I could reproduce the segfault in an environment with PySide2 5.14.1 from PyPI, and I can confirm the changes from this branch resolve it.

corranwebster:
Right, I have opened #773 against master.

PR #773, opened 2020-10-19
Fix PySide2 bytearrays

corranwebster:
This makes sure that data is explicitly extracted from QByteArrays rather than relying on toolkit wrappers to autoconvert to bytes-like things.


kitchoi (approved):
LGTM

Issue #696, opened 2020-09-25
Custom editors for DataViews


corranwebster:
The basic implementation of data views relies on the underlying toolkit to provide different editors when editing an item based on the type of the item's editor value.

We would like to provide a hook where users can provide their own editors for items.

This should be a feature of the view, rather than the model, so that different views of the same model can choose to display different editors (or not, as appropriate).

corranwebster:
# Design for DataView Widget Customization

We want to be able to allow individual cells in a data view to:

- render themselves in different ways than the standard
- use different editors, or handle editing non-standard data
- react to simple events, such as mouse clicks, to permit context menus and similar

To handle this, the design being suggested is that the DataViewWidget gains a new
trait holding a list of (for lack of a better term) ItemDelegates which have
4 key features:

- each ItemDelegate has an `is_delegate_for` test which takes the model, row,
  column, etc. and returns True or False if the delegate can handle the data.  This
  allows selection on things like the value type, the row or column of the data,
  whether the item is in a leaf row or not, and so on.

- each DataItemDelegate may provide an ItemRenderer, which is responsible for drawing
  the item when it is not being interacted with.  It is expected that there will be
  a standard set of renderers that are implemented, and because toolkit drawing is
  handled by the toolkit, there will need to be at least some toolkit-specific code
  in these classes.

- each DataItemDelegate may provide an ItemEditor, which has methods for managing the
  lifecycle of a toolkit control being used as an editor.  It is expected that most of
  these will either wrap pyface.fields classes, or TraitsUI editors or views.

- each DataItemDelegate has a list of ItemEventHandlers which can detect whether an
  event matches one that they can handle, and then pereform some action based on that.
  It is possible that drag and drop might use these hooks as well.

It is possible that the ItemDelegate may eventually grow new features as additional
requirements are discovered.

When looking to render a cell, edit a value, or react to an event, the DataViewWidget
(or associated classes, such as a QStyleItemDelegate) will test each delegate until
if finds one that matches the `is_delegate_for` test.  Then the appropriate method from
each will be used to call out to the sub-objects to perform the appropriate action.

For example, the QStyleItemDelegate subclass for the Qt backend might look something
like:
```
    class DataViewItemDelegate(QStyleItemDelegate):
    
        def paint(self, painter, options, index):
            # code to extract data_view, model, etc.
            model = self._data_view.model
            row = self._data_view.model()._index_to_row(index)
            column = self._data_view.model()._index_to_column(index)
            
[...]

PR #782, opened 2020-10-20
DataView Enum value type

corranwebster:
EnumValue type with tests and use in examples.  The value can take on a fixed set of values, each of which can optionally have text, color and an image associated with it.

At present this uses the editor type associated with the edit value, rather than a combobox, however as illustrated by the changes to the examples, it has independent utility.

This is intended as a first step towards allowing custom editors (#696) as enums are the simplest case for when we need an editor that is not covered by the standard `QStyledItemDelegate` editors.


kitchoi:
Just drive-by comment for consideration. The `EnumValue` here share a lot of similarities compared to `TextValue`, `ColorValue`, etc. 
It seems that instead of having so many classes and each implements its own `has_text`, `get_text`, `has_color`, `get_color`, we just need one that takes a number of optional callables, e.g. `to_text` (which is the `EnumValue.format` or the `TextValue.format`), `to_color` (which is the `EnumValue.colors`), ...

(This is kind of related to my early review comment on AbstractValueType to follow Qt pattern of using a mapping instead.)

pyface/data_view/value_types/enum_value.py
```diff
@@ -0,0 +1,162 @@
...
+    def has_color(self, model, row, column):
+        """ Whether or not the value has color data.
+
+        has_color : bool
+            Whether or not the value has data-associated color
+            values.
+        """
+        return (
+            self.colors is not None
+            and self.colors(model.get_value(row, column)) is not None
```
kitchoi, line 105: I'd suggest this `has_color` simply return `self.colors is not None` to match `has_text`. This would be a simpler API for `colors` and this has the added benefits not having to call `model.get_value` twice when there is a valid color (once in `has_colors` and once in `get_color`).

Developers supplying the `colors` callable should make sure any value that can be obtained from the model successfully can be turned into a color. They are going to have to think about this anyway in order to return None for those scenario, so they might as well decide a default color to use.  If we document the API for this `colors` (we should), it would be simpler to say `colors` must return a color, than to say `colors` may return a color or `None`, and then have to explain what the view is going to do if the returned value is `None`.

pyface/data_view/value_types/enum_value.py
```diff
@@ -0,0 +1,162 @@
...
+            The column in the data model being queried.
+
+        Returns
+        -------
+        has_image : bool
+            Whether or not the value has an image associated with it.
+        """
+        return (
+            self.images is not None
+            and self.images(model.get_value(row, column)) is not None
```
kitchoi, line 140: Same here.

pyface/data_view/value_types/enum_value.py
```diff
@@ -0,0 +1,162 @@
...
+        Returns
+        -------
+        text : str
+            The text to display.
+        """
+        return self.format(model.get_value(row, column))
+
+    def has_color(self, model, row, column):
+        """ Whether or not the value has color data.
+
```
kitchoi, line 98: The docstring is a bit broken up here.

pyface/data_view/value_types/enum_value.py
kitchoi: Image is different here that when there are no images, the cell appearance is different. But on the other hand, the data view item model seems to be ignoring anything that is not an IImageResource anyway, so returning, say `1` from `get_image`, would have no consequences.

kitchoi: This prompted me to look into error handling again. When the `DataViewGetError` and `DataViewSetError` were introduced, they were dealing with getting and setting values on the data models. The `images`, `colors` and `format` are dealing with getting different representations for a given value after data model says it is okay, and we might want to have a different error handling for that which can be applied across all value types?  It seems that for some representations, it is not worth crashing the application for. e.g. if `format` fails for whatever reason, does it make sense to crash? Likewise for tooltip, is error from failing to compute a tooltip worth crashing the application?

Maybe a separate issue for this...

pyface/data_view/value_types/enum_value.py
```diff
@@ -0,0 +1,162 @@
...
+    #: A function that converts a value to a string for display.
+    format = Callable(str, update_value_type=True)
+
+    #: A map from valid values to colors.
+    colors = Callable(None, update_value_type=True)
+
+    #: A map from valid values to images.
+    images = Callable(None, update_value_type=True)
+
+    def is_valid(self, model, row, column, value):
```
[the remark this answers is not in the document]
corranwebster: No.  Data flow still goes through here.  For example, if someone decides to use a text editor instead of a combo box, we still need this validation.  Data flow in the `QStyleItemDelegate` that I have sketched out is still through `setData` irrespective of the item editor, which will call this code.

pyface/data_view/value_types/enum_value.py
```diff
@@ -0,0 +1,162 @@
...
+
+class EnumValue(EditableValue):
+    """ Editable value that takes one of a collection of pre-set values.
+
+    Each value can be associated with text, colors and images by supplying
+    functions ``format``, ``colors`` and ``images``, respectively.
+    """
+
+    #: The list of values which are allowed for the value.
+    values = List()
```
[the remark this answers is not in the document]
corranwebster: No problems with dynamic values unless the user wants them to update while the editor is open and actively being edited, which I don't plan to support (I can't think of a valid use-case).

There is an issue of what if this changes and makes the current value invalid (or if the data model changes and makes the current value invalid), but we potentially have that issue in a bunch of places with existing value types.  My feeling is that we let that happen rather than trying to force a valid value.

pyface/data_view/value_types/enum_value.py
corranwebster: The reason for allowing `None` as a return value is that it meshes nicely with using `dict.get` when you have a dictionary mapping values to colors - my initial cut at this was to have the user specify a dictionary of colors (and similarly for images and text), but ended up going with a callable because it is more flexible.

I was in two minds about testing here vs. trying and failing in the getter, and I could go either way.  This is the constant tension between "look before you leap" vs. "better to ask forgiveness than permission".

kitchoi: Hmmm I think I am seeing this `EnumValue` (likewise for `BoolValue` and friends), as a convenient thing that can be used as the default representations for certain types of values. On the one hand, I see there is a pull towards a simple declarative API for defining this object. On the other hand, there is a pull for flexibility to customize. And I think the two can't be met without landing us in the middle where it is neither trivial to use nor trivial to customize.  I think the most flexible approach is for anyone who want flexibility to define their own value type.

pyface/data_view/value_types/enum_value.py
corranwebster: Currently in the implementations if any of the getters raise `DataViewGetError` it will be ignored, so an alternative approach is to have the `get_color` method do the isinstance check and raise if it isn't a `Color`.

But returning `1` from `get_image` would cause problems - and likely a crash with error reporting to the log, because the toolkit code expecting an `IImageResource` will fail.  This is going to be true for any getter that returns a type that the toolkit code can't handle (eg. a number from `get_text` or `get_tooltip`).  This is fine based on the design we have, as these are likely developer errors so we want to fail fast with good diagnostics.

But yes, this is probably a discussion for a separate issue.

kitchoi: > But returning 1 from get_image would cause problems - and likely a crash with error reporting to the log, because the toolkit code expecting an IImageResource will fail. 

For reference: I tried making `get_image` return `1` for the Qt implementation and it did not crash, it just falls through as if there are no images, because it did the `isinstance` check. I did not try with Wx though.  For color, the same cannot be said...

corranwebster: Images are currently ignored by Wx - they are possible, but the column-orientation of the underlying widget makes it hard.

pyface/data_view/value_types/enum_value.py
kitchoi: Also, I notice that with `has_color` + `get_color` + `colors(value) -> Color or None` where `None` means no color, if I put all these together it achieves what we have in TraitsUI TabularEditor's TabularModel using a single method:
https://github.com/enthought/traitsui/blob/8f782bbdee9437d8bc6a2ba941f6d4e21cea8a3d/traitsui/qt4/tabular_model.py#L93-L94

Maybe we could just fall back to that?

corranwebster: I went through that idea way back (see https://github.com/enthought/pyface/pull/543#discussion_r444024714).  It's fine for color but it breaks down for things like editor_value, and I wanted each of the channels to have a similar API.

But I am leaning towards raising if a `None` is returned by the callable, rather than checking for `None` explicitly in `has_color`.  As long as this is exceptional, rather than something which happens all the time, it should be fine.  Whatever way I go, I will make color and image follow the same convention.

kitchoi: > But I am leaning towards raising DataViewGetError if a None is returned by the callable, rather than checking for None explicitly in has_color. As long as this is exceptional, rather than something which happens all the time, it should be fine. Whatever way I go, I will make color and image follow the same convention.

OK!

corranwebster:
> Just drive-by comment for consideration. The `EnumValue` here share a lot of similarities compared to `TextValue`, `ColorValue`, etc.
> It seems that instead of having so many classes and each implements its own `has_text`, `get_text`, `has_color`, `get_color`, we just need one that takes a number of optional callables, e.g. `to_text` (which is the `EnumValue.format` or the `TextValue.format`), `to_color` (which is the `EnumValue.colors`), ...

There are probably some things which make sense to migrate up to the base `EditableValue` because we're effectively using the same implementation in every subclass.  Again, that would be a separate issue.

kitchoi:
> There are probably some things which make sense to migrate up to the base EditableValue because we're effectively using the same implementation in every subclass. Again, that would be a separate issue.

Yep, that sounds reasonable.

corranwebster:
@rahulporuri I think I have addressed all of Kit's concerns (the delta was not big).  If you have time to give this a final once-over, that would be good.

rahulporuri:
@corranwebster i'll review this by EoW. I hope that works for you.

corranwebster:
No rush - it's waited a year, it can wait a while longer.

examples/data_view/column_example.py
```diff
@@ -114,7 +101,10 @@ def get_image(self, model, row, column):
                 HasTraitsRowInfo(
                     title="Country",
                     value="address.country",
-                    value_type=CountryValue(flags=flags),
+                    value_type=EnumValue(
+                        values=['Canada', 'UK', 'USA'],
```
rahulporuri, line 105: nitpicky comment - you could use `list(flags.keys())` here instead of hardcoding the list

pyface/data_view/value_types/enum_value.py
```diff
@@ -0,0 +1,179 @@
+# (C) Copyright 2005-2020 Enthought, Inc., Austin, TX
```
rahulporuri, line 1:
```suggestion
# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
```

pyface/data_view/value_types/tests/test_enum_value.py
```diff
@@ -0,0 +1,110 @@
+# (C) Copyright 2005-2020 Enthought, Inc., Austin, TX
```
rahulporuri, line 1:
```suggestion
# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
```

pyface/data_view/value_types/enum_value.py
```diff
@@ -0,0 +1,179 @@
...
+    colors = Callable(None, update_value_type=True)
+
+    #: A map from valid values to images.
+    images = Callable(None, update_value_type=True)
+
+    def is_valid(self, model, row, column, value):
+        """ Whether or not the value is valid for the data item specified.
+
+        The default implementation returns True for all values.
+
```
rahulporuri, line 40:
```suggestion
```

pyface/data_view/value_types/enum_value.py
```diff
@@ -0,0 +1,179 @@
...
+            The color associated with the cell.
+        """
+        return self.colors(model.get_value(row, column))
+
+    def has_image(self, model, row, column):
+        """ Whether or not the value has an image associated with it.
+
+        The default implementation returns True if ``get_image``
+        returns a non-None value.
+
```
rahulporuri, line 143:
```suggestion
```

rahulporuri (approved):
Mostly LGTM with a few comments/corrections.

examples/data_view/column_example.py
corranwebster: This is possibly better:
```suggestion
                        values=sorted(flags.keys()),
```

PR #783, opened 2020-10-20
Rename version folders for built documentation

kitchoi:
[Targeting gh-pages]

Same motivation as https://github.com/enthought/traitsui/pull/1391

This PR renames the release-X.Y.Z folders to just X.Y.

This will break any external full URL referencing historical releases. However, links to historical releases have not been publicized nor referenced in the documentation, and hence it is unlikely they have been used. (If they are really really need, it is easy to resurrect them again).


kitchoi:
Note that unlike https://github.com/enthought/traitsui/pull/1391, a `release-7.1.0` folder had not been created and so there won't be a `7.1` folder here yet.

rahulporuri (approved):
LGTM. Tested the docs locally and the version pages are accessible (5.1, 6.0, 6.1 and 7.0).

PR #785, opened 2020-10-23
Attempt at a clearer example of data channels

corranwebster:
This bundles country info into an object that contains name and flag information with a custom value type that handles it.

Hopefully this example makes clearer the idea that an item in a model may be more complex than just a scalar value and may have multiple linked data channels.

This is a draft, because it may make sense to bundle this example in with the EnumType PR in some way.

### DIFF
--- a/examples/data_view/row_table_example.py
+++ b/examples/data_view/row_table_example.py
@@ -10,16 +10,18 @@
 
 import logging
 
-from traits.api import Bool, Dict, HasStrictTraits, Instance, Int, Str, List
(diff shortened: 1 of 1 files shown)


examples/data_view/row_table_example.py
```diff
@@ -30,22 +32,29 @@
...
-# The data model
 
 class Address(HasStrictTraits):
 
     street = Str()
 
     city = Str()
 
-    country = Str()
+    country = Instance(Country, allow_none=True)
```
corranwebster, line 57:
```suggestion
    country = Instance(Country, allow_none=False)
```
D'oh!

PR #787, opened 2020-10-26
Dont install traitsui from git source

### LINKED ISSUE #775: Don't install traitsui from source after 7.1.0 release
Once traitsui 7.1.0 is released (currently underway), it would be nice to stop installing traitsui from source - and instead use the released version of traitsui.

See https://github.com/enthought/pyface/blob/master/ci-src-requirements.txt and https://github.com/enthought/pyface/blob/3acd64425c8d9d1a80ad46aeca64852242e0d90e/etstool.py#L94-L99

rahulporuri:
fixes #775 

This PR updates CI/etstool module to install traitsui using edm instead of installing it from git source.

### DIFF
--- a/ci-src-requirements.txt
+++ b/ci-src-requirements.txt
@@ -1,1 +1,0 @@
-git+http://github.com/enthought/traitsui.git#egg=traitsui
--- a/etstool.py
+++ b/etstool.py
@@ -93,12 +93,12 @@ TRAITS_VERSION_REQUIRES = os.environ.get("TRAITS_REQUIRES", "")
 
 dependencies = {
     "traits" + TRAITS_VERSION_REQUIRES,
+    "traitsui",
     "numpy",
     "pygments",
     "coverage",
 }
 
-# NOTE : traitsui is always installed from source
 source_dependencies = {
     "traits": "git+http://github.com/enthought/traits.git#egg=traits",
 }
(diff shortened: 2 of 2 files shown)


ci-src-requirements.txt
```diff
@@ -1 +0,0 @@
-git+http://github.com/enthought/traitsui.git#egg=traitsui
```
kitchoi, line 1: Then I believe you will need to add this back to `source_dependencies` in `etstool.py` for the cron job to run against master.

kitchoi:
I think this should get reviewed by @corranwebster.
Historically Pyface and TraitsUI have been installing each other's master on their CIs, which to me, effectively means they are developed as if they were one distribution. One of the main purposes of separating code into two distributions would be such that the two can be released at different time, and such that the dependent distribution can opt to depend on an older version of another. To achieve that CI should not install each other masters, so I agree the change in principle simply based on the fact that Pyface and TraitsUI are separate distributions. However I am not sure in practice they were actually separate distributions, I am afraid the severed link will bring in even more unnecessary costs.

rahulporuri:
@kitchoi based on offline discussion, it looks like we should be able to merge these PRs now. right?

kitchoi (approved):

kitchoi:
> based on offline discussion, it looks like we should be able to merge these PRs now. right?

For future reference, the offline discussion referred to here is that @corranwebster confirmed to go ahead with installing released TraitsUI in Pyface.

Issue #781, opened 2020-10-20
Stop doing CI runs with Traits 6.0


corranwebster:
After the 7.1 release we are dropping support for Traits 6.0.  Tests should be run against 6.1 and, eventually, 6.2.

PR #788, opened 2020-10-28
Fix deprecation warning at compile time

kitchoi:
This PR resolves the following deprecation warning.

```
/Users/kchoi/.edm/envs/my_env/lib/python3.6/site-packages/pyface/__init__.py:77: DeprecationWarning: invalid escape sequence \.
  exclusion_patterns.append("\.data_view\.")
```

Such warnings are even harder to spot than import time deprecation warning: They only occur the first time the Python file is byte-compiled, usually when the distribution is freshly installed.

The particular line being changed will be removed for #781. But this PR alone is justified and can be backported for bug fix releases.


rahulporuri (approved):
LGTM

PR #790, opened 2020-11-04
Use importlib_metadata instead of pkg_resources for toolkit entry points

rahulporuri:
Use `importlib_metadata` (`importlib.metadata` starting python 3.8) instead of `pkg_resources`, specifically when searching through the entry points, not when looking for resources.

fixes #789

PR #791, opened 2020-11-04
Use importlib_resources instead of pkg_resources for resource files

rahulporuri:
This is a sister PR to #790 

This PR replaces the use of `pkg_resources` to access files from packages with the use of `importlib_resources`.

Note : I dont expect this to work on master because the `importlib_resources` version available on edm is older than the latest
version available via PyPI (3.3.0).

### DIFF
--- a/pyface/resource/resource_manager.py
+++ b/pyface/resource/resource_manager.py
@@ -113,15 +113,12 @@ class ResourceManager(HasTraits):
             # If we come across a reference to a module, use pkg_resources
             # to try and find the image inside of an .egg, .zip, etc.
             if isinstance(dirname, types.ModuleType):
-                from pkg_resources import resource_string
-
+                from importlib_resources import files
                 for path in subdirs:
                     for extension in extensions:
                         searchpath = "%s/%s%s" % (path, basename, extension)
                         try:
-                            data = resource_string(
-                                dirname.__name__, searchpath
-                            )
+                            data = files(dirname.__name__).joinpath(searchpath).read_bytes()
                             return ImageReference(
                                 self.resource_factory, data=data
                             )
--- a/pyface/resource/resource_manager.py
+++ b/pyface/resource/resource_manager.py
@@ -17,6 +17,8 @@ import collections.abc, glob, inspect, os, sys, types
 from os.path import join
 from zipfile import is_zipfile, ZipFile
 
+from importlib_resources import files
+
 from traits.api import HasTraits, Instance, List
 from traits.util.resource import get_path
 
@@ -113,12 +115,15 @@ class ResourceManager(HasTraits):
             # If we come across a reference to a module, use pkg_resources
             # to try and find the image inside of an .egg, .zip, etc.
             if isinstance(dirname, types.ModuleType):
-                from importlib_resources import files
                 for path in subdirs:
                     for extension in extensions:
                         searchpath = "%s/%s%s" % (path, basename, extension)
                         try:
-                            data = files(dirname.__name__).joinpath(searchpath).read_bytes()
+                            data = (
+                                files(dirname.__name__)
+                                .joinpath(searchpath)
+                                .read_bytes()
+                            )
                             return ImageReference(
                                 self.resource_factory, data=data
                             )
--- a/pyface/resource/resource_manager.py
+++ b/pyface/resource/resource_manager.py
@@ -112,8 +112,8 @@ class ResourceManager(HasTraits):
 
         for dirname in resource_path:
 
-            # If we come across a reference to a module, use pkg_resources
-            # to try and find the image inside of an .egg, .zip, etc.
+            # If we come across a reference to a module, try and find the
+            # image inside of an .egg, .zip, etc.
             if isinstance(dirname, types.ModuleType):
                 for path in subdirs:
                     for extension in extensions:
--- a/pyface/tests/test_image_resource.py
+++ b/pyface/tests/test_image_resource.py
@@ -30,8 +30,7 @@ is_pyqt4_windows = (
 
 
 SEARCH_PATH = os.fspath(files("pyface") / "images")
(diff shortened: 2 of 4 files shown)


pyface/resource/resource_manager.py
```diff
@@ -113,15 +113,12 @@ def _locate_image(self, image_name, resource_path, size):
             # If we come across a reference to a module, use pkg_resources
             # to try and find the image inside of an .egg, .zip, etc.
             if isinstance(dirname, types.ModuleType):
-                from pkg_resources import resource_string
-
+                from importlib_resources import files
```
kitchoi, line 116: Could this go to the top-level?

Also I think CI is complaining that it cannot import `files` from `importlib_resources`:
```
ImportError: Failed to import test module: pyface.tests.test_image_resource
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/unittest/loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface/tests/test_image_resource.py", line 17, in <module>
    from importlib_resources import files
ImportError: cannot import name 'files'
```

rahulporuri: yep because we're missing the latest `importlib_resources`

pyface/tests/test_image_resource.py
```diff
@@ -28,9 +30,9 @@
 )
 
 
-SEARCH_PATH = pkg_resources.resource_filename("pyface", "images")
-IMAGE_DIR = pkg_resources.resource_filename(__name__, "images")
-IMAGE_PATH = os.path.join(IMAGE_DIR, "core.png")
+SEARCH_PATH = str(files("pyface") / "images")
```
corranwebster, line 33: I think the "right" way to convert a `Path`-like to a string is `os.fspath`.
```suggestion
SEARCH_PATH = os.fspath(files("pyface") / "images")
```

That's assuming that we actually want a string and that a `Path` won't work better.

pyface/tests/test_image_resource.py
```diff
@@ -28,9 +30,9 @@
...
+IMAGE_DIR = files("pyface.tests") / "images"
+IMAGE_PATH = str(IMAGE_DIR / "core.png")
```
corranwebster, line 35:
```suggestion
IMAGE_PATH = os.fspath(IMAGE_DIR / "core.png")
```

Same here

pyface/resource/resource_manager.py
kitchoi: If this relies on a specific version of `importlib_resources` then I think this needs to be specified in the requirements. Then we would expect installation from EDM to fail because the required version is not available.

rahulporuri:
For us to be able to use `importlib_resources` while maintaining the package data (data in subdirectories), we need to use `importlib_resources>=1.1.0` because 1.1.0 is the first version which supported retrieving data from subdirectories. See https://importlib-resources.readthedocs.io/en/latest/changelog%20%28links%29.html#v1-1-0 and https://github.com/python/importlib_resources/issues/58

pyface/tests/test_image_resource.py
rahulporuri: > That's assuming that we actually want a string and that a Path won't work better.

The rest of the package isnt happy with using `Path`s which is why I converted them to filesystem paths.

pyface/resource/resource_manager.py
```diff
@@ -113,14 +115,14 @@ def _locate_image(self, image_name, resource_path, size):
...
-                from pkg_resources import resource_string
-
                 for path in subdirs:
                     for extension in extensions:
                         searchpath = "%s/%s%s" % (path, basename, extension)
                         try:
-                            data = resource_string(
-                                dirname.__name__, searchpath
+                            data = (
+                                files(dirname.__name__)
```
kitchoi, line 123: I am not sure about this... how does this work? I think `files` returns an iterable? Documentation says it can return None too: https://docs.python.org/3/library/importlib.metadata.html

kitchoi (commented):
First pass review... Some questions.

pyface/resource/resource_manager.py
rahulporuri: this is the recommended migration as documented here - https://importlib-resources.readthedocs.io/en/latest/migration.html#pkg-resources-resource-string.

kitchoi: I was reading `importlib_metadata` documentation but this comes from `importlib_resources`. 🤦 
Thanks.

pyface/tests/test_image_resource.py
```diff
@@ -28,9 +29,9 @@
 )
 
 
-SEARCH_PATH = pkg_resources.resource_filename("pyface", "images")
-IMAGE_DIR = pkg_resources.resource_filename(__name__, "images")
-IMAGE_PATH = os.path.join(IMAGE_DIR, "core.png")
+SEARCH_PATH = os.fspath(files("pyface") / "images")
+IMAGE_DIR = files("pyface.tests") / "images"
+IMAGE_PATH = os.fspath(IMAGE_DIR / "core.png")
```
kitchoi, line 34: Nit: The `IMAGE_DIR` is not converted, which looks inconsistent. Turns out it is only used once here, so we could also just remove `IMAGE_DIR` and have `IMAGE_PATH = os.fspath(files("pyface.tests") / "images" / "core.png")` instead.

kitchoi (approved):
LGTM

pyface/tests/test_image_resource.py
rahulporuri: done.

PR #798, opened 2020-11-12
Extend assertion failure message for debugging sporadic test failure

kitchoi:
I have been able to reproduce the issue #556 on OSX after making the GUI event loop work a lot. This looks to be related to this Qt issue: ~https://bugreports.qt.io/browse/QTBUG-63832~ https://bugreports.qt.io/browse/QTBUG-74233

This PR updates the assertion failure message with the time intervals between calls so that we can confirm the behaviour on CI as well.


kitchoi:
This PR has provided several useful information:
Travis CI OSX build disables AppNap, so that is not the reason why the test fails.
The changes here are not worth keeping any more as the assertion will be removed (see discussion in #556). Closing.

PR #801, opened 2020-11-12
Log debug message if import PythonShell and PythonEditor fail in api.py

kitchoi:
Follow up from https://github.com/enthought/pyface/pull/796
In particular, this: https://github.com/enthought/pyface/pull/796#issuecomment-725538154

If PythonEditor and PythonShell cannot be imported because pygments is not available, it might help if we log a debug message so that downstream projects have more information when the import fails.

The alternative would be to emit a warning, but for developers who don't care about these two classes, this could be too noisy.

### DIFF
--- a/pyface/api.py
+++ b/pyface/api.py
@@ -8,6 +8,7 @@
 #
 # Thanks for using Enthought open source!
 
+import logging as _logging
 
 from .about_dialog import AboutDialog
 from .application import Application
@@ -33,11 +34,15 @@ from .util._optional_dependencies import optional_import as _optional_import
 
 # Excuse pygments dependency (for Qt), otherwise re-raise
 with _optional_import(
-        "pygments", msg="PythonEditor not available due to missing pygments."):
+        "pygments",
+        msg="PythonEditor not available due to missing pygments.",
+        logger=_logging.getLogger(__name__)):
     from .python_editor import PythonEditor
 
 with _optional_import(
-        "pygments", msg="PythonShell not available due to missing pygments."):
+        "pygments",
+        msg="PythonShell not available due to missing pygments.",
+        logger=_logging.getLogger(__name__)):
     from .python_shell import PythonShell
 
 from .sorter import Sorter
--- a/pyface/api.py
+++ b/pyface/api.py
@@ -32,10 +32,12 @@ from .progress_dialog import ProgressDialog
 from .util._optional_dependencies import optional_import as _optional_import
 
 # Excuse pygments dependency (for Qt), otherwise re-raise
-with _optional_import("pygments", msg="PythonEditor not available"):
+with _optional_import(
+        "pygments", msg="PythonEditor not available due to missing pygments."):
     from .python_editor import PythonEditor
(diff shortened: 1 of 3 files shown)


pyface/util/_optional_dependencies.py
```diff
@@ -0,0 +1,40 @@
...
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+""" Utilities for handling optional import dependencies."""
+
+import contextlib
+import logging
+
+
+_PYFACE_LOGGER = logging.getLogger("pyface")
```
corranwebster, line 17: This makes me a bit twitchy, having the logger being used hard-coded.  Perhaps better would be to pass in the logger instance to `optional_import` so it can use the logger of the module where `optional_import` is called from.

pyface/api.py
```diff
@@ -28,13 +28,16 @@
 from .key_pressed_event import KeyPressedEvent
 from .message_dialog import error, information, warning, MessageDialog
 from .progress_dialog import ProgressDialog
-try:
+
+from .util._optional_dependencies import optional_import as _optional_import
+
+# Excuse pygments dependency (for Qt), otherwise re-raise
+with _optional_import("pygments", msg="PythonEditor not available"):
```
corranwebster, line 35: The log message doesn't mention that the reason it isn't available is because Pygments isn't installed - that would be useful to know.

pyface/util/_optional_dependencies.py
```diff
@@ -0,0 +1,40 @@
...
+        Name of the module that may fail to be imported.
+        If matched, the ImportError will be silenced
+    msg : str
+        Log message to be emitted.
+    """
+    try:
+        yield
+    except ImportError as exception:
+        if exception.name == dependency_name:
+            _PYFACE_LOGGER.debug(msg, exc_info=True)
```
corranwebster, line 38: Maybe `info` rather than `debug`?

pyface/util/_optional_dependencies.py
corranwebster: It would also make testing easier

pyface/util/_optional_dependencies.py
kitchoi: I still prefer debug because this is intended for debugging purposes. But that is not a strong preference.

kitchoi: I think typically info message does not contain exception traceback info (the `exc_info=True`) bit. Here I think we want to show the traceback which will include the ImportError on pygments.
I could do both, I suppose?

Like this?
https://github.com/enthought/pyface/blob/7e32dbfd55de290ecde1da3ef792ff5be5da40c1/pyface/base_toolkit.py#L240-L241

kitchoi:
Thank you @corranwebster 
I have addressed the comments, apart from the logging level: I am not entirely sure which of `info` or `debug` we should use or if we should log both.

corranwebster (approved):
LGTM

PR #807, opened 2020-11-13
Fix Qt widget visibility

### LINKED ISSUE #786: DataView widget disappears from UI when embedded in a tab
Hi there ! I am experiencing a UI issue when trying to use the `DataView` widget. When I include it in a UI with a tab, it disappears if I switch focus to another tab.

Two implementations were tried to include `DataView` in my traitsui `View`. Either embedding it in a traitsui `CustomEditor`, to be then placed in a ```UItem('data_model', editor=DataModelEditor(), style='custom')```:

```Python3
def _data_view_factory(parent, editor):
    """ Returns the Qt Widget corresponding to the editor. """
    widget = DataViewWidget(
        data_model=editor.value,
    )
    editor._widget = widget
    widget._create()
    return widget.control


class DataModelEditor(CustomEditor):

    #: Overwrite the parent class trait to include a default
    #: factory
    factory = Callable(_data_view_factory)

```

or either subclassing `traitsui.qt4.editor.Editor` and `traitsui.api.EditorFactory` (in a a dirty way, I admit it), such as in the following snippet:

```Python3
class MyEditor(Editor):

    def init(self, parent):
        widget = DataViewWidget(
            data_model=self.value,
        )
        widget._create()
        self.control = widget.control

    def update_editor(self):
        self.control._widget.data_model = self.value


class MyEditorFactory(EditorFactory):

    def simple_editor(self, ui, obj, name, description, parent):
        return MyEditor(parent, factory=self, ui=ui,
                        object=obj,
                        name=name,
                        description=description)
```

Any idea what's happening ?

Thanks in advance for your help !

### LINKED ISSUE #806: Visibility not tracked correctly in Qt
In the following code, when run under Qt:
```
import sys

from traits.api import Instance
from pyface.api import ApplicationWindow, GUI
from pyface.fields.api import TextField


    widget = Instance(TextField)

    def _create_contents(self, parent):
        """ Create and return the window's contents.
        Parameters
        ----------
        parent : toolkit control
            The window's toolkit control to be used as the parent for
            widgets in the contents.
            
        Returns
        -------
        control : toolkit control
            A control to be used for contents of the window.
        """
        self.widget = TextField(parent=parent)
        self.widget._create()
        return self.widget.control
    


if __name__ == '__main__':
    window = MainWindow()
    
    gui = GUI()
    gui.process_events()
    window.open()
    gui.process_events()

    # text field is visible at this point
    print('visible:', window.widget.visible)
    print('isVisible():', window.widget.control.isVisible())
    print('isHidden():', window.widget.control.isHidden())
    print()

    window.visible = False
    gui.process_events()
    print('visible:', window.widget.visible)
    print('isVisible():', window.widget.control.isVisible())
    print('isHidden():', window.widget.control.isHidden())
    print()

    window.visible = True
    gui.process_events()
    # text field is not visible at this point, but should be
    print('visible:', window.widget.visible)
    print('isVisible():', window.widget.control.isVisible())
    print('isHidden():', window.widget.control.isHidden())
    
    gui.start_event_loop()
```
if you run it then the text field is not visible.  This is in contrast with if we had created a `QLineEdit` directly in the `MainWindow` then it would be visible after all the interactions.  Similarly, when run in the Wx backend (with the appropriate changes to remove the Qt-specific code), this ends with the `TextField` visible.

This is the same issue as #786 but stripped down to as minimal an example as possible to remove possible DataView and TraitsUI interactions.

The cause appears to be https://github.com/enthought/pyface/blob/e4e4b3303fdcb0ee002d6414a49a2240d3e185e0/pyface/ui/qt4/widget.py#L124-L125
What is happening is that when the parent window is has it's visible state set to `False`, the child gets a `QHideEvent` because it is being removed from view, and this then triggers setting the `visible` to `False`, which in Qt's behaviour model means that the widget is not just invisible, it is hidden.  This means that the text box does not appear when the enclosing window becomes visible.

I think this should be tracking the `isHidden()` state rather than the `isVisible()` state - this would match the behaviour of `visible_when` in TraitsUI as well as the behaviour of bare toolkit widgets and the Wx backend.

corranwebster:
This fixes #806 and fixes #786 along the lines of the suggested work-arounds from those issues.

Tests for expected behaviour are included.

### DIFF
--- a/pyface/tests/test_widget.py
+++ b/pyface/tests/test_widget.py
@@ -11,8 +11,10 @@
 
 import unittest
 
+from traits.api import Instance
 from traits.testing.unittest_tools import UnittestTools
 
+from ..application_window import ApplicationWindow
 from ..toolkit import toolkit_object
 from ..widget import Widget
 
@@ -37,6 +39,28 @@ class ConcreteWidget(Widget):
         else:
             control = None
         return control
+
+
+class MainWindow(ApplicationWindow):
+
+    widget = Instance(ConcreteWidget)
+
+    def _create_contents(self, parent):
+        """ Create and return the window's contents.
+        Parameters
+        ----------
+        parent : toolkit control
+            The window's toolkit control to be used as the parent for
+            widgets in the contents.
+
+        Returns
+        -------
+        control : toolkit control
+            A control to be used for contents of the window.
+        """
+        self.widget = ConcreteWidget(parent=parent)
+        self.widget._create()
+        return self.widget.control
 
 
 class TestWidget(unittest.TestCase, UnittestTools):
@@ -131,6 +155,43 @@ class TestConcreteWidget(unittest.TestCase, GuiTestAssistant):
                 self.widget.visible = False
 
         self.assertFalse(self.widget.control.isVisible())
+
+    def test_contents_visible(self):
+        window = MainWindow()
+        window._create()
+
+        try:
+            with self.event_loop():
+                window.open()
+
+            with self.assertTraitDoesNotChange(window.widget, "visible"):
+                with self.event_loop():
+                    window.visible = False
+
+            with self.assertTraitDoesNotChange(window.widget, "visible"):
+                with self.event_loop():
+                    window.visible = True
+        finally:
+            window.destroy()
+
+    def test_contents_hidden(self):
+        window = MainWindow()
+        window._create()
+
+        try:
+            with self.event_loop():
+                window.open()
+                window.widget.visible = False
+
+            with self.assertTraitDoesNotChange(window.widget, "visible"):
+                with self.event_loop():
+                    window.visible = False
+
+            with self.assertTraitDoesNotChange(window.widget, "visible"):
+                with self.event_loop():
+                    window.visible = True
+        finally:
+            window.destroy()
 
     def test_enable(self):
         with self.event_loop():
--- a/pyface/tests/test_widget.py
+++ b/pyface/tests/test_widget.py
@@ -166,10 +166,12 @@ class TestConcreteWidget(unittest.TestCase, GuiTestAssistant):
             with self.event_loop():
                 window.open()
 
+            # widget visible trait stays True when parent is hidden
             with self.assertTraitDoesNotChange(window.widget, "visible"):
                 with self.event_loop():
                     window.visible = False
 
+            # widget visible trait stays True when parent is shown
             with self.assertTraitDoesNotChange(window.widget, "visible"):
                 with self.event_loop():
                     window.visible = True
@@ -185,10 +187,12 @@ class TestConcreteWidget(unittest.TestCase, GuiTestAssistant):
                 window.open()
                 window.widget.visible = False
 
+            # widget visible trait stays False when parent is hidden
             with self.assertTraitDoesNotChange(window.widget, "visible"):
                 with self.event_loop():
                     window.visible = False
 
+            # widget visible trait stays False when parent is shown
             with self.assertTraitDoesNotChange(window.widget, "visible"):
                 with self.event_loop():
                     window.visible = True
@@ -204,6 +208,7 @@ class TestConcreteWidget(unittest.TestCase, GuiTestAssistant):
             with self.event_loop():
                 window.open()
 
+            # widget visibile trait stays True when parent is hidden
             with self.assertTraitDoesNotChange(window.widget, "visible"):
                 with self.event_loop():
                     window.visible = False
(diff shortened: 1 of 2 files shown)


pyface/ui/qt4/widget.py
```diff
@@ -105,6 +105,10 @@ def __event_filter_default(self):
 class WidgetEventFilter(QtCore.QObject):
     """ An internal class that watches for certain events on behalf of the
     Widget instance.
+
+    This filter watches for show and hide events to make sure that visible
+    state of the widget is the opposite of Qt's isHidden() state.  This is
+    needed in case other code hides the toolkit widget
```
kitchoi, line 111: I believe the `visible` trait is meant for toggling, not for query.
e.g. one can toggle visibility like this:
```
def toggle_visibility(widget):
    widget.visible = not widget.visible
```

But if one wants to know whether a widget is actually visible by the user, this is not the trait to use.
https://github.com/enthought/pyface/blob/e9b7d968f5fbfb239d70c4c1e095e0672aa0b458/pyface/i_widget.py#L29-L30

Just reading the documentation, I'd have thought the trait can also be used for querying the visibility state.

While using the attribute name `visible` to mean `not hidden` might be convenient for the toggling use case, it will preclude the addition of a separate attribute that actually reflects visibility should a new use case for query arise.  Shall we just add a `hidden` trait to decouple them?  Then the `visible` trait can be implemented like a property, where its getter reflects the value from Qt `isVisible`.

kitchoi (commented):
Some suggestions on the tests, but the main question is:

Shall we just add a hidden trait to decouple the query of visibility and the toggling of visible?

corranwebster:
This PR is a bugfix PR, so while there is some value in splitting out hidden state vs. visible state as separate traits, that's a separate issue and would be a separate PR.

pyface/tests/test_widget.py
```diff
@@ -132,6 +158,75 @@ def test_visible(self):
...
+            with self.event_loop():
+                window.open()
+
+            with self.assertTraitDoesNotChange(window.widget, "visible"):
+                with self.event_loop():
+                    window.visible = False
+
+            with self.assertTraitDoesNotChange(window.widget, "visible"):
+                with self.event_loop():
+                    window.visible = True
```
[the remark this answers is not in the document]
corranwebster: This is testing that when you toggle the visibility of the parent, the `visible` trait of the child does not change, which is the desired behaviour.

pyface/tests/test_widget.py
```diff
@@ -132,6 +158,75 @@ def test_visible(self):
...
+                window.open()
+                window.widget.visible = False
+
+            with self.assertTraitDoesNotChange(window.widget, "visible"):
+                with self.event_loop():
+                    window.visible = False
+
+            with self.assertTraitDoesNotChange(window.widget, "visible"):
+                with self.event_loop():
+                    window.visible = True
```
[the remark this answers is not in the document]
corranwebster: Same thing here - toggling the visiblity of the parent does not toggle visibility of the child.

pyface/ui/qt4/widget.py
corranwebster: For better or worse, `visible` is tracking the opposite of what Qt calls `hidden`. As mentioned in the comments for the issue, this is matching the way that `visible` behaves on the Wx backend and the way that TraitsUI's `visible_when` behaves (and I think, although I have not checked, how the `visible` attribute of an `Action` works).

So this is the right change for now.

There may well be room for a read-only trait that can be queried for the "am I shown on the screen right now" state, but for backwards compatibility it should probably not be called `visible`.

pyface/tests/test_widget.py
```diff
@@ -132,6 +158,75 @@ def test_visible(self):
...
+
+            self.assertFalse(window.widget.control.isVisible())
+            self.assertTrue(window.widget.control.isHidden())
+        finally:
+            window.destroy()
+
```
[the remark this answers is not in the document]
corranwebster: I was trying to minimize the number of Qt-specific tests, but I will add these.

pyface/tests/test_widget.py
```diff
@@ -132,6 +158,75 @@ def test_visible(self):
...
+                with self.event_loop():
+                    window.visible = False
+
+            self.assertFalse(window.widget.control.isVisible())
+            self.assertFalse(window.widget.control.isHidden())
+
+            with self.assertTraitChanges(window.widget, "visible"):
+                with self.event_loop():
+                    window.widget.control.hide()
+
```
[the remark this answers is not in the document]
corranwebster: I'd rather not add explicit value tests for the `assertTraitDoesNotChange` - it didn't change so additional asserts are not illuminating.  I'll add comments about what is going on.

pyface/tests/test_widget.py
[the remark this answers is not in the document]
kitchoi: Could you please add this explanation to the test? This being the intent of the test wasn't obvious to me at all when I read the test. If it is not obvious to me, it is likely to be the case for someone else, and it would save them a lot of time puzzling or getting confused if we have just this sentence as a comment. Same for the other tests.

pyface/ui/qt4/widget.py
kitchoi: I did read the wx code but it is not obvious to me that the behaviour in wx is also about toggling.  Maybe it is, but none of the tests here checks the wx states to ensure this is what the trait does, so I am not sure.

If it is intended to behave like `visible_when`, then the `visible` is a toggle for the developers to change the widget visibility. It should not be updated via Qt event right?

If I meant for the widget to be hidden, then `widget.visible` is false. In that case, `if event_type == QtCore.QEvent.Show` should not occur, so the first branch is not used.  In fact, if I remove this branch, none of the tests fail.

I think the code would be clearer if we have this instead:
```
if event_type == QtCore.QEvent.Hide:
    widge.visible = not widget.control.isHidden()
```
Then it is more straightforward `visible` is simply `not hidden`. If I make this change, all the tests (include the ones I proposed) still pass.

We should also update the documentation for `visible` to indicate that this trait cannot be used for query, otherwise it is tempting to misuse it:
https://github.com/enthought/pyface/blob/e9b7d968f5fbfb239d70c4c1e095e0672aa0b458/pyface/i_widget.py#L29-L30

kitchoi: > There may well be room for a read-only trait that can be queried for the "am I shown on the screen right now" state, but for backwards compatibility it should probably not be called visible.

Sigh, I don't know what that new trait should be called if `visible` is already taken here. My only hope is that it won't be called `hidden`, because it would just be doing exactly the opposite of Qt, deliberately messing with developers.

corranwebster: > if event_type == QtCore.QEvent.Hide:
>     widge.visible = not widget.control.isHidden()

This was my first instinct too, but the problem is that _setting_ (ie. calls to `setVisible()`) has side-effects in Qt, so we should set _only_ when we need to.  I think I tried this and it caused issues with the display of the widget (but I could be mis-remembering the path I went through to come up with this fix - I certainly tried something like this).  Qt

As for another trait, yes I agree `hidden` is a bad name, maybe `displayed` or something like that - and I certainly don't want people getting intuition from the way Qt handles visibility/hidden state.  But that's a discussion for a separate issue.  Note that I haven't ever needed that information at the sort of level that Pyface etc. are working (it can be important for eg. halting video playback to save CPU, but if you were implementing that, you're probably working directly with Qt).

kitchoi: > This was my first instinct too, but the problem is that setting (ie. calls to setVisible()) has side-effects in Qt, so we should set only when we need to. I think I tried this and it caused issues with the display of the widget (but I could be mis-remembering the path I went through to come up with this fix - I certainly tried something like this). Qt

That would be weird, setting the `visible` will only cause `setVisible` to be invoked if the trait has changed, right? If we can't simply set the `visible` trait to sync up with Qt, it smells like a bigger problem and is worth looking into.

kitchoi: If we cannot reproduce the "it caused issues with the display of the widget", then I'd still prefer doing the obviously right thing: set the `visible` to what Qt thinks it is. If the issue you said exists, we can fix it when it surfaces, knowing what exactly the problem is.  Otherwise it feels like we are working around something we don't exactly know what, and that workaround might introduce new problems, and more workarounds to the point where nothing is fixable.

corranwebster: FWIW I think I am going crazy - I simplified the whole thing to one branch:
```
        if event_type in {QtCore.QEvent.Show, QtCore.QEvent.Hide}:
            widget.visible = not widget.control.isHidden()
```
and it seems to work correctly.  I could have sworn that while I was hammering out the fix I tried either the above, or what you were suggesting and there were problems, but I can't see any right now, and tests pass.  There must have been some other orthogonal issue that was messing with things.

kitchoi (approved):
LGTM.
Thank you! I can verify (on Python 3.9 + OSX + PySide2 5.15) that this fixed #786.

PR #813, opened 2020-11-25
Copy Undo over from apptools

aaronayres35:
As mentioned in step 6a of https://github.com/enthought/ets/blob/master/docs/source/eeps/eep-6.rst
this PR simply copies over the `apptools.undo` subpackage from apptools (and updates references to apptools in the code to reference pyface instead).

After the next pyface release we will want to deprecate `apptools.undo` in apptools.


pyface/undo/abstract_command.py
```diff
@@ -0,0 +1,76 @@
+# ------------------------------------------------------------------------------
```
rahulporuri, line 1: We need to add an additional enthought copyright header here which starts from the year 2007. This applies for this file and other files which have the Riverbank computing limited copyright header.

pyface/undo/abstract_command.py
```diff
@@ -0,0 +1,76 @@
...
+@provides(ICommand)
+class AbstractCommand(HasTraits):
+    """The AbstractCommand class is an abstract base class that implements the
+    ICommand interface.
+    """
+
+    #### 'ICommand' interface #################################################
+
+    # This is the data on which the command operates.
+    data = Any
```
rahulporuri, line 31: I think it'll be good to follow the new convention of using `data = Any()` instead of `data = Any` for trait definitions. This applies here and all of the other trait definitions on new classes in this PR.

pyface/undo/undo_manager.py
```diff
@@ -0,0 +1,122 @@
...
+class UndoManager(HasTraits):
+    """The UndoManager class is the default implementation of the
+    IUndoManager interface.
+    """
+
+    #### 'IUndoManager' interface #############################################
+
+    # This is the currently active command stack and may be None.  Typically it
+    # is set when some sort of editor becomes active.
+    active_stack = Instance("apptools.undo.api.ICommandStack")
```
rahulporuri, line 41: not sure why we can't just import and use `ICommandStack` here instead of using the string.

pyface/undo/i_undo_manager.py
```diff
@@ -0,0 +1,63 @@
...
+    """The undo manager interface.  An undo manager is responsible for one or
+    more command stacks.  Typically an application would have a single undo
+    manager.
+    """
+
+    #### 'IUndoManager' interface #############################################
+
+    # This is the currently active command stack and may be None.  Typically it
+    # is set when some sort of editor becomes active.
+    active_stack = Instance("apptools.undo.api.ICommandStack")
```
rahulporuri, line 29: It looks like `IUndoManager` and `ICommandStack` depend on one another - which is why we can't import and use `ICommandStack` here. Can you add a comment here regarding the same?

rahulporuri (commented):
1. Can you also move the documentation for undo from apptools to pyface?
2. Can you check the test coverage for all of the new code? I think we will want to/need to add more tests.

pyface/undo/undo_manager.py
corranwebster: There is a circular import, I think - `UndoManager` references `ICommandStack` and vice-versa.

docs/source/undo.rst
```diff
@@ -0,0 +1,275 @@
...
+
+RedoAction
+..........
+
+The ``RedoAction`` class is a canned PyFace action that redoes the last
+command undone of the active command stack.
+
+
+.. _API: api/index.html
+.. _example: https://svn.enthought.com/enthought/browser/AppTools/trunk/examples/undo/
```
aaronayres35, line 275: This link is clearly out of place here, but I believe it was actually previously broken on apptools as well.  Likewise, this final section of "API Overview" seems unnecessary with auto built api docs.  This file needs to be updated, but I think I will leave those changes for a separate PR unless a reviewer would like to see them done here, in which case I am happy to do so

aaronayres35:
> 3. Can you check the test coverage for all of the new code? I think we will want to/need to add more tests.

Here is a screenshot of the coverage report:
Screen Shot 2020-11-30 at 9 05 47 AM

docs/source/undo.rst
rahulporuri: Let's get them fixed in this PR instead of after the PR.

rahulporuri:
>  Should I try to write them as part of this PR, or do that in a subsequent PR @rahulporuri?

Can you add those as part of this PR? IMO its not too much to review.

aaronayres35:
`pyface.undo.action` is currently untested, but I am unsure on what the benefits would be of testing.  The classes exposed by the api basically just have a `perform` method which calls a method on `command_stack` or `undo_manager` which are already tested.  Also, testing if `perform` works is handled by the tests for `Action`.  It feels incorrect to leave this sub package without tests, but I am unsure what specifically it would be useful to test (rather than just writing noisy tests for the sake of writing something).

One thing I see that may be beneficial is `AbstractCommandStackAction` has a method `_on_stack_updated` that simply calls `self._update_action()` when the `active_stack` is updated.  This can change whether the actions are enabled, and their names, and probably warrants a test.

docs/source/undo.rst
```diff
@@ -0,0 +1,72 @@
...
+  A command is done by pushing it onto a command stack.  The last command can
+  be undone and redone by calling appropriate command stack methods.  It is
+  also possible to move the stack's position to any point and the command stack
+  will ensure that commands are undone or redone as required.
+
+  A command stack maintains a *clean* state which is updated as commands are
+  done and undone.  It may be explicitly set, for example when the data being
+  manipulated by the commands is saved to disk.
+
+  Canned PyFace actions are provided as wrappers around command stack methods
```
rahulporuri, line 45: What's a "Canned PyFace action"? Let's just go with "PyFace actions ..."

examples/undo/commands.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # The data being operated on.
+    data = Instance(Label)
+
+    # The name of the command.
+    name = Str("&Increment size")
+
+    #### Private interface ####################################################
+
+    _incremented_by = Int
```
rahulporuri, line 50:
```suggestion
    _incremented_by = Int()
```

nit-picky comment

examples/undo/commands.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # The data being operated on.
+    data = Instance(Label)
+
+    # The name of the command.
+    name = Str("&Decrement size")
+
+    #### Private interface ####################################################
+
+    _decremented_by = Int
```
rahulporuri, line 93:
```suggestion
    _decremented_by = Int()
```

examples/undo/example.py
```diff
@@ -0,0 +1,96 @@
...
+# Description: <Enthought undo package component>
+# -----------------------------------------------------------------------------
+
+
+# Standard library imports.
+import logging
+
+# Enthought library imports.
+from pyface.api import GUI, YES
+from pyface.workbench.api import Workbench
```
rahulporuri, line 31: can you open an issue regd porting the undo example code from pyface workbench to pyface tasks?

examples/undo/model.py
```diff
@@ -0,0 +1,56 @@
...
+from traits.api import Enum, HasTraits, Int, Str
+
+
+class Label(HasTraits):
+    """The Label class implements the data model for a label."""
+
+    #### 'Label' interface ####################################################
+
+    # The name.
+    name = Str
```
rahulporuri, line 36:
```suggestion
    name = Str()
```

pyface/undo/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+""" Supports undoing and scripting application commands.
+    Part of the AppTools project of the Enthought Tool Suite.
+"""
```
rahulporuri, line 12:
```suggestion
""" Supports undoing and scripting application commands.
"""
```

pyface/undo/action/tests/test_actions.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def setUp(self):
+        self.stack = CommandStack()
+        self.undo_manager = UndoManager()
+        self.stack.undo_manager = self.undo_manager
+        self.undo_manager.active_stack = self.stack
+
+        self.command = SimpleCommand()
+
+    def test_udpate(self):
```
rahulporuri, line 31:
```suggestion
    def test_update(self):
```

rahulporuri (approved):
Mostly LGTM with a few minor comments.

examples/undo/example.py
aaronayres35: Opened #821

rahulporuri:
@aaronayres35 before you merge, can you update the coverage report?

aaronayres35:
> > 1. Can you check the test coverage for all of the new code? I think we will want to/need to add more tests.
> 
> Here is a screenshot of the coverage report:
> Screen Shot 2020-11-30 at 9 05 47 AM

This is the updated coverage report after the added tests:
Screen Shot 2020-12-01 at 8 06 37 AM

PR #827, opened 2020-12-02
Set IS_RELEASED back to False

aaronayres35:
This PR simply sets `IS_RELEASED` back to False.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ MAJOR = 7
 MINOR = 2
 MICRO = 0
 PRERELEASE = "rc1"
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM.

Not sure if the convention is to set the `"rc1"`  back to `""` or if that's done when the actual release is made.

aaronayres35:
> LGTM.
> 
> Not sure if the convention is to set the `"rc1"` back to `""` or if that's done when the actual release is made.

For 7.1.0 we just left `"rc1"` at this stage and reverted it to `""` during the actual release

PR #837, opened 2020-12-09
[WIP] Adding menu items and tool bars to tasks applications

rahulporuri:
This PR adds some documentation on how to add menu items and tool bars to a tasks application. Note that the document is not included in the index at the moment as i am not sure where it fits in to the existing pyface docs.


aaronayres35 (approved):
LGTM! (Although note I did not build the docs)

The only comment I have is that I think there could be more context added about menubars, menu items, and tool bars, etc.  At this point the documentation portion doesn't provide a _ton_ on top of the example added in this PR.  Some of the non-code words in the documentation could even be added as comments to the example code and give a similar effect.
I still believe it is good to have a separate example code file and typed documentation, but a little more context in the written documentation I think would be beneficial (even if the content is pretty straight forward)

rahulporuri:
I am unable to revert this to a draft PR so i'm marking this as a WIP for now. Thanks for the review and the comments @aaronayres35 . I'll try to add more meaningful text and also extend the example to actually use a TasksApplication instead of just a Task - that way, the tutorial can be extended to also include contributing menu items via tasks extensions.

PR #838, opened 2020-12-09
Backports for 7.2.0 and update changelog

aaronayres35:
[Targeting maint/7.2]

This PR backports merged PRs to the `maint/7.2` branch in preparation for the upcoming release. ~It also updates the changelog to include the relevant backported PRs. Only the last commit which updates the changelog needs to be reviewed (the others have been reviewed already).~ EDIT: None of the changes are relevant for changelog


CHANGES.txt
```diff
@@ -37,6 +37,7 @@ Features
 
 Fixes
 
+* Fix loading resource images from a module that is not a package (#836)
```
kitchoi, line 40: I think the one change this made compared to 7.1.0 was to do with fixing the behaviour of unexpectedly getting unittest `__main__` module when the code is invoked as a script from within a test 🤯 . Prior to switching from pkg_resources to importlib_resources, loading resource images from a non-package does work owing to the use of `__file__`. So it might not need to be included in the changelog?

aaronayres35: ah yes, that makes sense.  I will remove it, thanks!

kitchoi (approved):
LGTM

Issue #755, opened 2020-10-15
Autodoc tools not finding classes imported via `toolkit` mechanism


corranwebster:
For example, the documentation page for `pyface.widget` is blank, and references of the form
```
:class:`pyface.widget.Widget`
```
don't work.  We should at a minimum have the namespace of these modules visible in the docs, and ideally get the links to work as well (but that may be harder)

PR #843, opened 2020-12-10
Improve documentation for pyface.action and pyface.tasks.action submodules

rahulporuri:
This PR improves documentation for the following submodules - `pyface.tasks.action` and `pyface.action` submodules.

Specifically, 

- the api modules now have a module docstring, which gets used in the autogenerated API docs
- the in-code comments for traits have been updated such that sphinx picks them up
- two spelling mistakes have been fixed.

### DIFF
--- a/pyface/action/api.py
+++ b/pyface/action/api.py
@@ -11,10 +11,10 @@
 
 """
 
-API for the ``pyface.action`` submodule.
+API for the ``pyface.action`` subpackage.
 
-Action and subclasses
----------------------
+Actions
+-------
 
 - :class:`~.Action`
 - :class:`~.FieldAction`
@@ -33,29 +33,24 @@ Action Event
 
 - :class:`~.ActionEvent`
 
-Action Manager and subclasses
------------------------------
+Action Managers
+---------------
 
 - :class:`~.ActionManager`
-
-Toolkit-specific subclasses:
-
 - ``MenuManager``
 - ``MenuBarManager``
+- ``StatusBarManager``
 - ``ToolBarManager``
 - ``ToolPaletteManager``, only for the Wx toolkit.
-- ``StatusBarManager``. Note that the ``StatusBarManager`` is not actually an
-  ``ActionManager`` subclass but it is listed here given it's similarily in
-  function to the other subclasses.
 
-Action Manager item and subclasses
-----------------------------------
+Action Manager Items
+--------------------
 
 - :class:`~.ActionManagerItem`
 - :class:`~.ActionItem`
 
-Group and subclasses
---------------------
+Layout support
+--------------
 
 - :class:`~.Group`
 - :class:`~.Separator`
--- a/pyface/action/api.py
+++ b/pyface/action/api.py
@@ -9,6 +9,68 @@
 # Thanks for using Enthought open source!
 
 
+"""
+
+API for the ``pyface.action`` submodule.
+
+Action and subclasses
+---------------------
+
+- :class:`~.Action`
+- :class:`~.FieldAction`
+- :class:`~.GUIApplicationAction`
+- :class:`~.ListeningAction`
+- :class:`~.TraitsUIWidgetAction`
+- :class:`~.WindowAction`
+
+Action Controller
+-----------------
+
+- :class:`~.ActionController`
+
+Action Event
+------------
+
+- :class:`~.ActionEvent`
+
+Action Manager and subclasses
+-----------------------------
+
+- :class:`~.ActionManager`
+
+Toolkit-specific subclasses:
+
+- ``MenuManager``
+- ``MenuBarManager``
+- ``ToolBarManager``
+- ``ToolPaletteManager``, only for the Wx toolkit.
+- ``StatusBarManager``. Note that the ``StatusBarManager`` is not actually an
+  ``ActionManager`` subclass but it is listed here given it's similarily in
+  function to the other subclasses.
+
+Action Manager item and subclasses
+----------------------------------
+
+- :class:`~.ActionManagerItem`
+- :class:`~.ActionItem`
+
+Group and subclasses
+--------------------
+
+- :class:`~.Group`
+- :class:`~.Separator`
+
+Useful Application and Window actions
+-------------------------------------
+
+- :class:`~.AboutAction`
+- :class:`~.CloseActiveWindowAction`
+- :class:`~.CreateWindowAction`
+- :class:`~.ExitAction`
+- :class:`~.CloseWindowAction`
+
+"""
+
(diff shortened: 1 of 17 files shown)


pyface/action/api.py
```diff
@@ -9,6 +9,68 @@
...
+- :class:`~.ActionManager`
+
+Toolkit-specific subclasses:
+
+- ``MenuManager``
+- ``MenuBarManager``
+- ``ToolBarManager``
+- ``ToolPaletteManager``, only for the Wx toolkit.
+- ``StatusBarManager``. Note that the ``StatusBarManager`` is not actually an
+  ``ActionManager`` subclass but it is listed here given it's similarily in
```
aaronayres35, line 48: should be "similarity"

pyface/action/menu_bar_manager.py
```diff
@@ -8,7 +8,11 @@
 #
 # Thanks for using Enthought open source!
 
-""" A menu bar manager realizes itself in errr, a menu bar control. """
+""" A toolkit-specific menu bar manager that realizes itself in a menu bar
+control.
+
+- :attr:`~.MenuBarManager`
+"""
```
aaronayres35, line 15: When building the docs, these "- :attr:`~.MenuBarManager`" lines just show up as 

- MenuBarManager

Without any link.  Not a suggested change now, but maybe as a future improvement, for toolkit specific things we could link to both implementations somehow?  e.g. one link to docs for https://github.com/enthought/pyface/blob/master/pyface/ui/qt4/action/menu_bar_manager.py and another for wx

pyface/tasks/action/listening_action.py
```diff
@@ -7,6 +7,9 @@
 # is also available online at http://www.enthought.com/licenses/BSD.txt
 #
 # Thanks for using Enthought open source!
-# Backward compatibility import
+
+""" This module exists purely for backwards compatibility.
+Please use pyface.action.listening_action.ListeningAction instead.
```
aaronayres35, line 12: could this be a link as well, rather than just the name

aaronayres35 (approved):
LGTM, I built the docs and everything looked good. Just a couple comments

pyface/action/api.py
```diff
@@ -9,6 +9,68 @@
...
+-----------------
+
+- :class:`~.ActionController`
+
+Action Event
+------------
+
+- :class:`~.ActionEvent`
+
+Action Manager and subclasses
```
kitchoi, line 36: Perhaps the mention of "subclasses" can be removed?  This documentation can stay relatively high-level, the MROs are details.  The API documentation for each of these classes will mention what they are subclassing.  Then you would not need to add the note about ``StatusBarManager``. It feels a bit out of place.

Even better would be section titles that hint what these classes are useful for.

pyface/action/menu_bar_manager.py
kitchoi: > Without any link.

There is already an issue: https://github.com/enthought/pyface/issues/755
Note that Qt and Wx implementations don't necessarily have exactly the same definition. Cherry-picking the ones I know:
The enum definition for DataViewWidget looks like this in Qt:
https://github.com/enthought/pyface/blob/3ea7832cfaa076eacec2b2ccbab07589fe599893/pyface/ui/qt4/data_view/data_view_widget.py#L100-L104

But it looks like this in Wx:
https://github.com/enthought/pyface/blob/3ea7832cfaa076eacec2b2ccbab07589fe599893/pyface/ui/wx/data_view/data_view_widget.py#L42-L46

We can't merge the two into one without having to pick one over another.
Another hard problem is the dependency on the toolkit while generating the doc. Some mocking (in doc config) may be needed, but that is not necessarily trivial either.

pyface/tasks/action/api.py
```diff
@@ -8,6 +8,57 @@
 #
 # Thanks for using Enthought open source!
 
+"""
+
+API for the ``pyface.tasks.action`` submodule.
+
+ActionSchema, subclasses and aliases
```
kitchoi, line 15: Same here, I'd skip the emphasis on the MRO.

pyface/action/api.py
```diff
@@ -9,6 +9,68 @@
...
+-----------------------------
+
+- :class:`~.ActionManager`
+
+Toolkit-specific subclasses:
```
kitchoi, line 41: Not sure my understanding of "toolkit-specific" is the same as what you are intending.
If a class is toolkit-specific, then I need to know which toolkit it is designed for, e.g. MenuManager may only support Qt and I need to know that. But I think these classes are aiming at a toolkit independent API. And ToolPaletteManager only has a Wx implementation.

I am guessing you mean these classes depend on one GUI toolkit, but not down to the specific of which one.

pyface/action/api.py
```diff
@@ -9,6 +9,68 @@
...
+  ``ActionManager`` subclass but it is listed here given it's similarily in
+  function to the other subclasses.
+
+Action Manager item and subclasses
+----------------------------------
+
+- :class:`~.ActionManagerItem`
+- :class:`~.ActionItem`
+
+Group and subclasses
```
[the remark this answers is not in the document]
rahulporuri: I'm going with "Layout support" 👍

pyface/action/api.py
rahulporuri: > This documentation can stay relatively high-level, the MROs are details.

Good point. Removed the "... and subclasses" part in the section title and the separation between toolkit-specific implementations vs the base class and the comment about `StatusBarManager`.

> Even better would be section titles that hint what these classes are useful for.

Not entirely sure how best to phrase that other than "Action Managers"

pyface/action/api.py
rahulporuri: > I am guessing you mean these classes depend on one GUI toolkit, but not down to the specific of which one.

Yes. That's what I meant. But you're right - that's a implementation detail and clearly, it shouldn't be mentioned in the api! Removed this implementation detail.

kitchoi (approved):
LGTM. Thank you!

pyface/action/api.py
```diff
@@ -9,6 +9,63 @@
...
+Action Managers
+---------------
+
+- :class:`~.ActionManager`
+- ``MenuManager``
+- ``MenuBarManager``
+- ``StatusBarManager``
+- ``ToolBarManager``
+- ``ToolPaletteManager``, only for the Wx toolkit.
+
```
kitchoi, line 45: Hmm, these don't have `:class:`, should they have that prefix?

kitchoi: Scratch that. These are the ones that don't have links.

PR #844, opened 2020-12-10
Improve api documentation in pyface.tasks

rahulporuri:
This PR improves documentation in the `pyface.tasks` submodule in the following ways

- adds a module docstring to `pyface.tasks.api`
- adds a module docstring to the modules that import toolkit-specific classes/implementations
- updates trait comments so that they get picked up by sphinx for api docs.

### DIFF
--- a/pyface/tasks/advanced_editor_area_pane.py
+++ b/pyface/tasks/advanced_editor_area_pane.py
@@ -8,7 +8,10 @@
 #
 # Thanks for using Enthought open source!
 
-""" Toolkit-specific implementation of the ``AdvancedEditorAreaPane``. """
+""" Toolkit-specific implementation of the ``AdvancedEditorAreaPane``.
+
+- :attr:`~.AdvancedEditorAreaPane`
+"""
 
 from pyface.toolkit import toolkit_object
 
--- a/pyface/tasks/advanced_editor_area_pane.py
+++ b/pyface/tasks/advanced_editor_area_pane.py
@@ -8,8 +8,7 @@
 #
 # Thanks for using Enthought open source!
 
-""" Toolkit-specific implementation of the ``AdvancedEditorAreaPane``.
-
+"""
 - :attr:`~.AdvancedEditorAreaPane`
 """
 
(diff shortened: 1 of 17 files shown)


pyface/tasks/advanced_editor_area_pane.py
```diff
@@ -7,7 +7,12 @@
 # is also available online at http://www.enthought.com/licenses/BSD.txt
 #
 # Thanks for using Enthought open source!
-# Import the toolkit specific version.
+
+""" Toolkit-specific implementation of the ``AdvancedEditorAreaPane``.
```
kitchoi, line 11: Usually when "toolkit-specific" is mentioned, I expect a specific toolkit being associated with it, e.g. Foo is a Qt-specific implementation of Bar.

The original comment is fine as in it is importing the Qt-specific or the Wx-specific AdvancedEditorAreaPane, but it is commenting about the implementation details. The public API of this module should not need to worry about that. AdvancedEditorAreaPane itself should have a toolkit independent API. So the sentence reads erroneously alarming to me: It is toolkit-specific, so I need to be using a particular toolkit in order to use this class, what is that toolkit?

Perhaps we can just remove this sentence unless we have a better replacement for it?

(Likewise for the other occurrences.)

pyface/tasks/i_editor_area_pane.py
```diff
@@ -42,22 +42,22 @@ class IEditorAreaPane(ITaskPane):
...
     editors = List(IEditor)
 
-    # A list of extensions for file types to accept via drag and drop.
-    # Note: This functionality is provided because it is very common, but drag
-    # and drop support is in general highly toolkit-specific. If more
-    # sophisticated support is required, subclass an editor area implementation.
+    #: A list of extensions for file types to accept via drag and drop.
+    #: Note: This functionality is provided because it is very common, but drag
+    #: and drop support is in general highly toolkit-specific. If more
+    #: sophisticated support is required, subclass an editor area implementation.
```
kitchoi, line 54: (Feel free to ignore, but I wonder if this makes the line too long for the nonexisting flake8...)

aaronayres35: I confirmed it does trigger flake8 (some other changes in this PR do as well), but there are many other flake8 errors in the files in this PR not related to this PR.

aaronayres35 (approved):
LGTM as well aside from Kit's comments.

PR #846, opened 2020-12-10
Address a few sphinx warnings

rahulporuri:
Not sure if this is the right fix but this is a fix. This PR addresses the following sphinx warnings : 

```
pyface\tasks\tasks_application.py:docstring of pyface.tasks.tasks_application.TaskFactory.factory:3: WARNING: Inline strong start-string without end-string.
pyface\timer\do_later.py:docstring of pyface.timer.do_later.do_later:5: WARNING: Inline strong start-string without end-string.
pyface\timer\do_later.py:docstring of pyface.timer.do_later.do_later:6: WARNING: Inline strong start-string without end-string.
pyface\timer\do_later.py:docstring of pyface.timer.do_later.do_after:7: WARNING: Inline strong start-string without end-string.
pyface\timer\do_later.py:docstring of pyface.timer.do_later.do_after:8: WARNING: Inline strong start-string without end-string.
```

Interpreting the "Inline strong start-string" as `"**"`, the fix I employed was to simply escape the characters i.e. `"\**"`

### DIFF
--- a/pyface/timer/do_later.py
+++ b/pyface/timer/do_later.py
@@ -55,7 +55,10 @@ def do_after(interval, callable, *args, **kwargs):
         The time interval in milliseconds to wait before calling.
     callable : callable
         The callable to call.
-    *args, \**kwargs : tuple, dict
+    \*args
+        Positional arguments to be passed through to the callable.
+    \*\*kwargs
+        Keyword arguments to be passed through to the callable.
         Arguments to be passed through to the callable.
     """
     return CallbackTimer.single_shot(
--- a/pyface/timer/do_later.py
+++ b/pyface/timer/do_later.py
@@ -38,7 +38,7 @@ def do_later(callable, *args, **kwargs):
     ----------
     callable : callable
         The callable to call in 50ms time.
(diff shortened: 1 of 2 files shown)


pyface/timer/do_later.py
```diff
@@ -55,7 +55,7 @@ def do_after(interval, callable, *args, **kwargs):
         The time interval in milliseconds to wait before calling.
     callable : callable
         The callable to call.
-    *args, **kwargs : tuple, dict
+    *args, \**kwargs : tuple, dict
```
kitchoi, line 58: While this got rid of the warning, the rendered parameters still look odd (but they were already looking odd before this PR).
Screenshot 2020-12-15 at 17 44 51

Looks like we need to split `*args` and `**kwargs` into its own line... something like

```suggestion
    \*args
        Positional arguments to be passed through to the callable.
    \*\*kwargs
        Keyword arguments to be passed through to the callable.
```

kitchoi (approved):
Confirmed the warnings in `timer` package are gone. The rendered doc still looks a bit odd but this PR has not made it worse.

PR #850, opened 2020-12-15
Declare python_requires to support Python >= 3.6

### LINKED ISSUE #845: Declare Python 3.5 unsupported
Python 3.5 has passed its end-of-life. Traits has dropped its support https://github.com/enthought/traits/pull/1312 (has not been released as of today. Traits 6.1.1 still nominally supports Python 3.5).

However CI against Python 3.5 has been removed on Pyface (https://github.com/enthought/pyface/pull/470), so it is harder to claim Pyface still supports Python 3.5.

This issue is for officially declare Pyface stops supporting Python 3.5, which includes updating `setup.py`'s `classifiers` metadata, and possibly the `python_requires` metadata.  This will make package installer avoid the next release of pyface and fall back to the latest release that does not have this restriction.

kitchoi:
Closes #845

This PR updates `setup.py` such that package installer will avoid the next release if the Python version is less than 3.6.

Pyface no longer supports Python 2, and CI for Python 3.5 has been removed so it would be difficult to continue its support, Python 3.5 has also reached its end-of-life.

Similar to https://github.com/enthought/traitsui/pull/1436

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -50,7 +50,7 @@ using::
 
     python etstool.py test-all
 
-Currently supported runtime values are ``2.7`` and ``3.5``, and currently
+Currently supported runtime values include ``3.6``, and currently
 supported toolkits are ``null``, ``pyqt``, ``pyside`` and ``wx``.  Not all
 combinations of toolkits and runtimes will work, but the tasks will fail with
 a clear error if that is the case.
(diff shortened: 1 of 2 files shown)


aaronayres35 (approved):
LGTM

PR #851, opened 2020-12-15
Use viewcode sphinx extension

rahulporuri:
Using this extension means that API docs now link to source code. See example below which now includes the source button.

sphinx-viewcode


aaronayres35 (approved):
LGTM!  I tested building the docs and it works as expected

PR #853, opened 2020-12-16
Fix importing QtWebEngine after toolkit selection

### LINKED ISSUE #581: Test failures with PyQt5 5.15.0 from PyPI
This is probably not something we can fix but still something we can probably pin to the repository to notify users. As far as I know, this is an issue with the PyQt5 PyPI wheel.

```
ERROR: test_imports (pyface.ui.qt4.tests.test_qt_imports.TestPyfaceQtImports)
  File "C:\Users\rporuri\work\github\ets\pyface\pyface\qt\QtWebKit.py", line 19, in <module>
    from PyQt5.QtWebEngine import *
ModuleNotFoundError: No module named 'PyQt5.QtWebEngine'

During handling of the above exception, another exception occurred:

  File "C:\Users\rporuri\work\github\ets\pyface\pyface\ui\qt4\tests\test_qt_imports.py", line 24, in test_imports
    import pyface.qt.QtWebKit
  File "C:\Users\rporuri\work\github\ets\pyface\pyface\qt\QtWebKit.py", line 28, in <module>
    from PyQt5.QtWebKit import *
ModuleNotFoundError: No module named 'PyQt5.QtWebKit'
```

kitchoi:
Fix #581

First of all, we need a separate distribution for QtWebEngine with PyQt5. This PR expands the `extras_require` so that `pip install pyface[pyqt5]` brings in `PyQtWebEngine`.

After installing `PyQtWebEngine`, importing `QtWebEngineWidgets` after instantiating a QCoreApplication results in an ImportError.  This PR fixes this by trying to import QtWebEngineWidgets before instantiating the QApplication.

All of this has to do with import side effect so it is hard to test. If the test mentioned in #581 is run on its own, it passes on master:
```
$ python -m unittest pyface.ui.qt4.tests.test_qt_imports.TestPyfaceQtImports
.
Ran 1 test in 0.213s

OK
```
But if it is run after some other tests that have instantiated the QApplication, it fails (on master), as in #581:
```
$ python -m unittest pyface.ui.qt4.tests.test_gui pyface.ui.qt4.tests.test_qt_imports.TestPyfaceQtImports
ERROR: test_imports (pyface.ui.qt4.tests.test_qt_imports.TestPyfaceQtImports)
  File "/Users/kchoi/Work/ETS/pyface/pyface/qt/QtWebKit.py", line 20, in <module>
    from PyQt5.QtWebEngineWidgets import (
ImportError: QtWebEngineWidgets must be imported before a QCoreApplication instance is created

During handling of the above exception, another exception occurred:

  File "/Users/kchoi/Work/ETS/pyface/pyface/ui/qt4/tests/test_qt_imports.py", line 24, in test_imports
    import pyface.qt.QtWebKit
  File "/Users/kchoi/Work/ETS/pyface/pyface/qt/QtWebKit.py", line 28, in <module>
    from PyQt5.QtWebKit import *
ModuleNotFoundError: No module named 'PyQt5.QtWebKit'

Ran 2 tests in 0.062s

FAILED (errors=1)
```
With this PR, the second test command also passes.

### DIFF
--- a/pyface/__init__.py
+++ b/pyface/__init__.py
@@ -22,7 +22,7 @@ __requires__ = ["importlib-metadata", "importlib-resources>=1.1.0", "traits>=6"]
 __extras_require__ = {
     "wx": ["wxpython>=4", "numpy"],
     "pyqt": ["pyqt>=4.10", "pygments"],
-    "pyqt5": ["pyqt5", "pygments"],
+    "pyqt5": ["pyqt5", "PyQtWebEngine", "pygments"],
     "pyside": ["pyside>=1.2", "pygments"],
     "pyside2": ["pyside2", "shiboken2", "pygments"],
     "test": ["packaging"],
--- a/pyface/__init__.py
+++ b/pyface/__init__.py
@@ -22,7 +22,7 @@ __requires__ = ["importlib-metadata", "importlib-resources>=1.1.0", "traits>=6"]
 __extras_require__ = {
     "wx": ["wxpython>=4", "numpy"],
     "pyqt": ["pyqt>=4.10", "pygments"],
-    "pyqt5": ["pyqt5", "PyQtWebEngine", "pygments"],
+    "pyqt5": ["pyqt5", "pygments"],
     "pyside": ["pyside>=1.2", "pygments"],
     "pyside2": ["pyside2", "shiboken2", "pygments"],
     "test": ["packaging"],
--- a/pyface/ui/qt4/init.py
+++ b/pyface/ui/qt4/init.py
@@ -38,8 +38,8 @@ if _app is None:
         # pyface.qt.QtWebKit tries QtWebEngineWidgets first, but
         # if QtWebEngineWidgets is present, it must be imported prior to
         # creating a QCoreApplication instance, otherwise importing
-        # QtWebEngineWidgets later would fail. Import it here first before
-        # creating the instance.
+        # QtWebEngineWidgets later would fail (see enthought/pyface#581).
+        # Import it here first before creating the instance.
         from pyface.qt import QtWebKit
     except ImportError:
         # This error will be raised in the context where
--- a/pyface/ui/qt4/init.py
+++ b/pyface/ui/qt4/init.py
@@ -34,6 +34,17 @@ if qt_api == "pyqt":
 _app = QtGui.QApplication.instance()
 
 if _app is None:
+    try:
+        # pyface.qt.QtWebKit tries QtWebEngineWidgets first, but
+        # if QtWebEngineWidgets is present, it must be imported prior to
+        # creating a QCoreApplication instance, otherwise importing
+        # QtWebEngineWidgets later would fail. Import it here first before
+        # creating the instance.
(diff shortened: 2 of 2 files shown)


kitchoi:
Note that CI passing does not verify the PR has fixed the issue, because CI is installing PyQt5 from EDM with a version that includes `QtWebKit` instead of `QtWebEngine`.  The issue has to be verified with an installation using PyQt5/PyQtWebEngine  (from PyPI for example).

pyface/__init__.py
```diff
@@ -22,7 +22,7 @@
 __extras_require__ = {
     "wx": ["wxpython>=4", "numpy"],
     "pyqt": ["pyqt>=4.10", "pygments"],
-    "pyqt5": ["pyqt5", "pygments"],
+    "pyqt5": ["pyqt5", "PyQtWebEngine", "pygments"],
```
corranwebster, line 25: This is only needed for `pyqt5` >= 5.12, so probably should have that encoded in the PyQt version.

kitchoi: Did you mean `"pyqt5": ["pyqt5>=5.12", "PyQtWebEngine", "pygments"]`?

kitchoi: Or I could leave this `"PyQtWebEngine"` out but instead document this additional requirement somewhere.

corranwebster: I was thinking the first of these, unless there is a potential problem with that.

kitchoi: You did make me become aware that PyQtWebEngine is only distributed on PyPI for PyQt5>=5.12. But did you mean that it is only _available_ for pyqt5>=5.12, rather than _needed_?

Specifying "pyqt5>=5.12" would to me means that there are reasons within Pyface that causes pyqt5<5.12 not to be supported, is that true?

So far the ">=5.12" bit seems to me a requirement of PyQtWebEngine, not a requirement of Pyface. If there is a distribution of PyQtWebEngine that would support PyQt5 5.11, I am not sure Pyface should disallow it (I don't think Pyface is setting these minimum requirements based on what are tested on CI - because if it does it would be very restrictive.)

I think only these two options currently make sense to me:
(1) `["pyqt5", "pygments"]`
(2) `["pyqt5", "PyQtWebEngine", "pygments"]`

In any case, these are extras requirements, so they are already optional - they may even be ignored by some installers (e.g. edm).

Perhaps I should revert the change to "extras_require" so we can focus on the changes in the import order which fixes the import error in #581, the error occurs even if one already has the required Qt web dependencies installed.

kitchoi:
I have reverted the change to `extras_require` for now. It can be added again if it becomes necessary.

pyface/ui/qt4/init.py
```diff
@@ -34,6 +34,17 @@
 _app = QtGui.QApplication.instance()
 
 if _app is None:
+    try:
```
rahulporuri, line 37: It might be good to leave a link to the underlying issue here - https://github.com/enthought/pyface/issues/581 - for future reference.

rahulporuri (approved):
LGTM

PR #854, opened 2021-01-06
Update year range in copyright headers

rahulporuri:
It's 2021 so the year range should be X-2021, not X-2020. The changes were done using automated regex, similar to what was done in the PR https://github.com/enthought/envisage/pull/392.  Regex search term - `Copyright (\d{4})-2020` - replaced with `Copyright $1-2021`. Done on VS Code.


aaronayres35 (approved):
LGTM

PR #855, opened 2021-01-08
Remove coverage reporting on PRs

rahulporuri:
This PR removes coverage reporting for Pull Requests. The coverage reports have been visual noise at best in recent times and have not really contributed meaningful information in the PRs.

Apart from these code changes, i don't know if there are other changes that we need to make to completely sever ties with codecov.


rahulporuri:
Looks like there is a webbook that I will need to delete after this PR is merged 
codecov-webhook

aaronayres35 (approved):
LGTM

We still want to keep coverage around, just not the codecov reporting correct?  
That is why we left it as a dependency in etstool.py and it is still used when running tests etc.

rahulporuri:
> We still want to keep coverage around, just not the codecov reporting correct?
That is why we left it as a dependency in etstool.py and it is still used when running tests etc.

Yup. We still want to be able to see coverage report locally if we want to - we're only removing the automated coverage report on github PRs.

PR #860, opened 2021-01-19
No longer need to do Python 2 configuration for sip

### LINKED ISSUE #859: Remove prepare_pyqt4 from pyface.qt
The code to handle setting `sip` APIs is no longer needed in Python 3 code:
https://github.com/enthought/pyface/blob/f2b2639187688659d291117e1733cc428d0dcfa6/pyface/qt/__init__.py#L24-L27

This function should be removed, which will improve compatibility with other Qt-based code.

corranwebster:
Fixes #859 

The code removed should effectively be a no-op on Python 3, since it was configuring `sip` to behave in a Python 3-compatible way.

### DIFF
--- a/pyface/qt/__init__.py
+++ b/pyface/qt/__init__.py
@@ -21,37 +21,6 @@ QtAPIs = [
 ]
 
 
-def prepare_pyqt4():
-    """ Set PySide compatible APIs. """
-    # This is not needed for Python 3 and can be removed when we no longer
-    # support Python 2.
-    try:
-        # required for PyQt >= 4.12.2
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM but IIUC, this already happens with PyQt4 + Python 3 under the hood in `sip`. Is that correct?

corranwebster:
Yep.  This is code that changes the behaviour os `sip` under PyQt4 + Python 2 so that it behaves like PyQt4 + Python 3 (and also PySide).  So under Python 3 this is just setting the values to what they should already be.

PR #861, opened 2021-01-19
Remove support for PySide

corranwebster:
This does part but not all of #510 - this removes support for PySide, but not PyQt4.

Traits no longer supports any version of Python compatible with the original PySide, so this has been dead code for a while.

### DIFF
--- a/pyface/qt/QtScript.py
+++ b/pyface/qt/QtScript.py
@@ -12,15 +12,7 @@ from . import qt_api
 if qt_api == "pyqt":
     from PyQt4.QtScript import *
 
-elif qt_api == "pyqt5":
-    import warnings
-
-    warnings.warn(DeprecationWarning("QtScript is not supported in PyQt5"))
-
-elif qt_api == "pyside2":
-    import warnings
-
-    warnings.warn(DeprecationWarning("QtScript is not supported in PyQt5"))
-
 else:
-    from PySide.QtScript import *
+    import warnings
+
+    warnings.warn(DeprecationWarning("QtScript is not supported in PyQt5"))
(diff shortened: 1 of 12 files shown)


pyface/qt/QtScript.py
```diff
@@ -12,15 +12,7 @@
...
-elif qt_api == "pyqt5":
-    import warnings
-
-    warnings.warn(DeprecationWarning("QtScript is not supported in PyQt5"))
-
-elif qt_api == "pyside2":
+else:
     import warnings
 
     warnings.warn(DeprecationWarning("QtScript is not supported in PyQt5"))
```
rahulporuri, line 18:
```suggestion
    warnings.warn(DeprecationWarning("QtScript is not supported in PyQt5/PySide2"))
```

rahulporuri (approved):
Code changes LGTM but I have a couple of orthogonal questions.

> Traits no longer supports any version of Python compatible with the original PySide, so this has been dead code for a while.

I'm not entirely sure what you mean by this. It looks like PySide officially supported Python <= 3.4 (https://pypi.org/project/PySide/#files) but I suspect we can still use PySide on Python 3.6.

Also, do we want this merged before Pyface 7.2.0? We are already planning to drop support for Traits < 6.2.0 in the upcoming Pyface release. The next pyface release is planned for early-to-mid February.

corranwebster:
> I'm not entirely sure what you mean by this. It looks like PySide officially supported Python <= 3.4 (https://pypi.org/project/PySide/#files) but I suspect we can still use PySide on Python 3.6.

The last version of Python that can be used with the original PySide is Python 3.4.  PySide2 is the version for compatibility with Python >= 3.5  - there is no way to get original PySide on Python 3.6+

corranwebster:
> Also, do we want this merged before Pyface 7.2.0?

It should probably be in the next release, yes.

PR #862, opened 2021-01-21
Expose keyboard focus on Widgets

corranwebster:
This PR adds two methods to the base Widget class to set the keyboard focus and to check whether the widget currently has keyboard focus.

This is needed to make it easier to create toolkit-independent TraitsUI Editors that use Pyface widgets.

The tests are skipped on Linux, because there keyboard focus is global, and so a widget doesn't get focus unless the window also has focus/is active, which is hard to guarantee, and is disruptive to development as it prevents the tests from running in the background.

I will open an issue to track the testing of focus on linux.

### DIFF
--- a/pyface/tests/test_widget.py
+++ b/pyface/tests/test_widget.py
@@ -34,11 +34,12 @@ class ConcreteWidget(Widget):
             control.Show(self.visible)
         elif toolkit_object.toolkit in {"qt4", "qt"}:
             from pyface.qt import QtGui
+            from pyface.qt.QtCore import Qt
 
             control = QtGui.QWidget(parent)
             control.setEnabled(self.enabled)
             control.setVisible(self.visible)
-            control.setFocusPolicy(QtGui.StrongFocus)
+            control.setFocusPolicy(Qt.StrongFocus)
         else:
             control = None
         return control
--- a/pyface/tests/test_widget.py
+++ b/pyface/tests/test_widget.py
@@ -315,3 +315,12 @@ class TestConcreteWidget(unittest.TestCase, GuiTestAssistant):
                 self.widget.enabled = False
 
         self.assertFalse(self.widget.control.isEnabled())
+
+    def test_focus(self):
+        with self.event_loop():
+            self.widget.create()
+
+        with self.event_loop():
+            self.widget.focus()
+
+        self.assertTrue(self.widget.control.has_focus())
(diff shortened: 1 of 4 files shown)


rahulporuri:
there seems to be one failure on linux - 

```
FAIL: test_focus (pyface.tests.test_widget.TestConcreteWidget)
  File "/home/travis/.edm/envs/pyface-test-3.6-pyqt/lib/python3.6/site-packages/pyface/tests/test_widget.py", line 326, in test_focus
    self.assertTrue(self.widget.has_focus())
```

rahulporuri (commented):
I'm not sure why a Widget wont have control - and i'm not sure what the behavior should be if `control` is None. At the moment, the methods do nothing if `control` is None - should that be part of the public API (and documented in the docstrings)

corranwebster:
The `control` can be `None` either before the `create` method is called, or after the `destroy` method is called.  These methods should never be called when the widget is in either of these states, but there is the possibility of it happening eg. in a delayed trait handler that gets called after the widget has been destroyed.

In both cases the control _can't_ have focus, so it is OK to return `False` unilaterally from `has_focus` (ie. it correctly represents the situation).

Generally guarding toolkit calls with a check that the control exists is standard across most of these sorts of methods, so I don't think it needs to be called out specifically for the focus methods.  In particular, we don't really want people relying on this behaviour.

pyface/tests/test_widget.py
```diff
@@ -315,3 +319,16 @@ def test_enabled(self):
...
+    @unittest.skipIf(
+        is_linux,
+        "Linux keyboard focus is False unless window is active",
+    )
+    def test_focus(self):
+        with self.event_loop():
+            self.widget.create()
+
+        with self.event_loop():
+            self.widget.focus()
```
rahulporuri, line 332:
```suggestion
        with self.event_loop():
            self.widget.create()

        self.assertFalse(self.widget.has_focus())

        with self.event_loop():
            self.widget.focus()
```

rahulporuri (approved):
Mostly LGTM with one minor suggestion

pyface/tests/test_widget.py
corranwebster: I was cautious about this, because that may depend on OS behaviour (the OS may automatically put keyboard focus in the first widget in the active window), but I'm willing to give it a shot and see if it works.

PR #865, opened 2021-01-22
Stop running CI on traits 6.0

### LINKED ISSUE #781: Stop doing CI runs with Traits 6.0

aaronayres35:
fixes #781 

This PR updates the travis.yml file to require traits 6.1+ not 6.0+.  It also removes a traits version check that prevented data_view test from being run for traits !=6.1.  Not we will want these tests to run on 6.2+ as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ env:
 matrix:
   include:
     - env: RUNTIME=3.6 TOOLKITS="pyqt pyqt5 pyside2 wx"
-    - env: RUNTIME=3.6 TOOLKITS="pyqt5 wx" TRAITS_REQUIRES="^=6.0"
+    - env: RUNTIME=3.6 TOOLKITS="pyqt5 wx" TRAITS_REQUIRES="^=6.1"
   fast_finish: true
 
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,6 @@ env:
 matrix:
   include:
     - env: RUNTIME=3.6 TOOLKITS="pyqt pyqt5 pyside2 wx"
-    - env: RUNTIME=3.6 TOOLKITS="pyqt5 wx" TRAITS_REQUIRES="^=6.1"
   fast_finish: true
 
 cache:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -27,7 +27,7 @@ env:
 matrix:
   include:
     - env: RUNTIME=3.6 TOOLKITS="pyqt pyqt5 pyside2 wx"
-    - env: RUNTIME=3.6 TOOLKITS="pyqt5 wx" TRAITS_REQUIRES="^=6.0"
+    - env: RUNTIME=3.6 TOOLKITS="pyqt5 wx" TRAITS_REQUIRES="^=6.1"
```
rahulporuri, line 30: Not sure why we need to test against both the latest version of traits (6.2) and the older version (6.1) for the pyqt5 and wx toolkits. I think we can remove the env where we set `TRAITS_REQUIRES`

rahulporuri (changes requested):
Apart from these two changes, we will also want to specify in `setup.py` that we require `traits>=6.2.0`.

rahulporuri (approved):
I think this LGTM.

The fishy thing (that i suspect but have not verified) is that traits 6.1.1 is probably installed via edm first - and then updated to traits 6.2.0 via pip because of the change in version requirement. Can you confirm this?

If that is indeed the case, I would hold off on merging this PR until traits 6.2.0 is available via edm - and rerun CI on this PR, just to be super safe (and to calm my super paranoia).

aaronayres35:
> I think this LGTM.
> 
> The fishy thing (that i suspect but have not verified) is that traits 6.1.1 is probably installed via edm first - and then updated to traits 6.2.0 via pip because of the change in version requirement. Can you confirm this?
> 
> If that is indeed the case, I would hold off on merging this PR until traits 6.2.0 is available via edm - and rerun CI on this PR, just to be super safe (and to calm my super paranoia).

It looks like CI is actually just being run with traits 6.1 from edm currently, even though the setup.py explicitly requires traits 6.2+.  There are no explicit version restrictions on the CI config files as the PR currently stands.  I will wait until 6.2 is available through edm and rerun CI on this PR

rahulporuri:
Rerunning CI now that Travis 6.2.0 is available via EDM.

rahulporuri:
Looks like traits 6.2.0 is now being used in the CI. Still LGTM.

PR #881, opened 2021-02-05
Replace depends_on with observe in properties

aaronayres35:
This PR is part of the transition to depending on traits 6.2.  This PR replaces all use of `depends_on` with `observe` in property definitions throughout the code base.  I did so via VS code find an replace, but manually went through to confirm the changes/updated a couple occurrences where we observed containers and needed to use the updated traits mini language.

### DIFF
--- a/pyface/action/gui_application_action.py
+++ b/pyface/action/gui_application_action.py
@@ -28,7 +28,7 @@ class GUIApplicationAction(ListeningAction):
 
     # 'ListeningAction' interface --------------------------------------------
 
-    object = Property(depends_on="application")
+    object = Property(observe="application")
 
     # 'WindowAction' interface -----------------------------------------------
 
@@ -53,7 +53,7 @@ class ActiveWindowAction(GUIApplicationAction):
 
     # 'ListeningAction' interface --------------------------------------------
 
-    object = Property(depends_on="application.active_window")
+    object = Property(observe="application.active_window")
 
     # ------------------------------------------------------------------------
     # Protected interface.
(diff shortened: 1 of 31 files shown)


pyface/action/gui_application_action.py
```diff
@@ -28,7 +28,7 @@ class GUIApplicationAction(ListeningAction):
 
     # 'ListeningAction' interface --------------------------------------------
 
-    object = Property(depends_on="application")
+    object = Property(observe="application")
```
aaronayres35, line 31: flag (see other comment)

pyface/action/listening_action.py
```diff
@@ -125,7 +125,7 @@ def _object_changed(self, old, new):
         for kind in ("enabled", "visible"):
             method = getattr(self, "_%s_update" % kind)
             name = getattr(self, "%s_name" % kind)
-            if old:
+            if old and old is not Undefined:
```
aaronayres35, line 128: This change was a bit strange, but it is necessary because of https://github.com/enthought/pyface/pull/881/files#r571222562 and here:https://github.com/enthought/traits/blob/aece1feca33e51890f9d0054606a7a241cb2776e/traits/has_traits.py#L327 I believe.

This also made me think that maybe `Undefined` should not evaluate to true in if statements like this, but I do not have enough context on its use to say this at all confidently (it just seems like it would be intuitive).

rahulporuri: it's worth bringing this up on the channel and hearing from the rest of the team.

rahulporuri (approved):
LGTM

pyface/action/listening_action.py
corranwebster: Suggestion from side-discussion: perhaps use `if isinstance(old, HasTraits)` (and the equivalent for `new` below).

PR #882, opened 2021-02-05
Remove unneeded setting of comparison_mode

aaronayres35:
As of traits 6.2 it is no longer necessary to specify a trait comparison mode of `ComparisonMode.identity`.  This PR simply removes a couple of occurrences of that from examples in the code base.


rahulporuri (approved):
LGTM.

PR #885, opened 2021-02-06
Fix undo/redo merge bugs

### LINKED ISSUE #883: Pushing mergeable Commands after undo doesn't delete rest of stack
Given a `CommandStack` in normal operation (ie. not building a macro) where you are doing a sequence of steps of the form:

- `push` some commands
- `undo` one or more commands
- `push` a command

then all commands after the current command are deleted from the stack here:
https://github.com/enthought/pyface/blob/05eefd485fa3831d39fdf4156f24bdc0d8e5140f/pyface/undo/command_stack.py#L211

However if the command which is pushed happens to be mergeable, then the remainder of the stack is not cleared.  It should be.

For comparison, the `UndoHandler` of TraitsUI does do this correctly.

### LINKED ISSUE #884: Pushing mergeable Commands at clean point shouldn't merge
Given a `CommandStack` in normal operation (ie. not building a macro) where you are doing a sequence of steps of the form:

- `push` a command (or otherwise have the current command be mergeable)
- mark the stack as `clean` at that point (eg. user saves at that point)
- `push` a command that can be merged with the current command

This should not be merged, as then the "clean" state would be modified which might prevent operations that are contingent on being in a "dirty" state (eg. "Save" a document).

This is not state that the `Command` can see, so it can't change its behaviour based on this, so the `CommandStack` is the right place to handle it.

corranwebster:
Add tests for undo/redo and fix #883 and fix #884 

Not much more.

### DIFF
--- a/pyface/undo/command_stack.py
+++ b/pyface/undo/command_stack.py
@@ -185,7 +185,7 @@ class CommandStack(HasTraits):
 
         # See if the command can be merged with the previous one.
         if len(self._macro_stack) == 0:
-            if self._index >= 0:
+            if self._index >= 0 and not self._stack[self._index].clean:
                 merged = self._stack[self._index].command.merge(command)
             else:
                 merged = False
@@ -201,6 +201,9 @@ class CommandStack(HasTraits):
 
         # Do nothing more if the command was merged.
         if merged:
+            if len(self._macro_stack) == 0:
+                del self._stack[self._index+1:]
+            self.undo_manager.stack_updated = self
             return result
 
         # Only update the command stack if there is no current macro.
--- a/pyface/undo/command_stack.py
+++ b/pyface/undo/command_stack.py
@@ -199,9 +199,11 @@ class CommandStack(HasTraits):
         # Execute the command.
         result = command.do()
 
-        # Do nothing more if the command was merged.
+        # Update the stack state for a merged command.
         if merged:
             if len(self._macro_stack) == 0:
+                # If not in macro mode, remove everything after the current
+                # command from the stack.
                 del self._stack[self._index+1:]
             self.undo_manager.stack_updated = self
             return result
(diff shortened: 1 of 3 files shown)


pyface/undo/command_stack.py
```diff
@@ -201,6 +201,9 @@ def push(self, command):
 
         # Do nothing more if the command was merged.
```
rahulporuri, line 202: this comment needs to be updated - or removed - given the recent changes.

pyface/undo/tests/test_command_stack.py
```diff
@@ -31,21 +34,92 @@ def test_empty_command_stack(self):
...
+
+    def test_n_merge_command_pushed(self):
+        n = 4
+        with self.assert_n_commands_pushed(self.stack, 1):
+            self.command = MergeableCommand()
+            self.stack.push(self.command)
+            for i in range(n):
+                command = MergeableCommand()
+                with self.assertTraitChanges(self.stack.undo_manager,
+                                         'stack_updated', count=1):
```
rahulporuri, line 87: nit-picky suggestion

```suggestion
                with self.assertTraitChanges(self.stack.undo_manager,
                                             'stack_updated', count=1):
```

rahulporuri (approved):
LGTM with two minor comments.

rahulporuri:
Also, given that the 7.3.0 release is imminent, do we need to backport this PR to the 7.2 maintenance branch for a 7.2.1 bug fix release? We could backport the changes purely in principle, in case someone requests a 7.2.1 specifically and can't use 7.3.0.

corranwebster:
If 7.3 is about to come out, then we don't need to do the backport, but maybe keep the flag on the PR in case for some reason we later need a 7.2.1 release and then we can easily work out what needs to be backported.

PR #891, opened 2021-02-10
Set IS_RELEASED back to False

aaronayres35:
[targeting maint/7.3]

This PR simply sets `IS_RELEASED` back to False.


rahulporuri (approved):
LGTM

PR #880, opened 2021-02-04
Convert dynamic on_trait_change to observe part 2

aaronayres35:
closes #637 
closes #732 

This PR makes the last round of updates to replace use of `on_trait_change` with `observe`.  With the changes in this PR, `on_trait_change` is no longer present anywhere in the code base.  This PR also makes an update to the documentation section on `timers`.  Please confirm the changes made there are reasonable.  

Note to reviewer, be cautious: some of the code undergoing changes in this PR is not covered by the test suite.  I made some knowingly broken changes and ran the test suite only to find everything still passing. 

This PR also drive by removes some commented out code from the code base.

PR #892, opened 2021-02-11
don't use trait() for extended trait names

aaronayres35:
see comment: https://github.com/enthought/pyface/pull/880#discussion_r573969099

This will need a backport before 7.3.0 is officially released!

This PR replaces `trait(_____, optional=True)` with just some if statements to ensure we don't try to hook or unhook things which don't exist.


rahulporuri (approved):
LGTM

PR #894, opened 2021-02-22
Fix some doc warnings

aaronayres35:
This PR simply fixes a couple of warnings that occur when building the documentation. One of them was caused by an actual bug in the code (seems like a copy-pasta error)

### DIFF
--- a/pyface/preference/preference_dialog.py
+++ b/pyface/preference/preference_dialog.py
@@ -15,4 +15,4 @@
 
 from pyface.toolkit import toolkit_object
 
-PreferencesDialog = toolkit_object("preference.preference_dialog:PreferenceDialog")
+PreferenceDialog = toolkit_object("preference.preference_dialog:PreferenceDialog")
--- a/pyface/preference/preference_dialog.py
+++ b/pyface/preference/preference_dialog.py
@@ -15,4 +15,4 @@
 
 from pyface.toolkit import toolkit_object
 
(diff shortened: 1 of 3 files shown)


pyface/preference/preference_dialog.py
```diff
@@ -15,4 +15,4 @@
 
 from pyface.toolkit import toolkit_object
 
-TreeViewer = toolkit_object("preference.preference_dialog:PreferenceDialog")
```
rahulporuri, line 18: I'm a bit apprehensive of this change because it changes a public attribute in a public module. Can you check to see if anyone tries importing `TreeViewer` from `pyface.preference.preference_dialog`?

rahulporuri (approved):
LGTM

pyface/preference/preference_dialog.py
aaronayres35: So the API module tries to import `PreferenceDialog` but it doesn't exist, cause it was accidentally named TreeViewer here.  I think it was copy pasted from this:
https://github.com/enthought/pyface/blob/a3d751a2384b75f38572d907eb15ae9c95311a39/pyface/viewer/tree_viewer.py#L18
Other places import `TreeViewer` from here.  No one tries to get `TreeeViewer` from `pyface/preference/preference_dialog.py`

rahulporuri: awesome. still LGTM

PR #900, opened 2021-03-09
listen to correct trait name

aaronayres35:
This PR fixes a bug found via the traitsui cron job, see https://github.com/enthought/traitsui/actions/runs/619678526 

Previously we were trying to listen to a trait `read_only_changed` which didn't exist.  we meant to listen to `read_only`.
The new test failures surfaced from the switch from `on_trait_change` to `observe`.  The code was still broken before (we were trying to listen to a non-existent trait), `on_trait_change` was just quiet about it.

Unfortunately the relevant code does not seem to be under test in `pyface` itself?  I did not see any pyface wx test suite failures, but I used pyface from this branch on the traitsUI wx test suite and saw this resolved the errors.  
The allowed wx failures do let things like this slip by though 😬

### DIFF
--- a/pyface/ui/wx/grid/grid.py
+++ b/pyface/ui/wx/grid/grid.py
@@ -259,7 +259,7 @@ class Grid(Widget):
         self.observe(
             self._on_default_cell_bg_color_changed, "default_cell_bg_color"
         )
-        self.observe(self._on_read_only_changed, "read_only_changed")
+        self.observe(self._on_read_only_changed, "read_only")
         self.observe(self._on_selection_mode_changed, "selection_mode")
         self.observe(
             self._on_column_label_height_changed, "column_label_height"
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM

>  The allowed wx failures do let things like this slip by though 😬

I don't remember how many test failures there are with the wx backend. Do you think it's worth skipping those failures, opening an issue and unskip the testsuite for wx?

aaronayres35:
> I don't remember how many test failures there are with the wx backend. Do you think it's worth skipping those failures, opening an issue and unskip the testsuite for wx? I don't want to spend time making wx work correctly but I don't want us introducing new regressions either.

On the most recently merged PR on traits UI the travis wx job had 11 failures and 3 errors.  I agree. I believe when wx failures were first allowed this number was lower than it is now ... meaning we may have already introduced some regressions along the way (I haven't checked to confirm this though).  I think your proposal make sense.  The alternative of explicitly checking the CI log for new failures / keeping track of which ones already exist on each PR / review is proving hard to remember to do

PR #907, opened 2021-03-22
Allow passing a text format to a MessageDialog

aaronayres35:
Inspired by https://github.com/enthought/traitsui/pull/1546

This PR allows a user to specify the text format they want to use to a message dialog, to help avoid issues like https://github.com/enthought/traitsui/issues/1543

Note, there used to be a `LogText` option, and since qt 5.14 there is a `MarkdownText` option.  I have not added those here to avoid special handling for different Qt versions / because we have not had an explicit use case.  However, if they are needed or wanted, they could be added.

### DIFF
--- a/pyface/message_dialog.py
+++ b/pyface/message_dialog.py
@@ -13,7 +13,12 @@
 
 # Convenience functions.
 def information(
-    parent, message, title="Information", detail="", informative=""
+    parent,
+    message,
+    title="Information",
+    detail="",
+    informative="",
+    text_format="auto"
 ):
     """ Convenience method to show an information message dialog.
 
@@ -39,11 +44,19 @@ def information(
         severity="information",
         detail=detail,
         informative=informative,
+        text_format=text_format,
     )
     dialog.open()
 
 
-def warning(parent, message, title="Warning", detail="", informative=""):
+def warning(
+    parent,
+    message,
+    title="Warning",
+    detail="",
+    informative="",
+    text_format="auto"
+):
     """ Convenience function to show a warning message dialog.
 
     Parameters
@@ -68,11 +81,18 @@ def warning(parent, message, title="Warning", detail="", informative=""):
         severity="warning",
         detail=detail,
         informative=informative,
+        text_format=text_format,
     )
     dialog.open()
 
 
-def error(parent, message, title="Error", detail="", informative=""):
+def error(
+    parent,
+    message,
+    title="Error",
+    detail="",
+    informative="",
+    text_format="auto"):
     """ Convenience function to show an error message dialog.
 
     Parameters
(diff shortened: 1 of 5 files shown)


rahulporuri (commented):
At the moment, the code will be broken with the wx toolkit - can you add the `text_format` trait on the wx `MessageDialog` class as well?

pyface/message_dialog.py
```diff
@@ -30,6 +35,9 @@ def information(
         "Show details").
     informative : str
         Explanatory text to display along with the message.
+    text_format : str
+        Specifies what text format to use in the resulting message dialog.
+        One of "auto", "plain", or "rich".
```
rahulporuri, line 40: nit - it might be worth adding a short comment here - that this is only used by the Qt backend.

rahulporuri (approved):
LGTM

corranwebster:
Some questions/comments:

- this isn't the only place where Qt will assume rich text, do we need to make similar changes elsewhere?
- I think this breaks on Wx if you use the helper functions; either the trait needs to be added to the Wx version of the dialog (possibly with the only allowed options being 'auto' and 'plain') or the argument needs to be removed from the helper functions
- I think there are other helpers for message dialogs as methods in `i_window.py`

aaronayres35:
> * this isn't the only place where Qt will assume rich text, do we need to make similar changes elsewhere?

This is true, and will require a more thorough audit of the codebase.  I am happy to do that and try to tack them on to this PR, or alternatively open a separate catch-all issue for auditing the code base for any such cases and having other separate PRs as needed.

> * I think this breaks on Wx if you use the helper functions; either the trait needs to be added to the Wx version of the dialog (possibly with the only allowed options being 'auto' and 'plain') or the argument needs to be removed from the helper functions

I had added the trait on `pyface.ui.wx.message_dialog.MessageDialog`, that is all that is needed correct?  I can remove the "rich" option there, but then it is awkward because the `pyface.message_dialog` where the helpers are defined will say it is available in the docstrings of the helpers.

> * I think there are other helpers for message dialogs as methods in `i_window.py`

I've updated these now, thank you for pointing those out!  I noticed `pyface.wx.dialog` also has its own helper functions for creating message dialogs but I did not update these (as they don't use `pyface.ui.wx.message_dialog` but rather create their own `wx.MessageDialog` directly.  Tbh they feel fairly redundant and I'm not sure why they are needed.  Perhaps they can be removed or deprecated?

rahulporuri:
> ... alternatively open a separate catch-all issue for auditing the code base for any such cases and having other separate PRs as needed.

let's go with ^ option instead and merge this PR now.

rahulporuri (approved):
Still LGTM

PR #915, opened 2021-03-26
Handle the events correctly

rahulporuri:
This was missed in the `on_trait_change` to `observe` changes introduced in #880 . The methods were previously receiving
`obj` as an argument to the trait change handler which should have been replaced with `event.object` in #880.

This was discovered when fixing `F821` flake8 failures in pyface.


aaronayres35 (approved):
LGTM

Nice catch, thank you!

PR #923, opened 2021-03-27
Fix F841 flake8 errors

rahulporuri:
This PR fixes the F841 flake8 errors - "local variable is assigned but never used".

### DIFF
--- a/pyface/ui/null/action/action_item.py
+++ b/pyface/ui/null/action/action_item.py
@@ -86,14 +86,6 @@ class _Tool(HasTraits):
         self.item = item
         self.tool_bar = tool_bar
 
-        # Create an appropriate tool depending on the style of the action.
-        action = self.item.action
-
-        # If the action has an image then convert it to a bitmap (as required
-        # by the toolbar).
-        if action.image is not None:
-            action.image.create_image()
-
         self.control_id = 1
         self.control = None
         if controller is not None:
--- a/pyface/ui/null/action/action_item.py
+++ b/pyface/ui/null/action/action_item.py
@@ -92,15 +92,7 @@ class _Tool(HasTraits):
         # If the action has an image then convert it to a bitmap (as required
         # by the toolbar).
         if action.image is not None:
-            image = action.image.create_image()
-            path = action.image.absolute_path
-            bmp = image_cache.get_bitmap(path)
-
-        else:
-            from pyface.api import ImageResource
-
-            image = ImageResource("foo")
-            bmp = image.create_bitmap()
+            action.image.create_image()
 
         self.control_id = 1
         self.control = None
--- a/pyface/ui/qt4/clipboard.py
+++ b/pyface/ui/qt4/clipboard.py
@@ -49,6 +49,7 @@ class Clipboard(BaseClipboard):
         mime_data = cb.mimeData()
         if mime_data.hasFormat(PYTHON_TYPE):
             serialized_data = BytesIO(mime_data.data(PYTHON_TYPE).data())
+            _ = load(serialized_data)
             obj = load(serialized_data)
         return obj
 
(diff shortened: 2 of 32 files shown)


pyface/ui/wx/grid/tests/test_simple_grid_model.py
```diff
@@ -56,9 +56,6 @@ def test_get_value(self):
 
     def test_is_cell_empty(self):
 
-        rows = self.model.get_row_count()
-        columns = self.model.get_column_count()
-
         self.assertEqual(
```
aaronayres35, line 59: the second assert below is a hard coded magic number 10, perhaps we should just use rows+1/columns+1 there?

pyface/ui/wx/action/tool_palette_manager.py
```diff
@@ -77,15 +77,15 @@ def create_tool_palette(self, parent, controller=None):
...
             if len(group.items) > 0:
                 # Is a separator required?
                 ## FIXME : Does the palette need the notion of a separator?
-                ##                 if previous_non_empty_group is not None and group.separator:
-                ##                     tool_bar.AddSeparator()
+                ## if previous_non_empty_group is not None and group.separator:
+                ##     tool_bar.AddSeparator()
                 ##
-                ##                 previous_non_empty_group = group
+                ## previous_non_empty_group = group
```
aaronayres35, line 88: why not just delete the commented code?

pyface/ui/null/action/action_item.py
```diff
@@ -92,15 +92,7 @@ def __init__(
...
-            image = action.image.create_image()
-            path = action.image.absolute_path
-            bmp = image_cache.get_bitmap(path)
-
-        else:
-            from pyface.api import ImageResource
-
-            image = ImageResource("foo")
-            bmp = image.create_bitmap()
+            action.image.create_image()
```
aaronayres35, line 95: why do we need to keep this?

pyface/ui/qt4/clipboard.py
```diff
@@ -49,7 +49,7 @@ def _get_object_data(self):
         mime_data = cb.mimeData()
         if mime_data.hasFormat(PYTHON_TYPE):
             serialized_data = BytesIO(mime_data.data(PYTHON_TYPE).data())
-            klass = load(serialized_data)
+            _ = load(serialized_data)
```
aaronayres35, line 52: I don't understand the point of this line given the following line

I assume this line needs to stay though hence the change / not a deletion?

aaronayres35 (approved):
LGTM

Just a few questions /comments.  I searched the relevant files for things being removed and I believe these should be safe

pyface/ui/wx/grid/tests/test_simple_grid_model.py
rahulporuri: fixed. waiting to see if CI is happy.

pyface/ui/qt4/clipboard.py
rahulporuri: it looks like we need to load the data twice - and the first time we call `load`, it returns the `klass` information and the second time we call `load`, we get the actual object from the clipboard. I had originally removed the call entirely - and I saw test failures - which is how I know that we need two calls. See CI failures in the first commit of this PR

rahulporuri: i added comments in the code reflecting the same

pyface/ui/null/action/action_item.py
rahulporuri: I thought there was a reason but when i looked a second time, i couldn't figure out why. It looks like it can be removed and I removed it.

pyface/ui/wx/action/tool_palette_manager.py
rahulporuri: I don't have a very good reason here. It just looked like meaningful code and that there might be some value in keeping it.

PR #926, opened 2021-04-07
Create generic module for GuiTestAssistant and ModalDialogTester

### LINKED ISSUE #905: Create pyface.util.gui_test_assistant
At the moment, the `GuiTestAssistant` is only available with the Qt backend at https://github.com/enthought/pyface/blob/65813858730a61f25618e98669bcc5347d44935a/pyface/ui/qt4/util/gui_test_assistant.py - but there is no generic module which users can import the `GuiTestAssistant` object from.

It'd be useful to create a new `pyface.util.gui_test_assistant` module which imports the actual `GuiTestAssistant` from the relevant toolkit and exposes it for convenience.

aaronayres35:
closes #905 

This PR creates modules for `GuiTestAssistant` and `ModalDialogTester` so that they can more easily be imported via 
`from pyface.util.{gui_test_assistant/modal_dialog_tester} import {GuiTestAssistant/ModalDialogTester}`

It is still important to now check if the imported item is `Unimplemented` as these are both currently only supported for qt.

This PR also adds very simply tests to test importing the objects from the modules

### DIFF
--- a/pyface/util/gui_test_assistant.py
+++ b/pyface/util/gui_test_assistant.py
@@ -1,0 +1,17 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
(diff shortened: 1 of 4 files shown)


rahulporuri (approved):
LGTM

PR #931, opened 2021-04-20
Replace "Either" trait with "Union"

rahulporuri:
This PR replaces the `Either` trait with the `Union` trait.

> A new Union trait type has been added. This is intended as a simpler replacement for the existing Either trait type, which will eventually be deprecated.

See https://github.com/enthought/traits/releases/tag/6.1.0.

See similar PR in traitsui https://github.com/enthought/traitsui/pull/1581

### DIFF
--- a/pyface/tasks/action/schema.py
+++ b/pyface/tasks/action/schema.py
@@ -20,7 +20,6 @@ from pyface.util.id_helper import get_unique_id
 from traits.api import (
     Bool,
     Callable,
-    Either,
     Enum,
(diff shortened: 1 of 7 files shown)


aaronayres35 (approved):
LGTM, but same comment again as from my reviews on https://github.com/enthought/traitsui/pull/1581 and https://github.com/enthought/envisage/pull/405

rahulporuri:
I've reverted a few potentially buggy/untested/unverified changes in the PR.

PR #932, opened 2021-04-20
fix how/when layout calcs are done in wx ImageButton

aaronayres35:
Previously we only did various layout calculations in the `__init__` method.  this ignored the possibility of say starting with `image` as `None` and then setting it later, leading to incorrect layout of things.  See the full saga here:
https://github.com/enthought/traitsui/pull/1566/files#r616858523


rahulporuri (approved):
LGTM

PR #935, opened 2021-04-23
Remove unnecessary try/except around ETSConfig.provisional_toolkit

### LINKED ISSUE #927: Remove unnecessary try/except around ETSConfig.provisional_toolkit
https://github.com/enthought/pyface/blob/004d6ee43acaa51b17e78707f8aeb1118d818725/pyface/base_toolkit.py#L81-L103

aaronayres35:
closes #927 

Note, that later in the file we were already just directly using `ETSConfig.provisional_toolkit` not `provisional_toolkit` (effectively assuming the try portion of the try...except would succeed).  This PR removes the whole try...except which is now unnecessary.


rahulporuri (approved):
LGTM

PR #936, opened 2021-04-23
replace sys.exc_traceback with sys.exc_info()[2]

### LINKED ISSUE #930: Replace use of sys.exc_traceback with call to sys.exc_info
`sys.exc_traceback` has been deprecated for a long time and needs to be replaced with a call to `sys.exc_info`. Ref https://docs.python.org/2.7/library/sys.html#sys.exc_traceback.

See https://github.com/enthought/pyface/blob/master/pyface/base_toolkit.py#L170. Originally reported in https://github.com/enthought/pyface/issues/350#issuecomment-820164769

Note that this error was observed because the environment variable `ETS_DEBUG` was set to `True`. We might want to set this environment variable on the cron jobs that we run across ETS projects.

aaronayres35:
closes #930 

`sys._exc_traceback` has been deprecated and was causing an error if running the test suite with `ETS_DEBUG=True`.  This PR replaces it and we no longer see that error.

Unfortunately, running the test suite with `ETS_DEBUG=True` is still not a pretty sight, with numerous errors.  Those can be addressed in separate PRs

### DIFF
--- a/pyface/base_toolkit.py
+++ b/pyface/base_toolkit.py
@@ -167,7 +167,7 @@ class Toolkit(HasTraits):
                     # the traceback's stack frame mentions the toolkit in question.
                     import traceback
 
-                    frames = traceback.extract_tb(sys.exc_traceback)
+                    frames = traceback.extract_tb(sys.exc_info()[2])
                     filename, lineno, function, text = frames[-1]
                     if package not in filename:
                         raise


pyface/base_toolkit.py
```diff
@@ -167,7 +167,7 @@ def __call__(self, name):
                     # the traceback's stack frame mentions the toolkit in question.
                     import traceback
 
-                    frames = traceback.extract_tb(sys.exc_traceback)
+                    frames = traceback.extract_tb(sys.exc_info()[2])
```
aaronayres35, line 170: `sys.exc_info()` returns a tuple: `(type, value, traceback)`, see docs here: https://docs.python.org/3/library/sys.html#sys.exc_info

rahulporuri (approved):
LGTM.

> Unfortunately, running the test suite with `ETS_DEBUG=True` is still not a pretty sight, with numerous errors. Those can be addressed in separate PRs

Can you create an issue for that? We might not be able to get to it before the release.

aaronayres35:
> > Unfortunately, running the test suite with `ETS_DEBUG=True` is still not a pretty sight, with numerous errors. Those can be addressed in separate PRs
> 
> Can you create an issue for that? We might not be able to get to it before the release.

I've opened #937

PR #939, opened 2021-04-27
Cleanup code handling wx version < 2.8

rahulporuri:
This PR removes code that was handling wx version < 2.8. Wx less than 2.8 isnt even available via pypi for python 3. Ref https://pypi.org/project/wxPython/#history


aaronayres35 (approved):
LGTM

PR #944, opened 2021-06-14
Update super usage in the "qt" backend

rahulporuri:
This PR updates `super` usage in a semi-automated fashion. We used regex-based search and replace to update the `super` usage. Note that there were two instances where a super was being called with the wrong arguments - which got fixed in this PR.

See similar PR https://github.com/enthought/traits/pull/1280

### DIFF
--- a/pyface/ui/qt4/code_editor/replace_widget.py
+++ b/pyface/ui/qt4/code_editor/replace_widget.py
@@ -18,7 +18,9 @@ from .find_widget import FindWidget
 
 class ReplaceWidget(FindWidget):
     def __init__(self, parent):
-        super().__init__(parent)
+        # We explicitly call __init__ on the classes which FindWidget inherits from
+        # instead of calling FindWidget.__init__.
+        super(FindWidget, self).__init__(parent)
         self.adv_code_widget = weakref.ref(parent)
 
         # QFontMetrics.width() is deprecated and Qt docs suggest using
--- a/pyface/ui/qt4/code_editor/replace_widget.py
+++ b/pyface/ui/qt4/code_editor/replace_widget.py
@@ -18,7 +18,7 @@ from .find_widget import FindWidget
 
 class ReplaceWidget(FindWidget):
     def __init__(self, parent):
-        super(FindWidget, self).__init__(parent)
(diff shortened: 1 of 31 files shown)


pyface/ui/qt4/code_editor/replace_widget.py
```diff
@@ -18,7 +18,7 @@
 
 class ReplaceWidget(FindWidget):
     def __init__(self, parent):
-        super(FindWidget, self).__init__(parent)
```
rahulporuri, line 21: this was one of the two potentially buggy calls to `super`

pyface/ui/qt4/tests/test_gui.py
```diff
@@ -35,7 +35,7 @@ class SimpleApplication(HasStrictTraits):
     application_running = Event()
 
     def __init__(self):
-        super(HasStrictTraits, self).__init__()
```
rahulporuri, line 38: the second of two instances of potentially buggy super usage

pyface/ui/qt4/code_editor/replace_widget.py
aaronayres35, line 21: Note this class is not `FindWidget`. (although I believe this will still work fine, just pointing it out.
Also, the only different between this `__init__` method and the `FindWidget` `__init__` method is the line `form_layout.addRow("Rep&lace", self._create_replace_control())`

pyface/ui/qt4/tests/test_gui.py
aaronayres35, line 38: Just pointing out this also differs from the rest in that it did not call `super(SimpleApplication, self).__init__()` previously

aaronayres35 (approved):
LGTM

I went through and checked for any "potentially buggy" cases as we have called them previously.  Only found 2 which I believe should be fine.

pyface/ui/qt4/code_editor/replace_widget.py
rahulporuri: Ahh. Looks like this needs to stay as is.

aaronayres35 (approved):
LGTM

PR #945, opened 2021-06-14
Update super usage in the "wx" and "null" backends

rahulporuri:
This PR updates `super` usage in a semi-automated fashion. We used regex-based search and replace to update the `super` usage. Note that there was one instance where a `super` was being called with the wrong arguments - which got fixed in this PR and a redundant redefinition of `__init__` which got removed.

See similar PR https://github.com/enthought/pyface/pull/944 and https://github.com/enthought/traits/pull/1280

### DIFF
--- a/pyface/ui/wx/grid/simple_grid_model.py
+++ b/pyface/ui/wx/grid/simple_grid_model.py
@@ -38,17 +38,6 @@ class SimpleGridModel(GridModel):
     columns = Either(None, List(Instance(GridColumn)))
 
     # ------------------------------------------------------------------------
-    # 'object' interface.
-    # ------------------------------------------------------------------------
-    def __init__(self, **traits):
-        """ Create a SimpleGridModel object. """
-
-        # Base class constructor
-        super(SimpleGridModel, self).__init__(**traits)
-
-        return
-
-    # ------------------------------------------------------------------------
     # 'GridModel' interface.
     # ------------------------------------------------------------------------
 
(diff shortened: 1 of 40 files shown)


pyface/ui/wx/python_shell.py
```diff
@@ -309,7 +309,7 @@ def Destroy(self):
         self.redirectStdin(False)
         builtins.raw_input = self.raw_input
         self.destroy()
-        super(PyShellBase, self).Destroy()
```
rahulporuri, line 312: This is a potentially buggy use of `super`

pyface/ui/wx/grid/simple_grid_model.py
```diff
@@ -37,17 +37,6 @@ class SimpleGridModel(GridModel):
...
-    # ------------------------------------------------------------------------
-    # 'object' interface.
-    # ------------------------------------------------------------------------
-    def __init__(self, **traits):
-        """ Create a SimpleGridModel object. """
-
-        # Base class constructor
-        super(SimpleGridModel, self).__init__(**traits)
-
-        return
```
rahulporuri, line 49: we removed this redundant redefinition of `__init__`

pyface/ui/wx/python_shell.py
aaronayres35: ~This is strange as I believe `PyShellBase` only has a `destroy` method not a `Destroy` method:
https://wxpython.org/Phoenix/docs/html/wx.py.shell.Shell.html#wx.py.shell.Shell.destroy~
Actually `Destroy` is defined much further up the inheritance tree: https://wxpython.org/Phoenix/docs/html/wx.Window.html#wx.Window.Destroy

aaronayres35 (approved):
LGTM

PR #946, opened 2021-06-14
Cleanup super usage in toolkit agnostic code

rahulporuri:
This PR updates `super` usage in a semi-automated fashion. We used regex-based search and replace to update the `super` usage. Note that there was one unnnecessary redefinition of `__init__` which got removed in this PR.

See similar PR #944, #945 and https://github.com/enthought/traits/pull/1280

### DIFF
--- a/pyface/dock/dock_sizer.py
+++ b/pyface/dock/dock_sizer.py
@@ -431,9 +431,6 @@ class DockItem(HasPrivateTraits):
     #  Implementation of the 'owner' property:
     # ---------------------------------------------------------------------------
 
-    def __init__(self, **kw):
-        super().__init__(**kw)
-
     @cached_property
     def _get_owner(self):
         if self.parent is None:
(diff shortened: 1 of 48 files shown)


pyface/dock/dock_sizer.py
```diff
@@ -3830,9 +3830,6 @@ class DockInfo(HasPrivateTraits):
     # Dock Control:
     control = Instance(DockItem)
 
-    def __init__(self, **kw):
-        super(DockInfo, self).__init__(**kw)
```
rahulporuri, line 3834: this is an unnecessary redefinition of `__init__` which we removed

pyface/dock/dock_sizer.py
```diff
@@ -432,7 +432,7 @@ class DockItem(HasPrivateTraits):
     # ---------------------------------------------------------------------------
 
     def __init__(self, **kw):
-        super(DockItem, self).__init__(**kw)
+        super().__init__(**kw)
```
aaronayres35, line 435: can this just be removed entirely?

aaronayres35 (approved):
LGTM

just one additional unnecessary redefinition of `__init__` that could be removed

pyface/dock/dock_sizer.py
rahulporuri: removed

PR #949, opened 2021-06-14
Remove testing on pyqt4 from CI

aaronayres35:
Part of https://github.com/enthought/ets/issues/57

This PR removes testing with pyqt4 from CI


rahulporuri (approved):
LGTM

Issue #948, opened 2021-06-14
Use GitHub Actions CI instead of Travis CI


rahulporuri:
Along with the rest of ETS, we would like to move PyFace to use GitHub Actions CI instead of Travis CI.

PR #950, opened 2021-06-14
Port CI to gh-actions

aaronayres35:
part of #948 (which can be closed when we drop travis / appveyor)

This PR adds two _nearly_ identical work flows. One of which will run the test suite on every PR using packages from edm, the other which will run the test suite as a cron job using ets packages from source.


rahulporuri (approved):
LGTM

PR #951, opened 2021-06-15
Its "wx.StyledTextCtrl", not "wxStyledTextCtrl"

rahulporuri:
Call `__init__` on the right class - `wx.StyledTextCtrl` instead of `wxStyledTextCtrl`.


aaronayres35 (approved):
LGTM

Issue #899, opened 2021-03-08
Rewrite "*_changed" static trait listeners to use traits observe instead


rahulporuri:
The static trait change listeners `*_changed` (not `_anytrait_changed`) uses `on_trait_change` internally so we need to replace those methods to use `observe` instead. We need to rename the methods to ensure that traits `on_trait_change` doesn't use them by default - and then decorate them with the `observe` decorator.

PR #954, opened 2021-06-16
Update  _*_changed static trait listeners to use traits observe

aaronayres35:
Part of #899 

This PR replaces some of the instances of `_*_changed` static trait listeners with `observe` from the code base.  There are still plenty more that will need to be addressed in follow up PRs.

### DIFF
--- a/pyface/dock/dock_sizer.py
+++ b/pyface/dock/dock_sizer.py
@@ -36,6 +36,7 @@ from traits.api import (
     Undefined,
     Bool,
     cached_property,
+    observe,
 )
 from traitsui.dock_window_theme import dock_window_theme
 from traitsui.wx.helper import BufferDC
--- a/pyface/dock/dock_sizer.py
+++ b/pyface/dock/dock_sizer.py
@@ -2008,7 +2008,8 @@ class DockControl(DockItem):
     #  Handles the 'feature_changed' trait being changed:
     # ---------------------------------------------------------------------------
 
-    def _feature_changed(self):
+    @observe("feature_changed")
+    def _feature_changed_updated(self, event):
         """ Handles the 'feature_changed' trait being changed
         """
         self.set_feature_mode()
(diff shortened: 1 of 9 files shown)


pyface/dock/dock_sizer.py
```diff
@@ -2008,7 +2009,8 @@ def _activated_fired(self):
     #  Handles the 'feature_changed' trait being changed:
     # ---------------------------------------------------------------------------
 
-    def _feature_changed(self):
+    @observe("feature_changed")
+    def _feature_changed_updated(self, event):
```
aaronayres35, line 2013: I believe this had been a bug before I will look into writing a regression test

aaronayres35: EDIT: done

pyface/dock/tests/test_dock_sizer.py
```diff
@@ -0,0 +1,27 @@
...
+from unittest.mock import MagicMock
+
+from pyface.dock.dock_sizer import DockControl
+from pyface.toolkit import toolkit
+
+not_wx = toolkit.toolkit != "wx"
+
+class TestDockControl(unittest.TestCase):
+
+    @unittest.skipIf(not_wx, "This test is specific to the wx backend")
```
aaronayres35, line 21: To my surprise, `pyface/dock/dock_sizer.py` is wx specific although it lives in a generic module.https://github.com/enthought/pyface/blob/1a416eeea5146477a52f2dd1dbe6550111c54f2d/pyface/dock/dock_sizer.py#L20

rahulporuri:
@aaronayres35  looks like the `wx` import is causing CI to fail.

aaronayres35:
> @aaronayres35 looks like the `wx` import is causing CI to fail.

Agh, my mistake.  I pulled the problem import inside the test - hopefully that should fix things

pyface/dock/dock_sizer.py
```diff
@@ -2008,7 +2009,8 @@ def _activated_fired(self):
     #  Handles the 'feature_changed' trait being changed:
     # ---------------------------------------------------------------------------
 
-    def _feature_changed(self):
```
rahulporuri, line 2011: ahh looks like this was buggy before - `feature_changed` is the event trait

rahulporuri (approved):
LGTM

PR #955, opened 2021-06-17
Remove PyQt support in etstool CI utility

rahulporuri:
This PR removes support for the `PyQt` toolkit in the CI utility `etstool.py`. Now that we don't test against `PyQt` on CI anymore (see #949 ), we dont need to support `PyQt` in the `etstool` utility.


aaronayres35 (approved):
LGTM

PR #958, opened 2021-06-18
Address a few deprecation warnings when using python 3.8 and pyqt5

aaronayres35:
This PR addresses a few deprecation warnings observed using python 3.8 and pyqt5 on the traitsui and pyface test suites.


rahulporuri (approved):
LGTM

PR #959, opened 2021-06-19
IImage base interface for images

### LINKED ISSUE #909: More flexible image support
Currently all images used anywhere in Pyface or TraitsUI need to be `ImageResource` instances.  This is a limited use-case intended for images shipped with an application and stored as files.

There are frequent cases where a more generic image is to be displayed. If the image comes from a file, then the `ImageResource` infrastructure can be made to work by supplying an absolute path to the image, but for images in memory or from remote sources (eg. a URL) there is no way to display them other than saving out to a temporary file and then reading it back in.

For complex use-cases (eg. image editors) Enable/Chaco is often the right solution, but there are many use-cases where simple display of in-memory or url-based images is desirable.

A possible approach to this is:

- [ ] add an `IImage` interface which includes the `create_bitmap` and `create_icon` methods from `IImageResource`, but makes no assumptions about the location or storage of the data
- [ ] have `IImageResource` be a sub-interface of `IImage`
- [ ] change the `Image` trait to accept any `IImage` rather than any `IImageResource`

Everything up to here is effectively backwards compatible, as it is just injecting a new base interface.

- [ ] create other interfaces and concrete implementations for other types of images, such as `IImageBuffer` for an image stored in a memory buffer, or `IPILImage` for PIL/Pillow images, possibly `IURLImage` for a remote image, etc.

This might cause some problems in the rare case where the code is asking for the absolute path of an `IImageResource` (eg. in the About dialog, where it is generating HTML) but these can either be fixed or the API changed to be more restrictive if needed - worst case we might need a way to flag that the `Image` trait only accepts a particular sub-interface.

ETA: The `convert_bitmap` function also needs to be dealt with - it provides a cache for toolkit bitmaps, but that doesn't make sense for highly dynamic images, particularly if they are sharing a buffer somewhere.  It is used by TraitsUI's image editor, for example.  The simplest modification is to only do caching for `IImageResource` instances, and otherwise just return the bitmap and let the class do any caching needed.

Edit once more: `convert_bitmap` doesn't take into account image sizes; code should use an `ImageCache` instead, or we should re-think image caching.  This is a separate issue.

corranwebster:
This is the minimal functionality to add a new base `IImage` interface underneath `IImageResource` to potentally allow dynamic image types (eg. PIL/pillow images, numpy arrays, kiva GCs, SVG images) to be more easily used with Pyface and TraitsUI.

It implements additional concrete `IImage` class that wrap numpy arrays.

It also includes updates to the `Image` trait so that it accepts `IImage` instances as well as `IImageResource` and strings.

This fixes #909.  It should be backward compatible except as noted there.


rahulporuri:
@corranwebster is this PR ready for review?

corranwebster:
Yes - I have some extra implementation classes that are on my personal laptop, but they can go into a separate PR.

As of the last commit, tests are working and should be clean.

pyface/array_image.py
```diff
@@ -0,0 +1,95 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
```
rahulporuri, line 1: why add these modules to the top level `pyface` namespace when the more relevant `pyface.image` subpackage exists?

pyface/i_image_resource.py
```diff
@@ -35,68 +37,6 @@ class IImageResource(Interface):
...
-        Returns
-        -------
-        image : toolkit image
-            The toolkit image corresponding to the resource and the specified
-            size.
-        """
-
-    # FIXME v3: The need to distinguish between bitmaps and images is toolkit
-    # specific so, strictly speaking, the conversion to a bitmap should be done
-    # wherever the toolkit actually needs it.
```
rahulporuri, line 67: Is there any useful information in this fixme that we want to retain? _i don't think so IIUC_

pyface/ui/qt4/util/image_helpers.py
```diff
@@ -11,23 +11,96 @@
...
+from enum import Enum
 
 from pyface.qt import qt_api
-from pyface.qt.QtGui import QImage
+from pyface.qt.QtCore import Qt
+from pyface.qt.QtGui import QImage, QPixmap, QIcon
 
 
-def QImage_to_array(qimage):
+class ScaleMode:
```
rahulporuri, line 25:
```suggestion
class ScaleMode(Enum):
```

pyface/ui/qt4/util/image_helpers.py
```diff
@@ -11,23 +11,96 @@
...
+    fast = Qt.FastTransformation
+    smooth = Qt.SmoothTransformation
+
+
+class AspectRatio:
```
rahulporuri, line 30:
```suggestion
class AspectRatio(Enum):
```

pyface/ui/wx/util/image_helpers.py
```diff
@@ -0,0 +1,177 @@
...
+and Qt wx.Images, as well as between the various image types in a standardized
+way.
+"""
+
+from enum import Enum
+
+import wx
+
+
+class ScaleMode:
```
rahulporuri, line 23:
```suggestion
class ScaleMode(Enum):
```

pyface/ui/wx/util/image_helpers.py
```diff
@@ -0,0 +1,177 @@
...
+    fast = wx.IMAGE_QUALITY_NORMAL
+    smooth = wx.IMAGE_QUALITY_HIGH
+
+
+class AspectRatio:
```
rahulporuri, line 28:
```suggestion
class AspectRatio(Enum):
```

pyface/ui/qt4/util/image_helpers.py
```diff
@@ -11,23 +11,96 @@
...
+    ignore = Qt.IgnoreAspectRatio
+    keep_constrain = Qt.KeepAspectRatio
+    keep_expand = Qt.KeepAspectRatioByExpanding
+
+
+def image_to_bitmap(image):
+    """ Convert a QImage to a QPixmap.
+
```
rahulporuri, line 38: i noticed some white space here

```suggestion

```

pyface/ui/qt4/util/image_helpers.py
```diff
@@ -11,23 +11,96 @@
...
+    ------
+    bitmap : QPixmap
+        The corresponding QPixmap.
+    """
+    return QPixmap.fromImage(image)
+
+
+def bitmap_to_image(bitmap):
+    """ Convert a QPixmap to a QImage.
+
```
rahulporuri, line 54: white space here too 

```suggestion

```

pyface/ui/qt4/util/image_helpers.py
```diff
@@ -11,23 +11,96 @@
...
+    ------
+    image : QImage
+        The corresponding QImage.
+    """
+    return bitmap.toImage()
+
+
+def bitmap_to_icon(bitmap):
+    """ Convert a QPixmap to a QIcon.
+
```
rahulporuri, line 70:
```suggestion

```

pyface/ui/qt4/util/image_helpers.py
```diff
@@ -99,3 +172,7 @@ def array_to_QImage(array):
     image._numpy_data = data
     return image
 
+
+# backwards compatible names
+array_to_QImage = array_to_image
+QImage_to_array = image_to_array
```
rahulporuri, line 178: do want to remove these in the next major release? i'm not sure if anyone outside of pyface uses them directly

pyface/ui/wx/util/image_helpers.py
```diff
@@ -0,0 +1,177 @@
...
+    """ Resize a toolkit bitmap to the given size. """
+    image = bitmap_to_image(bitmap)
+    image = resize_image(image, size, aspect_ratio, mode)
+    return image_to_bitmap(image)
+
+
+def image_to_array(image):
+    """ Convert a wx.Image to a numpy array.
+
+    This copies the data returned from Qt.
```
rahulporuri, line 101:
```suggestion
    This copies the data returned from wx.
```

pyface/ui/wx/util/image_helpers.py
```diff
@@ -0,0 +1,177 @@
...
+    array = np.empty(shape=(width, height, 4), dtype='uint8')
+    array[:, :, :3] = rgb_data
+    array[:, :, 3] = alpha
+    return array
+
+
+def array_to_image(array):
+    """ Convert a numpy array to a wx.Image.
+
+    This copies the data before passing it to Qt.
```
rahulporuri, line 131:
```suggestion
    This copies the data before passing it to wx.
```

pyface/ui/wx/util/image_helpers.py
```diff
@@ -0,0 +1,177 @@
...
+    ----------
+    array : ndarray
+        An N x M x {3, 4} array of unsigned 8-bit ints.  The image
+        format is assumed to be RGB or RGBA, based on the shape.
+
+    Return
+    ------
+    image : wx.Image
+        The wx.Image created from the data.  The pixel format is
+        wx.Image.Format_RGB32.
```
rahulporuri, line 143: it looks like this function might return an image with 4 channels so shouldn't this be `RGB32 or ARGB32`?

pyface/ui_traits.py
```diff
@@ -38,7 +39,10 @@
 #  Images
 # -------------------------------------------------------------------------------
 
+# cache of lookups from string to ImageResource instance
 image_resource_cache = {}
+
+# cache of converssions of ImageResource instances to toolkit bitmaps
```
rahulporuri, line 45:
```suggestion
# cache of conversions of ImageResource instances to toolkit bitmaps
```

pyface/util/image_helpers.py
```diff
@@ -0,0 +1,31 @@
...
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+""" Helper functions for working with images
+
+This module provides helper functions for converting between numpy arrays
+and toolkit images, as well as between the various image types in a standardized
+way.
+"""
```
rahulporuri, line 16:
```suggestion
This module provides helper functions for converting between numpy arrays
and toolkit images, as well as between the various image types in a standardized
way.

Helper functions
----------------

- :data:`~.array_to_image`
- :data:`~.bitmap_to_icon`
- :data:`~.bitmap_to_image`
- :data:`~.image_to_array`
- :data:`~.image_to_bitmap`
- :data:`~.resize_image`
- :data:`~.resize_bitmap`

Options for resizing images
---------------------------
- :data:`~.ScaleMode`
- :data:`~.AspectRatio`

"""
```

rahulporuri (approved):
Mostly LGTM with a bunch of comments/corrections/suggestions.

pyface/array_image.py
corranwebster: The `pyface.image` subpackage is all about loading images from zipfile libraries of images.  These modules are at the same level or above `pyface.image_resources`, and so should live at the same level.

pyface/i_image_resource.py
corranwebster: No - this came from Wx wanting `wx.Bitmap` objects in many cases, where Qt in the same situation likes `QImage` objects (or sometimes vice-versa).  Part of this change is making the distinction more concrete, but toolkits won't change what they actually use.

pyface/ui/qt4/util/image_helpers.py
corranwebster: They might be being used in enable, but yes, they should probably go in the next major release.

pyface/ui/wx/util/image_helpers.py
corranwebster: Actually this is a search and replace bug.  `wx.Image` doesn't have a notion of format, it's all "RGB" with an optional seperate alpha channel:
```suggestion
        The wx.Image created from the data.
```

pyface/util/image_helpers.py
corranwebster: I'm assuming this is needed because of the way that API docs are generated?

pyface/ui/qt4/util/image_helpers.py
```diff
@@ -99,3 +172,7 @@ def array_to_QImage(array):
     image._numpy_data = data
     return image
 
+
+# backwards compatible names
```
corranwebster, line 176:
```suggestion
# backwards compatible names - will be removed in Pyface 8
```

rahulporuri:
@corranwebster can you also merge the master branch into this branch? It looks like GitHub Actions isnt running on this branch. Or you could potentially wait until after #967 is merged.

pyface/ui/qt4/util/image_helpers.py
corranwebster: I've added a not to this effect, and will open an issue.

pyface/util/image_helpers.py
rahulporuri: yep. i thought it'll make the autogenerated api docs better.

pyface/ui/qt4/util/image_helpers.py
rahulporuri: thank you!

rahulporuri: and it doesn't look like they're used outside pyface - https://github.com/search?q=org%3Aenthought+QImage_to_array&type=code and https://github.com/search?q=org%3Aenthought+array_to_QImage&type=code

corranwebster:
The changes to the Enums broke something - looking into it.

corranwebster:
OK, looks like they might need to be `IntEnum`; or we shouldn't use enums at all (this was an experiment on my part).

PR #960, opened 2021-06-19
PIL/Pillow image class

corranwebster:
This builds on #959 to provide an image class for PIL/Pillow `Image` instances.  This is part of #909 but not essential to it - we could decide to defer PIL support.

This adds an optional dependency on `pillow` - this is already a dependency of WxPython, so this is only potentially a problem for Qt users.


rahulporuri:
i'm removing myself as a reviewer for now - i'll wait until after #959 is merged

pyface/ui_traits.py
```diff
@@ -42,7 +42,11 @@
 # cache of lookups from string to ImageResource instance
 image_resource_cache = {}
 
+<<<<<<< HEAD
+# cache of converssions of ImageResource instances to toolkit bitmaps
+=======
 # cache of conversions of ImageResource instances to toolkit bitmaps
+>>>>>>> master
```
rahulporuri, line 49:
```suggestion
# cache of conversions of ImageResource instances to toolkit bitmaps
```

corranwebster: Yep, caught and fixed.

corranwebster:
This makes pillow an optional dependency:
- there is now an optional dependency for pillow
- `PILImage` is only used if imported directly - it isn't in `pyface.api`

pyface/i_pil_image.py
```diff
@@ -0,0 +1,73 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+""" The interface for an image resource. """
```
rahulporuri, line 10:
```suggestion
""" The interface for a PIL Image. """
```

pyface/i_pil_image.py
```diff
@@ -0,0 +1,73 @@
...
+    def __init__(self, image, **traits):
+        super().__init__(image=image, **traits)
+
+    def create_bitmap(self, size=None):
+        """ Creates a bitmap image for this image.
+
+        Parameters
+        ----------
+        size : (int, int) or None
+            The desired size as a width, height tuple, or None if wanting
```
rahulporuri, line 45:
```suggestion
            The desired size as a (width, height) tuple, or None if wanting
```

pyface/i_pil_image.py
```diff
@@ -0,0 +1,73 @@
...
+            default image size.  This is currently ignored.
```
rahulporuri, line 46: It doesn't look like the `size` argument is ignored - it gets passed through all the way down to `resize_image` from the looks of it. Isn't it?

```suggestion
            default image size.
```

pyface/i_pil_image.py
```diff
@@ -0,0 +1,73 @@
...
+        return image_to_bitmap(self.create_image(size))
+
+    def create_icon(self, size=None):
+        """ Creates an icon for this image.
+
+        Parameters
+        ----------
+        size : (int, int) or None
+            The desired size as a width, height tuple, or None if wanting
+            default icon size.  This is currently ignored.
```
rahulporuri, line 64: same comments here as above

```suggestion
            The desired size as a (width, height) tuple, or None if wanting
            default icon size.
```

pyface/pil_image.py
```diff
@@ -0,0 +1,19 @@
...
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+""" The implementation of an image resource. """
```
rahulporuri, line 11:
```suggestion
""" The implementation of a IPILImage. """
```

pyface/ui/qt4/pil_image.py
```diff
@@ -0,0 +1,48 @@
...
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+from pyface.qt.QtGui import QIcon, QPixmap
```
rahulporuri, line 11: Tangent : ~I always wondered why these `pyface.qt` imports preceeded other imports. Why is that?~ Never mind. I think it's because the `pyface.qt` imports are technically Qt-imports which are third party packages and so they should exist before Enthought library imports.

I wish this was codified somewhere because the answer isn't obvious.

pyface/ui/qt4/pil_image.py
```diff
@@ -0,0 +1,48 @@
...
+    # 'IImage' interface.
+    # ------------------------------------------------------------------------
+
+    def create_image(self, size=None):
+        """ Creates a Qt image for this image.
+
+        Parameters
+        ----------
+        size : (int, int) or None
+            The desired size as a width, height tuple, or None if wanting
```
rahulporuri, line 34:
```suggestion
            The desired size as a (width, height) tuple, or None if wanting
```

pyface/ui/wx/pil_image.py
```diff
@@ -0,0 +1,52 @@
...
+    # 'IImage' interface.
+    # ------------------------------------------------------------------------
+
+    def create_image(self, size=None):
+        """ Creates a Wx image for this image.
+
+        Parameters
+        ----------
+        size : (int, int) or None
+            The desired size as a width, height tuple, or None if wanting
```
rahulporuri, line 34:
```suggestion
            The desired size as a (width, height) tuple, or None if wanting
```

pyface/ui/wx/util/image_helpers.py
```diff
@@ -33,7 +33,6 @@ class AspectRatio(IntEnum):
 
 def image_to_bitmap(image):
     """ Convert a wx.Image to a wx.Bitmap.
-
```
rahulporuri, line 36: Don't we normally have a new line in the function docstring between the description and the parameters?

rahulporuri (approved):
LGTM with a few nitpicky comments.

pyface/ui/qt4/pil_image.py
corranwebster: It's not done universally.  In this case it was likely just copy-pasted from `image_resource.py`.  In other cases as you note it's likely from taking Wx code and translating it and so the Wx imports get replaced by Qt imports in the same place.

It would be nice to standardize, but that would require going through and actually standardizing

pyface/ui/wx/util/image_helpers.py
corranwebster: Yes.  Not sure what happened here.  It might be a bad merge.

corranwebster:
Right, I think this is good.  If everything passes I will merge.

PR #962, opened 2021-06-22
Toggle Fields (CheckBoxField, RadioButtonField, etc.)

corranwebster:
This PR adds additional fields with toggling behaviour: check boxes, radio buttons, and toggle buttons.

The purpose of this functionality is to abstract the common API to permit an eventual refactor of the TraitsUI `BooleanEditor` to allow styles other than checkboxes.

### DIFF
--- a/docs/source/fields.rst
+++ b/docs/source/fields.rst
@@ -66,7 +66,7 @@ ToggleField and Subclasses
 ==========================
 
 The :py:class:`~pyface.fields.i_toggle_field.IToggleField` interface holds a
-boolean :py:attr:`~pyface.fields.i_spin_field.IToggleField.value` that is
+boolean :py:attr:`~pyface.fields.i_toggle_field.IToggleField.value` that is
 toggled between ``True`` and ``False`` by the widget.  The interface is
 implemented by several different concrete classes with different appearances
 but similar behaviour:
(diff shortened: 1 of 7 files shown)


docs/source/fields.rst
```diff
@@ -62,4 +62,25 @@ values, a :py:attr:`~pyface.fields.i_spin_field.IComboField.formatter` function
 should be provided - this defaults to either :py:func:`str` (Python 3+) or
 :py:func:`unicode` (Python 2).
 
+ToggleField and Subclasses
+==========================
 
+The :py:class:`~pyface.fields.i_toggle_field.IToggleField` interface holds a
+boolean :py:attr:`~pyface.fields.i_spin_field.IToggleField.value` that is
```
rahulporuri, line 69: should this be 

```suggestion
boolean :py:attr:`~pyface.fields.i_toggle_field.IToggleField.value` that is
```

pyface/fields/toggle_field.py
```diff
@@ -0,0 +1,22 @@
...
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+#
+# Author: Enthought, Inc.
+# Description: <Enthought pyface package component>
```
rahulporuri, line 12: nit: do we really need these or were they copy-pasted from some other module?

rahulporuri (approved):
Mostly LGTM with a couple of small comments

pyface/fields/toggle_field.py
corranwebster: Copy-pasted.  I don't think we've done the header normalization for pyface: some of the copyright dates were also out of sync.

PR #964, opened 2021-06-30
Address a Qt DeprecationWarning and ResourceWarnings

rahulporuri:
This PR addresses a specific Qt `DeprecationWarning` - replacing the deprecated `font_metrics.width` method with the
recommended `font_metrics.horizontalAdvance` method. This PR also fixes `ResourceWarning`s by properly closing temporary files that a test had created/opened.

Ref https://doc.qt.io/qt-5/qfontmetrics-obsolete.html#width and https://doc.qt.io/qt-5/qfontmetrics.html#horizontalAdvance


aaronayres35 (approved):
LGTM

I was not seeing the deprecation warnings on master, but nonetheless the changes LGTM.  Also I confirmed I no longer see the ResourceWarnings

rahulporuri:
> For some reason I was not seeing the deprecation warnings on master

I first noticed them when I ran the testsuite on Python 3.8 - not Python 3.6 that we use by default.

PR #969, opened 2021-07-07
Use "ubuntu-latest" in GitHub Actions CI

rahulporuri:
This PR updates GitHub Actions CI to use `ubuntu-latest` instead of using an older version of ubuntu. I'm not entirely sure why we were pinning CI to use an older version of ubuntu in the first place - we might just have copied over existing CI configuration from another ETS repository.

### DIFF
--- a/.github/workflows/ets-from-source.yml
+++ b/.github/workflows/ets-from-source.yml
@@ -13,7 +13,7 @@ jobs:
   test-with-edm:
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


aaronayres35 (approved):
Changes LGTM, but it looks like CI is still waiting on / expecting ubuntu-18.04 jobs to be run?

rahulporuri:
> Changes LGTM, but it looks like CI is still waiting on / expecting ubuntu-18.04 jobs to be run?

yeah. fixed it. merging now.

PR #976, opened 2021-07-13
HeadingText fixes

corranwebster:
Improvements and fixes for HeadingText:

- aligns Qt _create_control with standard API
- routes creation through create() call
- makes widget creation in init() optional
- deprecates automatic widget creation
- makes appearance of Qt and Wx similar (deprecates background image for wx)
- changes to text are dispatched on the ui thread
- general clean-up and standardization
- better testing
- various other small fixes

### DIFF
--- a/pyface/i_heading_text.py
+++ b/pyface/i_heading_text.py
@@ -10,11 +10,9 @@
 
 """ Heading text. """
 
+import warnings
 
-from traits.api import HasTraits, Instance, Int, Interface, Str
-
-
-from pyface.i_image_resource import IImageResource
+from traits.api import HasTraits, Int, Interface, Str
 
 
 class IHeadingText(Interface):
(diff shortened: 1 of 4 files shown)


pyface/i_heading_text.py
```diff
@@ -10,32 +10,109 @@
...
+
+    def __init__(self, parent=None, **traits):
+        """ Creates the heading text. """
+
+        if "image" in traits:
+            warnings.warn(
+                "background images are no-longer supported for Wx and the "
+                "'image' trait will be removed in a future Pyface update",
+                PendingDeprecationWarning,
+            )
```
rahulporuri, line 51: can you create an issue for this deprecation - and put it in the 8.0.0 milestone?

pyface/i_heading_text.py
```diff
@@ -10,32 +10,109 @@
...
+
+        if create:
+            # Create the widget's toolkit-specific control.
+            self.create()
+            warnings.warn(
+                "automatic widget creation is deprecated and will be removed "
+                "in a future Pyface version, use create=False and explicitly "
+                "call create() for future behaviour",
+                PendingDeprecationWarning,
+            )
```
rahulporuri, line 66: same comment here regd an issue - and we can put this in the 8.0.0 milestone too if that makes sense.

rahulporuri (approved):
LGTM

pyface/i_heading_text.py
corranwebster: This is part of #729 (particularly the comment at the bottom about automatic creation in `__init__`).

But I don't want to add it to a milestone until all of them are dealt with (ie. if I don't get it done this week it might bet Pyface 9 where we finally get rid of them).

pyface/i_heading_text.py
corranwebster: Will do.

corranwebster: Done #978

pyface/i_heading_text.py
corranwebster: I've added a comment to #729 which goes over the idea for the sequence of events needed to change the API safely.

PR #977, opened 2021-07-13
SplitWidget fixes

corranwebster:
Some improvements to the `SplitWidget` class and related classes (particularly `SplitPanel`)

- clean up of traits (including adding an "Orientation" trait)
- deprecate automatic creation of widgets in `SplitPanel.__init__` (part of #729)
- add an `Orientation` trait and use it where applicable

### DIFF
--- a/pyface/split_panel.py
+++ b/pyface/split_panel.py
@@ -10,6 +10,7 @@
 
 """ A panel that is split in two either horizontally or vertically. """
 
+import warnings
 
 from pyface.split_widget import SplitWidget
 from pyface.widget import Widget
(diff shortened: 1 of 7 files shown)


pyface/split_panel.py
```diff
@@ -22,11 +23,24 @@ class SplitPanel(Widget, SplitWidget):
...
+
+        if create:
+            # Create the widget's toolkit-specific control.
+            self.create()
+            warnings.warn(
+                "automatic widget creation is deprecated and will be removed "
+                "in a future Pyface version, use create=False and explicitly "
+                "call create() for future behaviour",
+                PendingDeprecationWarning,
+            )
```
rahulporuri, line 42: same comment here regarding an issue to handle this deprecation - looks like a single issue will suffice to handle all such deprecations related to the `create` method.

rahulporuri (approved):
LGTM with one small comment

pyface/split_panel.py
corranwebster: Same reply as before - I've updated #729 to add the sequence around this.

Actually, should probably add #729 to the 8.0 milestone.

Issue #979, opened 2021-07-13
Remove Wx IPythonWidget


corranwebster:
Last version of IPython supported looks to have been around 0.10 (!).  I would be shocked if this worked in any way right now.

PR #980, opened 2021-07-13
Fixes for `PythonEditor` and `PythonWidget`

corranwebster:
This mainly kicks-off the deprecation of automatic creation of the widgets in the `__init__` method.

This also deprecates the very old wx IPython widget - see #979.


rahulporuri (approved):
LGTM.

corranwebster:
> Looks like the warnings are showing up in the test run output. It would be nice/cleaner if they didn't.

Right - I forgot to adjust the test cases to avoid using auto-creation.  That needs fixing.

corranwebster:
That should fix the warnings from PythonEditor/PythonWidget.  I'll make a separate PR to fix any others.

PR #982, opened 2021-07-13
Implement ILayeredPanel.

corranwebster:
This takes the existing wx-only `LayeredPanel` and:

- adds an `ILayeredPanel` interface
- implements the Qt version of it
- adds tests for it
- deprecates autocreation in the `__init__` method
- updates the usage of `LayeredPanel` in `Wizard` and `PreferencesDialog`

It does not attempt to improve the `LayeredPanel` API (eg. by allowing removal of layers, etc.)

### DIFF
--- a/pyface/i_layered_panel.py
+++ b/pyface/i_layered_panel.py
@@ -14,6 +14,8 @@ A layered panel contains one or more named layers, with only one layer
 visible at any one time (think of a 'tab' control minus the tabs!).
 """
 
+import warnings
+
 from traits.api import Any, Dict, HasTraits, Interface, Str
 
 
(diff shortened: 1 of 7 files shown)


pyface/i_layered_panel.py
```diff
@@ -0,0 +1,99 @@
...
+        """ Creates a new LayeredPanel. """
+
+        create = traits.pop("create", True)
+
+        # Base class constructor.
+        super().__init__(parent=parent, **traits)
+
+        if create:
+            # Create the toolkit-specific control that represents the widget.
+            self.create()
```
rahulporuri, line 99: what about the warnings related to not automatically calling `create`?

rahulporuri (approved):
LGTM with one minor question

pyface/i_layered_panel.py
corranwebster: Thanks.

PR #983, opened 2021-07-13
Setup slack notification for cron job

rahulporuri:
This PR sets up the slack notification for the cron job.

I manually triggered the cron job to check that the slack notification gets fired. Ref https://github.com/enthought/pyface/actions/runs/1027257422.

One more step towards https://github.com/enthought/ets/issues/65


aaronayres35 (approved):
LGTM

Not sure what happened on the run though - failure was reported to slack but logs on workflow aren't there

rahulporuri:
> I think some of the failures might be just from having so many jobs running on GH actions at once across all ets repos

You're probably right 😅 There were definitely one too many jobs running yesterday. For now, i'll go ahead and merge this - and revisit this if things don't work as expected over the weekend.

PR #984, opened 2021-07-13
Fix generic observers for TraitGridModel

corranwebster:
This is a quick-fix for a place where `on_trait_change` -> `observe` didn't take into account the lack of a matcher.  For `on_trait_change` this matches everything, for `observer` it is an error.

To validate the fix run the `examples/grid.py` with the Wx toolkit.  It will fail without this fix, but run correctly after it.

### DIFF
--- a/pyface/ui/wx/expandable_panel.py
+++ b/pyface/ui/wx/expandable_panel.py
@@ -13,7 +13,7 @@
 
 import wx
 
-from traits.api import Instance
+from traits.api import Dict, Instance, Str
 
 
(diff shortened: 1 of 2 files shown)


corranwebster:
Arg! It looks like I managed to pull in the deprecation of autocreation in `ExpandablePanel` as well.

I'll leave it to the reviewer to decide whether this is an unforgivable sin (in which case I will close and untangle), or whether it's fine to include it.

corranwebster:
Closing because it's wrong and because the mac jobs are stuck.  Will fix later.

PR #985, opened 2021-07-14
Fix observer usage in `pyface.wx.grid`

corranwebster:
This is a quick-fix for a place where `on_trait_change` -> `observe` didn't take into account the lack of a second positional argument. For `on_trait_change` this matches everything, for observer it is an error.

To validate the fix run `examples/grid.py` with the Wx toolkit. It will fail without this fix, but run correctly after it.


pyface/ui/wx/grid/trait_grid_model.py
```diff
@@ -695,5 +698,7 @@ def __manage_column_listeners(self, collist, remove=False):
             for col in collist:
                 if isinstance(col, TraitGridColumn):
                     col.observe(
-                        self._on_columns_changed, remove=remove
-                    )
+                        self._on_columns_changed,
+                        match(lambda name, trait: True),
+                        remove=remove,
+                    )
```
rahulporuri, line 704:
```suggestion
                    )

```

rahulporuri (approved):
Thanks for the catch! LGTM!

Issue #987, opened 2021-07-14
Deprecate/remove `ToolPalette` and related code


corranwebster:
This is currently supported only by Wx, and the implementation is currently not working.

A tool palette is not a core widget in either Wx or Qt, and most interactions provided by a palette can be also handled by a toolbar (or a collection of toolbars).

Because it isn't working, it can probably be directly removed, but we could go through a brief deprecation phase if needed.

rahulporuri:
> Because it isn't working, it can probably be directly removed, but we could go through a brief deprecation phase if needed.

I would very much prefer not going through a deprecation phase, however brief, as it isn't used by active projects, especially because it isn't already working.

PR #988, opened 2021-07-14
Fix Wx examples

### LINKED ISSUE #986: Regressions in wxPython examples
A number of these that were working aren't for a variety of reasons.

corranwebster:
This PR implements some fixes so that some broken examples work again under WxPython:

- fixed a typo in an observed trait name
- added a WxPython 4.0 compatibility branch in a control
- fixed up (and simplified) the `ExpandingPanel` widget

No tests, but the examples should now work.

Did not fix (and not going to) the IPython example and the `tool_palette.py` example.  See #1 and #979  and #987.

This should fix #986


pyface/ui/wx/expandable_header.py
```diff
@@ -184,48 +181,31 @@ def _draw_title(self, dc):
...
-            self._background_bmp = self._header_bmp
-
-        self.control.Refresh()
-        event.Skip()
-
     def _on_down(self, event):
         """ Called when button is pressed. """
 
-        # print 'ImageButton._on_down'
+        #print('ImageButton._on_down')
```
rahulporuri, line 208:
```suggestion
```

rahulporuri (approved):
LGTM

Issue #940, opened 2021-04-27
 wx.DEFAULT is deprecated in favor of wx.FONTFAMILY_DEFAULT


rahulporuri:
https://github.com/enthought/pyface/blob/c5ce5a63045e921fcce462a5e8e1017f84a5bf10/pyface/wx/spreadsheet/default_renderer.py#L34

Note that similarly, other aliases have been created - https://wxpython.org/Phoenix/docs/html/MigrationGuide.html#font-pen-and-brush-styles

Palash-Vishnani:
I can work on this issue, if it is not assigned to anyone.

rahulporuri:
👋  @Palash-Vishnani . No one is actively working on this issue so please feel free to assign it to yourself and work on it. Thanks for showing interest in contributing to Pyface.

Palash-Vishnani:
@rahulporuri Do I just have to update `wx.DEFAULT` , `wx.NORMAL` and `wx.NORMAL` to the newest `wx.FONTFAMILY_DEFAULT` , `wx.FONTSTYLE_NORMAL` and `wx.FONTWEIGHT_NORMAL` by rewriting them?? Or do I have to do something else?

rahulporuri:
> Do I just have to update wx.DEFAULT , wx.NORMAL and wx.NORMAL to the newest wx.FONTFAMILY_DEFAULT , wx.FONTSTYLE_NORMAL and wx.FONTWEIGHT_NORMAL by rewriting them?? 

Yup. That's all that is needed here.

PR #992, opened 2021-07-15
Updated Line 34 as per #940

Palash-Vishnani:
Updated deprecated font-related aliases on line 34

### DIFF
--- a/pyface/wx/spreadsheet/default_renderer.py
+++ b/pyface/wx/spreadsheet/default_renderer.py
@@ -31,7 +31,7 @@ class DefaultRenderer(PyGridCellRenderer):
         self.foundary = font
         self.fontsize = fontsize
         self.font = wx.Font(
-            fontsize, wx.DEFAULT, wx.NORMAL, wx.NORMAL, 0, font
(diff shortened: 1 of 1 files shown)


corranwebster:
Thanks for the PR!

I think the point of #940 (although perhaps not clearly expressed) was that these constants need to be updated throughout the codebase (eg. here:
https://github.com/enthought/pyface/blob/004d6ee43acaa51b17e78707f8aeb1118d818725/examples/undo/example_editor_manager.py#L76
and here:
https://github.com/enthought/pyface/blob/004d6ee43acaa51b17e78707f8aeb1118d818725/pyface/wx/scrolled_message_dialog.py#L34
and so on...)

Palash-Vishnani:
@corranwebster @rahulporuri  If you want I can update all such constants through the codebase and then send you a PR, but it will take some time. Until then are you merging this one??

corranwebster:
I'll leave that decision up to @rahulporuri.

rahulporuri:
Thanks for the contribution @Palash-Vishnani ! I'm happy merging this PR as it is. If you can find the time, we'd appreciate any subsequent PRs that will help close #940 .

rahulporuri (approved):
LGTM

PR #993, opened 2021-07-15
Deprecate `__init__` methods which create the widget

corranwebster:
As discussed in #729 this does the discussed deprecation transforms in the remaining Wx widgets that use that approach.

Most of these widgets have not tests, but they are exercised by the examples.

### DIFF
--- a/pyface/ui/wx/image_widget.py
+++ b/pyface/ui/wx/image_widget.py
@@ -10,12 +10,11 @@
 
 """ A clickable/draggable widget containing an image. """
 
+import warnings
 
 import wx
 
-
 from traits.api import Any, Bool, Event
-
 
 from .widget import Widget
 
@@ -64,10 +63,24 @@ class ImageWidget(Widget):
 
     def __init__(self, parent, **traits):
         """ Creates a new widget. """
-
         # Base class constructors.
-        super().__init__(**traits)
 
+        create = traits.pop('create', True)
+
+        # Base-class constructors.
+        super().__init__(parent=parent, **traits)
+
+        # Create the widget!
+        if create:
+            self.create()
+            warnings.warn(
+                "automatic widget creation is deprecated and will be removed "
+                "in a future Pyface version, use create=False and explicitly "
+                "call create() for future behaviour",
+                PendingDeprecationWarning,
+            )
+
+    def _create_control(self, parent):
         # Add some padding around the image.
         size = (self.bitmap.GetWidth() + 10, self.bitmap.GetHeight() + 10)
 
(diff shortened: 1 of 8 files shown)


pyface/ui/wx/image_widget.py
```diff
@@ -96,7 +109,7 @@ def __init__(self, parent, **traits):
             wx.SystemSettings.GetColour(wx.SYS_COLOUR_3DHIGHLIGHT), 1, wx.SOLID
         )
 
-        return
+        return self.control
```
rahulporuri, line 112: nit-picky comment - in an earlier case, we didn't `self.control` in the `_create_control` method and instead chose to simply return it instead - whereas in this method, we're setting and returning `self.control`. Can i ask if there is a specific reason?

pyface/ui/wx/progress_dialog.py
```diff
@@ -46,20 +47,39 @@ def __init__(
...
         """
         Constructs a progress bar which can be put into a panel, or optionaly,
         its own window
 
         """
-        self._max = maximum
-        self.parent = parent
+        create = traits.pop("create", True)
+
+        # XXX minimum is ignored - it either should be deprecated or supported
```
rahulporuri, line 59: Can we have an issue please - to remove it outright for the next major release? I don't think we have the resources to figure out how to support `minimum` at the moment - and we can implement that functionality from scratch if we ever need to in the future.

pyface/ui/wx/progress_dialog.py
```diff
@@ -46,20 +47,39 @@ def __init__(
...
+        super().__init__(
+            parent=parent,
+            _max=maximum,
+            direction=direction,
+            size=size,
```
rahulporuri, line 64: looks like `size` is ignored as well - it isn't defined on the `IWidget` interface or on the wx `Widget` class

rahulporuri (approved):
LGTM with a couple of comments. I didn't test that the examples worked as expected with the wx toolkit though.

pyface/ui/wx/image_widget.py
corranwebster: There were some widgets with methods called in `_create_control` that expect `self.control` to not be populated.  These widgets need their create methods to be untangled and have some code moved to `_add_event_listeners` and other code moved to post-creation initialization.

For example, in this class, all of the binding methods should really be in `_add_event_listenrs`.

Since this PR is all about deprecating auto-creation, I wanted to make the smallest changes to achieve the required results.  I'll add an issue about untangling creation and initialization.

corranwebster: Issue created: #995

pyface/ui/wx/progress_dialog.py
corranwebster: Issue created for this and `minimum`: #996

Issue #972, opened 2021-07-08
Add ArrayImage and PILImage to pyface.api


corranwebster:
These have optional dependencies (numpy and pillow respectively) and so should be optionally imported when added to the `pyface.api` module.

Palash-Vishnani:
@corranwebster I can work on this issue.
Do I just have to edit this file: [https://github.com/enthought/pyface/blob/master/pyface/api.py](url) , or do I have to edit any more files to import the dependencies and modules??

Palash-Vishnani:
@corranwebster I can work on this issue.
Do I just have to edit this file: https://github.com/enthought/pyface/blob/master/pyface/api.py , or do I have to edit any more files to import the dependencies and modules??

rahulporuri:
> Do I just have to edit this file: https://github.com/enthought/pyface/blob/master/pyface/api.py

👋  @Palash-Vishnani . Yes, `pyface.api` is the only file that you need to modify. But, note that you should not add `PILImage`. The PR which introduces `PILImage` hasn't been merged yet (ref #960).

rahulporuri:
closing as #997 and #1002 solved this issue

PR #997, opened 2021-07-17
Added ArrayImage to pyface.api as per #972

Palash-Vishnani:


Palash-Vishnani:
@corranwebster @rahulporuri checkout this PR.

pyface/api.py
```diff
@@ -112,7 +112,7 @@
 from .layered_panel import LayeredPanel
 from .message_dialog import error, information, warning, MessageDialog
 from .progress_dialog import ProgressDialog
-
+from .array_image import ArrayImage
```
rahulporuri, line 115: The imports are sorted alphabetically in this file so can you please update the location of this import statement?

rahulporuri (commented):
This PR is almost ready. Apart from the comment below, can you also add `ArrayImage` to the module docstring at the top of the file? `ArrayImage` can be included in the "Miscellaneous" section.

Palash-Vishnani:
@rahulporuri It's fine now right??

corranwebster:
There's a technical gotcha in this which I think @rahulporuri is the person to make a decision.  Currently I think that NumPy is technically an optional dependency of Pyface with the Qt backend (it is required for Wx for various historical reasons).  This change means the NumPy is required when importing `pyface.api`.  For that reason, we should guard the import in the same way as the pieces which depend on Pygments:

https://github.com/enthought/pyface/blob/997c78f0df87f663fd93a2bd7010c9352290b335/pyface/api.py#L118-L123

but with numpy in the place of Pygments.

Alternatively @rahulporuri may say that depending on `numpy` is fine, but in that case we need to update the requirements here:

https://github.com/enthought/pyface/blob/997c78f0df87f663fd93a2bd7010c9352290b335/pyface/__init__.py#L21-L27

This is why I didn't put `ArrayImage` in the API module during the recent PR.

Palash-Vishnani:
@corranwebster @rahulporuri Soo what to do with this PR now? Do I need to do more modifications in it, or will it be merged after some time?

rahulporuri:
@Palash-Vishnani sorry for the delayed response but @corranwebster is right. I missed it when i first reviewed this PR. Can you make the changes that corran recommended i.e. something like this

```
 # Excuse pygments dependency (for Qt), otherwise re-raise 
 with _optional_import( 
         "numpy", 
         msg="ArrayImage not available due to missing numpy.", 
         logger=_logging.getLogger(__name__)): 
     from .array_image import ArrayImage
```

Can you please add this new import next to the other optional imports?

Palash-Vishnani:
@rahulporuri I'll do it but should I remove this import satement `from .array_image import ArrayImage`  and `ArrayImage` module docstring from the top? Or should I keep them as it is.?

rahulporuri:
> @rahulporuri I'll do it but should I remove this import satement `from .array_image import ArrayImage` from the file? 

Yes. You need to replace the `from .array_image import ArrayImage` with the new `with _optional_import(...):` code.

> And also do I need to write this statement in the Miscellaneous section of module docstrings: "Note that the :class:~.ArrayImage is only available if the numpy package is available in the Python environment."

Yes please. That's right. You need to add this information in the module docstring to make it clear to end users.

rahulporuri (approved):
LGTM. Merging now.

rahulporuri:
Thanks @Palash-Vishnani for the contribution!

If you're interested in doing more, you can also add the `PILImage` class to `pyface.api`. The changes there will almost be the same as this PR - except `PILImage` depends on the `pillow` python library so you need to use `with _optional_import("pillow", ...)`. With that PR, we will be able to close the issue #972.

PR #1003, opened 2021-07-22
Tests for ImageLibrary and related code

corranwebster:
These are more integration tests than unit tests, and don't attempt to test anything which changes global state (eg. adding to ImageLibrary, etc).

Total coverage is ~68%, but that's a lot better than 0 and these tests will hopefully be enough to:
- catch breaking changes
- allow future refactoring with some degree of confidence

### DIFF
--- a/pyface/image/tests/__init__.py
+++ b/pyface/image/tests/__init__.py
@@ -1,0 +1,9 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
--- a/pyface/image/tests/test_image.py
+++ b/pyface/image/tests/test_image.py
@@ -1,0 +1,582 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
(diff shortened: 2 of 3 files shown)


pyface/image/tests/__init__.py
```diff
@@ -0,0 +1,9 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
```
rahulporuri, line 9: we're leaving the `__init__` files empty and we're not adding the copyright headers in them.

```suggestion
```

pyface/image/tests/test_image.py
```diff
@@ -0,0 +1,582 @@
...
+    def test_extension(self):
+        volume_name, file_name = split_image_name("@icons:red_ball")
+
+        self.assertEqual(volume_name, "icons")
+        self.assertEqual(file_name, "red_ball.png")
+
+    def test_no_at_symbol(self):
+        volume_name, file_name = split_image_name("icons:red_ball.jpg")
+
+        # XXX this should probably raise rather than doing this
```
rahulporuri, line 69: Do you think it makes sense to open an issue for this instead of leaving this information hidden in a comment in a test?

rahulporuri (approved):

pyface/image/tests/test_image.py
corranwebster: Done.

PR #1004, opened 2021-07-23
Explicitly import "numpy" for "ArrayImage" import

rahulporuri:
This PR explicitly adds a `numpy` import before the `ArrayImage` import to ensure that an `ImportError` is raised instead of a `TraitError`, which will get caught and handled correctly by the `_optional_import` utility.

This was first observed thanks to the cron job - https://github.com/enthought/envisage/runs/3139307870


corranwebster (approved):
LGTM

PR #1005, opened 2021-07-26
Update almost all _fired methods to use observe

aaronayres35:
This PR simply replaces almost all uses of the `_fired` magically named methods with their observe equivalent.

The only lingering method is here:
https://github.com/enthought/pyface/blob/6c98a1a8b1fd2995efc99ff7d3221563458adb28/pyface/ui/qt4/dialog.py#L191-L195
Which I believe will need special care because this is also making use of the qt signal handling https://github.com/enthought/pyface/blob/6c98a1a8b1fd2995efc99ff7d3221563458adb28/pyface/ui/qt4/dialog.py#L172


rahulporuri (approved):
Almost LGTM. Looks like you committed the changes to `icons.zip` and `std.zip` files.

PR #1011, opened 2021-08-09
 Default branch has been renamed from "master" to "main"

rahulporuri:
This PR replaces the use of ``master`` in the repo with ``main`` as the default branch name has been updated.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -165,7 +165,7 @@ except ImportError as exc:
 # Useful aliases to avoid repeating long URLs.
 extlinks = {
     'github-examples': (
-        'https://github.com/enthought/pyface/tree/master/examples/%s',
+        'https://github.com/enthought/pyface/tree/main/examples/%s',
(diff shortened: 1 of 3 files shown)


docs/source/conf.py
```diff
@@ -165,7 +165,7 @@
 # Useful aliases to avoid repeating long URLs.
 extlinks = {
     'github-examples': (
-        'https://github.com/enthought/pyface/tree/master/examples/%s',
+        'https://github.com/enthought/pyface/tree/main/examples/%s',
```
rahulporuri, line 168: same comment here as in traitsui - i think we should be linking to the specific version/tag instead of just linking to main branch. will fix this in a separate PR.

mdickinson (approved):
LGTM

rahulporuri:
Update - same as with the other ETS repositories, the examples links are being redirected to use the `main` branch instead of `master` by github automagically.

PR #1013, opened 2021-08-11
Add an `_initialize_control()` method to `IWidget`

corranwebster:
Initially this does nothing, but it allows `super()` calls in `_create()` methods for widgets which want to do post-creation initialization that respects inheritance.

Updates `IField` and `IDataViewWidget` to take advantage of this.

This is a minor change that permits a number of other things to move forward more naturally.


rahulporuri (approved):
LGTM.

Apologies for the extremely delayed review of what is mostly a trivial PR.

Issue #738, opened 2020-10-11
Add an IContainer Interface


corranwebster:
There is already an informal interface used in a number of places where a `Widget` which contains other widgets, such as `IDialog`, `IApplicationWindow`, and possibly others, expect subclasses to override the `_create_contents` method to populate and layout the main area of the control.  These widgets may have additional contents beyond these things (eg. a dialog's buttons), so this method provides a hook for customization.

The interface would look  something like:
```
class IContainer(IWidget):
    def _create_contents(self, parent):
        pass
```
where `parent` is the parent widget for the contents (usually, but not always, `self.control`) and the method is expected to return a toolkit object.

Such an interface would be useful for a hypothetical `IPanel` (a possibly named/framed group within a larger UI).

By formalizing this interface, new classes would be more likely to follow it.

PR #1016, opened 2021-08-16
Add `LayoutWidget` class and use it for newer widget classes

corranwebster:
This adds infrastructure below `IWidget` for widgets that have opinions about sizing; in particular should they be allowed to expand, what weight should they have compared to other widgets when expanding, and are there bounds on how big they can get?  This is useful for widgets which are embedded in layouts and containers.

The PR provides an interface for this, an implementation for both toolkits, and makes this a superclass for `IField` and `IDataViewWidget` for concreteness.  Other widgets will be added tot he system in a separate PR.

This PR is working towards #738 (in that `LayoutWidget` instances are the contents of `IContainers`) and a general layout system for Pyface.  It also is a requirement for being able to directly use Pyface widgets in TraitsUI editors without needing any toolkit-specific code.

### DIFF
--- a/pyface/data_view/i_data_view_widget.py
+++ b/pyface/data_view/i_data_view_widget.py
@@ -228,9 +228,7 @@ class MDataViewWidget(HasStrictTraits):
         This method should create the control and assign it to the
         :py:attr:``control`` trait.
         """
-        self.control = self._create_control(self.parent)
-        self._initialize_control()
-        self._add_event_listeners()
+        super()._create()
 
         self.show(self.visible)
         self.enable(self.enabled)
@@ -239,6 +237,7 @@ class MDataViewWidget(HasStrictTraits):
         """ Initializes the toolkit specific control.
         """
         logger.debug('Initializing DataViewWidget')
+        super()._initialize_control()
         self._set_control_header_visible(self.header_visible)
         self._set_control_selection_mode(self.selection_mode)
         self._set_control_selection_type(self.selection_type)
--- a/pyface/i_layout_widget.py
+++ b/pyface/i_layout_widget.py
@@ -158,7 +158,7 @@ class MLayoutWidget(HasTraits):
         """
         raise NotImplementedError()
 
-    def _set_control_stretch(self, size_policy):
+    def _set_control_stretch(self, stretch):
         """ Set the stretch factor of the control.
 
         Toolkit implementations will need to override this method.
--- a/pyface/i_layout_widget.py
+++ b/pyface/i_layout_widget.py
@@ -1,0 +1,189 @@
+# (C) Copyright 2005-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+from traits.api import HasTraits, Int, Tuple, observe
+
+from pyface.i_layout_item import ILayoutItem, Size, SizePolicy
+from pyface.i_widget import IWidget
+
+
+class ILayoutWidget(IWidget, ILayoutItem):
+    """ Interface for widgets that can participate in layout.
+
+    Most widgets implement ILayoutWidget, but widgets like top-level windows,
+    menus, toolbars, etc. do not.
+    """
+    pass
+
+
+class MLayoutWidget(HasTraits):
+    """ A mixin for Widgets that can participate in layouts.
+
+    Most widgets implement ILayoutWidget, but widgets like top-level windows,
+    menus, toolbars, etc. do not.
+    """
+
+    #: The minimum size that the widget can take.
+    minimum_size = Size
+
+    #: The maximum size that the widget can take.
+    maximum_size = Size
+
+    #: Weight factor used to distribute extra space between widgets.
+    stretch = Tuple(Int, Int)
+
+    #: How the widget should behave when more space is available.
+    size_policy = Tuple(SizePolicy, SizePolicy)
+
+    def _create(self):
+        """ Create and initialize the toolkit control. """
+        # XXX remove once #1013 is merged.
+        self.control = self._create_control(self.parent)
+        self._initialize_control()
+        self._add_event_listeners()
+
+    def _initialize_control(self):
+        """ Initialize the toolkit control. """
+        self._set_control_minimum_size(self.minimum_size)
+        self._set_control_maximum_size(self.maximum_size)
+        self._set_control_stretch(self.stretch)
+        self._set_control_size_policy(self.size_policy)
+
+    def _add_event_listeners(self):
+        """ Add trait observers and toolkit binding. """
+        super()._add_event_listeners()
+        self.observe(
+            self._minimum_size_updated,
+            "minimum_size",
+            dispatch="ui",
+        )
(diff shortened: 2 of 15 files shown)


rahulporuri:
@corranwebster CI is failing on this PR. Can you take a look at the test failures?

corranwebster:
Arg. Didn't test with PySide2.  I know what the issue likely is (it's todo with me trying to be clever and use Python `enum.Enum` for constants.

corranwebster:
Had a deeper look yesterday, and this looks like a more fundamental issue around PySide2 - for some reason setting the size policy is failing to actually change the size policy in many instances.

corranwebster:
Looks like it was an issue with enums: while PyQt5 will accept arguments which are integer subclasses that match the constant value (like you get from `IntEnum`), PySide2 appears to require actual constant values.  This is resolved by using an `Enum` and asking for the `.value`.

Hopefully CI should now pass.

pyface/data_view/i_data_view_widget.py
```diff
@@ -228,9 +228,7 @@ def _create(self):
         This method should create the control and assign it to the
         :py:attr:``control`` trait.
         """
-        self.control = self._create_control(self.parent)
-        self._initialize_control()
-        self._add_event_listeners()
+        super()._create()
```
rahulporuri, line 231: ~does this PR depend on #1013 ?~ It looks like this PR makes some of the changes that are included in #1013. Is that intentional? Do the changes in this PR depend on #1013 ?

pyface/i_layout_widget.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def _get_control_maximum_size(self):
+        """ Get the maximum size of the control.
+
+        Toolkit implementations will need to override this method.
+        This method is only used for testing.
+        """
+        raise NotImplementedError()
+
+    def _set_control_stretch(self, size_policy):
```
rahulporuri, line 161:
```suggestion
    def _set_control_stretch(self, stretch):
```

pyface/testing/widget_mixin.py
```diff
@@ -0,0 +1,55 @@
...
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+
+from traits.testing.api import UnittestTools
+
+from pyface.action.api import Action, MenuManager
```
rahulporuri, line 14: unused imports

```suggestion
```

pyface/ui/qt4/layout_widget.py
```diff
@@ -0,0 +1,133 @@
...
+
+from traits.api import provides
+
+from pyface.qt import QtGui
+from pyface.i_layout_item import DEFAULT_SIZE, SizePolicy
+from pyface.i_layout_widget import ILayoutWidget, MLayoutWidget
+from pyface.ui.qt4.widget import Widget
+
+
+#: Maximum widget size (some versions of PyQt don't export it)
```
rahulporuri, line 21: PyQt or PyQt5? We don't need to add new code with support for PyQt as we've dropped support for it.

pyface/ui/qt4/layout_widget.py
```diff
@@ -0,0 +1,133 @@
...
+    fixed = QtGui.QSizePolicy.Fixed
+    preferred = QtGui.QSizePolicy.Preferred
+    expand = QtGui.QSizePolicy.Expanding
+
+
+@provides(ILayoutWidget)
+class LayoutWidget(MLayoutWidget, Widget):
+    """ A widget which can participate as part of a layout. """
+
+    def _set_control_minimum_size(self, size):
```
rahulporuri, line 40: it's not clear from the `MLayoutWidget` documentation that `size` needs to be a tuple where the first and second elements are width and height - although that might be the default behavior for most/all toolkits.

pyface/ui/wx/fields/field.py
```diff
@@ -35,28 +35,6 @@ class Field(MField, Widget):
...
 
-    # ------------------------------------------------------------------------
-    # IField interface
-    # ------------------------------------------------------------------------
-
-    def _initialize_control(self):
-        """ Perform any toolkit-specific initialization for the control. """
-        self.control.SetToolTip(self.tooltip)
-        self.control.Enable(self.enabled)
-        self.control.Show(self.visible)
```
rahulporuri, line 46: it's not clear to me why `Field.initialize_control` can be removed here nor do i understand why `Field._create` and `Field.destroy` can be removed either.

pyface/ui/wx/layout_widget.py
```diff
@@ -0,0 +1,121 @@
...
+        """ Get the size policy of the control
+
+        In Wx the size policy is set at layout time and can't be obtained from
+        the widget itself.  As a result, this method simply returns the
+        current trait value.  This method is only used for testing.
+        """
+        return self.size_policy
+
+
+def _size_to_wx_size(size: t.Tuple[int, int]) -> wx.Size:
```
rahulporuri, line 89: Any specific reason why you made use of `typing` here? We haven't used it in other parts of the codebase or in other ETS libraries.

rahulporuri (approved):
Mostly LGTM with a number of suggestions and a few questions.

pyface/data_view/i_data_view_widget.py
corranwebster: No, it doesn't depend on #1013 because it adds a new class between these classes and the base `Widget` class which _doesn't_ call `super`.

Now that #1013 is merged we will want to add that `super` call, but it can safely be in another PR if needs be (or if I make any modifications to this I will add it in to this PR).

pyface/ui/qt4/layout_widget.py
corranwebster: It's some but not all versions of PyQt5 - I don't remember which.

pyface/ui/wx/fields/field.py
corranwebster: These are now handled without any toolkit-specific code in the base `MField` class (via `_set_control_tooltip`, etc.)

The `_create` and `_destroy` aren't needed because their only purpose was to call `_add_event_listeners` and `_remove_event_listeners` which is now done in the `_create` and `_destroy` methods of `MLayoutWidget`.

pyface/ui/wx/layout_widget.py
corranwebster: It was an experiment on my part to see if it adds anything to use explicitly typed code and to play with patterns for doing so. This seemed like a fairly harmless corner to try it in, and one where the constraints are fairly clear.

Typing seems to be catching on, so I think we may well end up with our codebases being typed in the 2-3 year timeframe, so experiments are valuable.

If you feel strongly it should go, I will remove.

pyface/ui/qt4/layout_widget.py
corranwebster: Not _too_ concerned about that as this API is only used by toolkit implementers.  End-users will use the `minimum_size` trait which is set to the equivalent to `Tuple(Int, Int)`.

pyface/data_view/i_data_view_widget.py
corranwebster: I've merged in main and resolved all of this.

pyface/ui/wx/layout_widget.py
corranwebster: I've removed it.  At some point I may try with a more self-contained library.

PR #1018, opened 2021-09-08
More `LayoutWidget` subclasses

corranwebster:
This PR changes the inheritance of a collection of existing widget classes so that they inherit from `LayoutWidget` rather than directly from `Widget` where appropriate.  This adds capabilities but should be completely backwards compatible.

These were kept out of #1016 to simplify the review of that PR.


rahulporuri (approved):
LGTM

PR #1019, opened 2021-09-08
Widget tooltips

corranwebster:
With #1013 merged, it's now possible to move tooltip code from Fields to Widgets, so all widgets can have a tooltip.  This is PR does a fairly straightforward shift of the code between the two classes, and updates tests and documentation appropriately.

Code that uses tooltips on Fields should notice no change.

PR #1020, opened 2021-09-09
Widget context menus

corranwebster:
This is doing essentially the same as #1019 but for context menus, so that now any widget can have a context menu attached to it.

One consequence of this is that `Field` and the testing code for field have become very thin with no toolkit-specific code at all.  This PR is leaving them in that state, but I will open an issue to either do a clean-up or add some meat to them.

### DIFF
--- a/pyface/i_widget.py
+++ b/pyface/i_widget.py
@@ -11,7 +11,7 @@
 """ The base interface for all pyface widgets. """
 
 
-from traits.api import Any, Bool, HasTraits, Interface, Str
+from traits.api import Any, Bool, HasTraits, Interface, Instance, Str
 
 
 class IWidget(Interface):
@@ -34,6 +34,9 @@ class IWidget(Interface):
 
     #: A tooltip for the widget.
     tooltip = Str()
+
+    #: An optional context menu for the widget.
+    context_menu = Instance("pyface.action.menu_manager.MenuManager")
 
     # ------------------------------------------------------------------------
     # 'IWidget' interface.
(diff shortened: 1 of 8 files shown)


rahulporuri:
@corranwebster CI is failing on this PR as well on the wx toolkit

corranwebster:
Looks like a genuine conflict in the DataView code.

pyface/i_widget.py
```diff
@@ -35,6 +35,9 @@ class IWidget(Interface):
     #: A tooltip for the widget.
     tooltip = Str()
 
+    #: An optional context menu for the widget.
+    context_menu = Instance("pyface.action.menu_manager.MenuManager")
```
rahulporuri, line 39: What about the `show_context_menu` method that is defined on the `IField` interface? Update : It looks like an undefined/unused part of the `IField` interface which we're choosing not to carry over.

pyface/ui/wx/fields/field.py
```diff
@@ -28,25 +26,3 @@ class Field(MField, LayoutWidget):
...
-    # ------------------------------------------------------------------------
-
-    def _observe_control_context_menu(self, remove=False):
-        """ Toolkit specific method to change the control menu observer. """
-        if remove:
-            self.control.Unbind(
-                wx.EVT_CONTEXT_MENU, handler=self._handle_context_menu
-            )
-        else:
-            self.control.Bind(wx.EVT_CONTEXT_MENU, self._handle_context_menu)
```
rahulporuri, line 46: looks like there was a bug here which we coincidentally fixed

rahulporuri (approved):
LGTM with a few comments

pyface/i_widget.py
corranwebster: Yes, I realized it wasn't hooked up to anything, so no reason to carry it over.

PR #1027, opened 2021-09-28
Add a flake8 command to etstool.py

corranwebster:
This PR also adds a setup.cfg which makes flake8 run cleanly.  There are
things in the clean list that are likely errors and need to be fixed.

This doesn't add flake8 to CI, that should be a separate PR.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -99,6 +99,8 @@ dependencies = {
     "pygments",
     "coverage",
     "pillow",
+    "flake8",
+    "flake8_ets",
 }
 
 source_dependencies = {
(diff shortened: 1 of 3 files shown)


etstool.py
```diff
@@ -462,6 +464,30 @@ def test_all(edm):
...
+    '--strict/--not-strict',
+    default=False,
+    help="Use strict configuration for flake8 [default: --not-strict]",
+)
+def flake8(edm, runtime, toolkit, environment, strict):
+    """Run a flake8 check in a given environment."""
+    parameters = get_parameters(edm, runtime, toolkit, environment)
+    config = ""
+    if strict:
+        config = "--config=flake8_strict.cfg "
```
rahulporuri, line 484: I'm not sure what this file is. should this be `setup.cfg`?

Also, I don't fully understand what the point of "strict" vs "non-strict" options for flake8 is - don't we always want to use "strict"?

corranwebster: This is a copy-pasta from TraitsUI - the idea was strict is something like what we are aiming for (and so new code should pass strict), but there are a lot of issues that need fixing before old code could pass the strict version.

corranwebster: Obviously I'll fix one way or another.

rahulporuri (approved):
LGTM for now. We'll need to come back and make changes anyway as we slowly address the flake8 issues.

PR #1034, opened 2021-11-08
Fix a broken local import

### LINKED ISSUE #1033: workbench_window.py missing relative path in import
Simple one... Method _editor_manager_default(), is missing relative path in import statement.  Should be:

from .editor_manager import EditorManager

rahulporuri:
fixes #1033

This PR fixes a broken local import in the workbench machinery.

### DIFF
--- a/pyface/workbench/workbench_window.py
+++ b/pyface/workbench/workbench_window.py
@@ -257,7 +257,7 @@ class WorkbenchWindow(ApplicationWindow):
     def _editor_manager_default(self):
         """ Trait initializer. """
 
-        from editor_manager import EditorManager
+        from .editor_manager import EditorManager
 
         return EditorManager(window=self)
 


mdickinson (approved):
LGTM. Is this an isolated case? Are there tools we could use that would identify this (and other) broken imports?

rahulporuri:
> LGTM. Is this an isolated case? Are there tools we could use that would identify this (and other) broken imports?

I don't think this is an isolated case. I'm afraid there might be more such broken imports, which we aren't catching because the testsuite isn't that good and because we arent using the workbench code internally.

I will try to find more such broken imports and see if any useful tools exist (I'd be surprised if i find none)

rahulporuri:
> I will try to find more such broken imports

For now, I manually went over the workbench code in pyface and found no other broken imports. I just manually inspected the workbench modules, it didn't take long.

PR #1038, opened 2021-11-30
Fix SplitEditorAreaPane tab label and tooltip updates

### LINKED ISSUE #523: Qt SplitEditorAreaPane won't update an editor's name if it's not in the currently focused splitter
The title is a bit long, so I'll try to give detailed reproduction steps.

pyface version: 6.1.2
PyQt version: 4.12.1
Qt version: 4.8.7

1. Create a `Task` with a `SplitEditorAreaPane` as the central pane
2. Add an `Editor` to the pane which updates it's `name` trait in response to some external event (make sure this works before proceeding).
3. Add a second editor
4. Split the editor pane and drag the editor from (2) to the new tab area
5. Give the editor from (3) focus
6. Trigger editor from (2) so that its `name` changes.
7. Notice that the tab name didn't change.
8. Drag the editor from (2) back to the original tab. Its name should sync up now.

I dug into the `SplitEditorAreaPane` and I think I see the problem:
https://github.com/enthought/pyface/blob/90a96146988f8100ea954e363c7fe101c8ec5f72/pyface/ui/qt4/tasks/split_editor_area_pane.py#L309-L312

Because the `active_tabwidget` attribute changes when the currently focused editor changes, you can get in a situation where the `name` of an editor changes, but you get a `-1` for the index of the editor and then `setTabText` just ignores the call. Documentation: [QTabWidget::indexOf](https://doc.qt.io/archives/qt-4.8/qtabwidget.html#indexOf), [QTabWidget::setTabText](https://doc.qt.io/archives/qt-4.8/qtabwidget.html#setTabText)

corranwebster:
Ensure editor tab label and tooltip update correctly when not active; fixes #523.

Includes a small refactor to pull out common code that finds an editor's tabwidget and index in that tabwidget.

### DIFF
--- a/pyface/ui/qt4/tasks/split_editor_area_pane.py
+++ b/pyface/ui/qt4/tasks/split_editor_area_pane.py
@@ -151,7 +151,7 @@ class SplitEditorAreaPane(TaskPane, MEditorAreaPane):
     def activate_editor(self, editor):
         """ Activates the specified editor in the pane.
         """
-        active_tabwidget = editor.control.parent().parent()
+        active_tabwidget = self._get_editor_tabwidget(editor)
         active_tabwidget.setCurrentWidget(editor.control)
         self.active_tabwidget = active_tabwidget
         editor_widget = editor.control.parent()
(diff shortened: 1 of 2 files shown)


jwiggins (approved):
LGTM

PR #1041, opened 2021-11-30
Remove usages of `Trait()` in codebase

### LINKED ISSUE #1039: Remove uses of `Trait()`
Where possible, we want to remove all uses of the old `Trait()` trait constructor and replace with equivalent modern usage.

corranwebster:
Fixes #1039 

This replaces the pattern `Trait(default, None, TraitType)` with `Union(None, TraitType, default_value=default)`, which I think is equivalent.

### DIFF
--- a/pyface/ui/wx/grid/grid.py
+++ b/pyface/ui/wx/grid/grid.py
@@ -41,8 +41,8 @@ from traits.api import (
     Font,
     Instance,
     Int,
-    Trait,
     Undefined,
+    Union,
 )
 from pyface.wx.drag_and_drop import (
     PythonDropSource,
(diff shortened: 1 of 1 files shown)


pyface/ui/wx/grid/grid.py
```diff
@@ -103,7 +105,6 @@ class Grid(LayoutWidget):
     default_cell_bg_color = Color("white")
 
     # The default background color to use for read-only cells
-    # default_cell_read_only_color = Trait(Color("linen"), None, Color)
     default_cell_read_only_color = Color(wx.Colour(248, 247, 241))
```
stevenjkern, line 108: Should this one be structured like the other changes in this PR?
```suggestion
    default_cell_read_only_color = Union(None, Color,
                                         default_value=wx.Colour(248, 247, 241))
```

corranwebster: I thought about that, but since I removed a commented line and the non-commented code is working, I didn't want to change it and introduce a change in behaviour.

The change here would allow `None` as a value; and I can see that `None` probably isn't good for a trait which is supposed to be holding a default color value.  So I _think_ we're good without further changes.

stevenjkern (approved):
LGTM 👍🏻

PR #1044, opened 2021-11-30
Fix dock pane undocking

### LINKED ISSUE #1028: Undocking dock panes fails with PyQt5
When running the pyface example examples/tasks/basic/run.py and on-docking the Script Browser pane...

image

the application crashes due to a KeyError in the qt4 toolkit:

image

Apparently when undocking the area value gets 0, which is not included in INVERSE_AREA_MAP.

A quick workaround that makes the example working fine is:
```
def _receive_dock_area(self, area):
        with self._signal_context():
            if int(area) in INVERSE_AREA_MAP:
                self.dock_area = INVERSE_AREA_MAP[int(area)]
```

Using packages:
apptools==5.1.0
configobj==5.0.6
envisage==6.0.1
importlib-metadata==4.8.1
importlib-resources==5.2.2
numpy==1.21.2
Pillow==8.3.2
pyface==7.3.0
Pygments==2.10.0
PyQt5==5.15.4
PyQt5-Qt5==5.15.2
PyQt5-sip==12.9.0
scipy==1.7.1
shiboken2==5.15.2
six==1.16.0
traits==6.2.0
traitsui==7.2.1
zipp==3.6.0

System: Win10 64-bit

corranwebster:
Fixes #1028 

I am a little concerned there may be issues with serialization of undocked window state that are not fixed by this.


pyface/ui/qt4/tasks/dock_pane.py
```diff
@@ -185,7 +185,8 @@ def _set_visible(self, event):
 
     def _receive_dock_area(self, area):
         with self._signal_context():
-            self.dock_area = INVERSE_AREA_MAP[int(area)]
+            if int(area) not in INVERSE_AREA_MAP:
+                self.dock_area = INVERSE_AREA_MAP[int(area)]
```
rahulporuri, line 189: I'm confused by this change. If `int(area) not in INVERSE_AREA_MAP` is True, then the key `int(area) doesn't exist in the dict `INVERSE_AREA_MAP` and ` `INVERSE_AREA_MAP[int(area)]` should raise an exception. Right?

corranwebster: Yep, that's an error - apparently I can't copy code correctly :)

Not sure why it didn't crash horribly when I was testing.

```suggestion
            if int(area) in INVERSE_AREA_MAP:
                self.dock_area = INVERSE_AREA_MAP[int(area)]
```

pyface/ui/qt4/tasks/tests/test_dock_pane.py
```diff
@@ -75,3 +76,28 @@ def check_panes_and_exit(app_event):
         self.assertTrue(tool_attributes)
         for attr in tool_attributes:
             self.assertTrue(attr)
+
+    def test_dock_windows_undock(self):
```
rahulporuri, line 80: i'm not sure why this test wasn't failing earlier with the in/not in problem but from the looks of it, this test is manually undocking a dock pane and then checks the dock area of the dock panes in the app.

I'm not sure why by default undocking a dockpane means that it's `left` widget area but that's what we seem to be asserting at the end. Do we need to be asserting this at all? Do we care what this value is?

rahulporuri (approved):
Mostly LGTM with a couple of comments

pyface/ui/qt4/tasks/tests/test_dock_pane.py
corranwebster: I've manually verified that this doesn't crash, so I think the state is good.

The test is testing the dockpane area because it should not change when it is undocked (so yes, we do care about the value) - this was a deliberate choice as one option was to add an "undocked" dockpane area, but the behaviour (particularly when restoring from saved state) was unclear.

PR #1046, opened 2021-12-02
Clean up example imports

### LINKED ISSUE #1045: Clean up wizard examples
These have unused imports and are doing hacky things with `sys.path`.  These should be cleaned up.

corranwebster:
This PR simplifies the imports of many of the older examples, including removing the code that messed with `sys.path` in a number of cases.

Fixes #1045.


examples/node_tree.py
```diff
@@ -10,20 +10,16 @@
...
 
+import os
 
-import os, sys
-
-# Put the Enthought library on the Python path.
-sys.path.append(os.path.abspath(r"..\..\.."))
-
-
-from pyface.api import GUI, PythonShell, SplitApplicationWindow
```
rahulporuri, line 20: feel free to ignore - I think this import can stay where it is - the examples live outside the pyface package so can treat the pyface import as just a part of the ETS import group.

rahulporuri (approved):
LGTM with one minor comment.

examples/node_tree.py
corranwebster: True, but I'll leave it for now and add an issue, because I suspect that there are more cases.

PR #1057, opened 2021-12-17
Basic pyside6 support

corranwebster:
This makes a minimal set of changes to allow the test suite and most examples to run under PySide6.


corranwebster:
Closing, as it seems to have pulled in some experimental code.

pyface/qt/QtOpenGL.py
```diff
@@ -16,7 +16,7 @@
     from PyQt5.QtOpenGL import *
 
 elif qt_api == "pyqt6":
-    from PyQt6.QtOpenGL import *
+    from PyQt6.QtGui import *
```
rahulporuri, line 19: Is this change necessary? I still see a `PyQt6.QtOpenGL` subpackage.

rahulporuri (approved):
LGTM with one question

pyface/qt/QtOpenGL.py
corranwebster: That looks like an error - there may be something we need to do for Qt5 and 6 because the content of QtOpenGL is different between the two - but it can be in a different PR if needed.

corranwebster:
```suggestion
    from PyQt6.QtOpenGL import *
```

PR #1064, opened 2021-12-22
Fixes for wx color dialog

### LINKED ISSUE #1061: Wx ColorDialog doesn't change value
The Wx Color dialog doesn't change the value to the selected color on successful close on MacOS.

corranwebster:
Also add missing copyright header on color parser test file.

Fixes #1061

### DIFF
--- a/pyface/ui/wx/color_dialog.py
+++ b/pyface/ui/wx/color_dialog.py
@@ -50,10 +50,9 @@ class ColorDialog(Dialog):
     # ------------------------------------------------------------------------
 
     def close(self):
-        if self.control.GetReturnCode() == wx.ID_OK:
-            colour_data = self.control.GetColourData()
-            wx_colour = colour_data.GetColour()
-            self.color = Color.from_toolkit(wx_colour)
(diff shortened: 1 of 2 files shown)


pyface/ui/wx/color_dialog.py
```diff
@@ -50,10 +50,9 @@ def _create_contents(self, parent):
     # ------------------------------------------------------------------------
 
     def close(self):
-        if self.control.GetReturnCode() == wx.ID_OK:
```
rahulporuri, line 53: I don't understand why removing this check gives the expected behavior on Mac OS. This feels counter-intuitive.

corranwebster: It turns out it was either not returning, or had not set, the return code at the point where that's being run.  I was seeing the same thing with the Font dialog (I'd click "OK" but the value returned was not "ID_OK").  Don't know if it's a Wx bug or what, but there's no real problem with always setting the color value.

rahulporuri (approved):

PR #1076, opened 2022-01-12
Move schemas and related code out of Tasks

### LINKED ISSUE #658: Make TaskActionManagerBuilder._create_action_manager_recurse reusable without reference to a task
There exists code outside of pyface trying to reuse `TaskActionManagerBuilder._create_action_manager_recurse` in order to create an action manager without reference to a task. (Note: the method is private and therefore such usage outside of the library is strongly discouraged, and wouldn't have been allowed in some other programming languages).

These external usage look more or less like pyface's own `TaskActionManagerBuilder.create_action_manager`, except that the `additions_map` is replaced with something else:
https://github.com/enthought/pyface/blob/f60905affb95cc0024c1a0514da4f4d389e5a533/pyface/tasks/action/task_action_manager_builder.py#L43-L53

A fair amount of unpacking, grouping and ordering logic used by `_create_action_manager_recurse` do not actually use information from the `TaskActionManagerBuilder` instance. This function can be extracted and maintained external to the builder, and be made public.

corranwebster:
Schemas are generic, this PR moves the Schema code out of pyface.tasks and into pyface.action, making it available for use outside of tasks.  Additionally a version of `ActionManagerBuilder` is included to provide a Tasks-independent way of assembling the Schemas into action managers, which fixes #658.

Additionally, the code is slightly modified so that it does not import toolkit code as a side-effect.

The old locations are kept so imports should still work.  They will be removed at some yet-to-be-determined future point.


corranwebster:
Note that while there is a more immediate need of this change, this opens up the possibility of using Schemas in TraitsUI instead of directly creating pyface.action objects.

pyface/action/schema/schema.py
```diff
@@ -0,0 +1,230 @@
...
+    Instance("pyface.action.group.Group"),
+    Instance("pyface.action.menu_manager.MenuManager"),
+    Instance("pyface.action.schema.schema.GroupSchema"),
+    Instance("pyface.action.schema.schema.MenuSchema"),
+    Instance("pyface.action.schema.schema.Schema"),
+)
+
+
+class Schema(HasTraits):
+    """ The abstract base class for all Tasks action schemas.
```
aaronayres35, line 50:
```suggestion
    """ The abstract base class for all action schemas.
```

pyface/action/schema/tests/test_topological_sort.py
```diff
@@ -0,0 +1,103 @@
...
+        result, has_cycles = topological_sort(pairs)
+        self.assertTrue(not has_cycles)
+        self.assertEqual(result, [1, 2, 3, 4, 5, 6])
+
+    def test_topological_sort_3(self):
+        """ Does cycle detection work?
+        """
+        pairs = [(1, 2), (2, 3), (3, 1)]
+        result, has_cycles = topological_sort(pairs)
+        self.assertTrue(has_cycles)
```
aaronayres35, line 103: do we need to keep the old test module `pyface/tasks/tests/test_topological_sort.py` around or can it be removed?

pyface/tasks/action/api.py
```diff
@@ -58,6 +40,30 @@
 - :class:`~.TaskWindowToggleGroup`
 - :class:`~.CreateTaskWindowAction`
 
+ActionSchemas and aliases
+-------------------------
+
+Import of these objects from pyface.tasks.actions.api is deprecated and will
+be removed in a future Pyface release.  Instead use pyface.action.schema.api.
```
aaronayres35, line 47: Maybe we want to raise the deprecation warnings on this PR?  Or it may be too soon and comments are fine for now - but if we don't raise the warnings it would be good to open some GH issues so these are less likely to be forgotten

pyface/tasks/action/task_action_manager_builder.py
```diff
@@ -8,50 +8,38 @@
...
-            if addition.path:
-                additions_map[addition.path].append(addition)
-
-        manager = self._create_action_manager_recurse(schema, additions_map)
-        manager.controller = self.controller
-        return manager
-
     def create_menu_bar_manager(self):
         """ Create a menu bar manager from the task's menu bar schema and
             additions.
```
aaronayres35, line 45: nit: in `create_tool_bar_managers` we go through `self.task.extra_actions` instead of `self.additions`.  They are equivalent, but using `self.additions` feels more consistent

pyface/action/schema/action_manager_builder.py
```diff
@@ -0,0 +1,277 @@
...
+from .schema import Schema
+from .schema_addition import SchemaAddition
+from ._topological_sort import before_after_sort
+
+
+class ActionManagerBuilder(HasTraits):
+    """ Builds action managers from schemas, merging schema additions.
+    """
+
+    #: Any schema additions to apply to the schema being built.
```
aaronayres35, line 31: This trait doc needs an update

pyface/action/schema/action_manager_builder.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def _group_items_by_id(self, items):
+        """ Group a list of action items by their ID.
+
+        Action items are Schemas and Groups, MenuManagers, etc.
+
+        Return a dictionary {item_id: list_of_items}, and a list containing
+        all the ids ordered by their appearance in the `all_items` list. The
+        ordered IDs are used as a replacement for an ordered dictionary, to
+        keep compatibility with Python <2.7 .
```
aaronayres35, line 125: not relevant to this PR, but doing special things to keep py 2.7 compatibility could be removed / refactored at some point

pyface/tasks/tests/test_action_manager_builder.py
```diff
@@ -8,6 +8,7 @@
 #
 # Thanks for using Enthought open source!
 
+from contextlib import contextmanager
 import unittest
```
aaronayres35, line 14: imports from pyface.tasks.action.api below mayb should get updated to use new pyface.actions.schema.api

pyface/tasks/tests/test_action_manager_builder.py
```diff
@@ -8,6 +8,7 @@
 #
```
aaronayres35, line 8: possibly rename this module `test_task_action_manager_builder.py`?

pyface/action/schema/tests/test_action_manager_builder.py
```diff
@@ -0,0 +1,443 @@
...
+    Group,
+    MenuManager,
+    MenuBarManager,
+)
+from pyface.tasks.action.api import (
+    GroupSchema,
+    MenuSchema,
+    MenuBarSchema,
+    SchemaAddition,
+)
```
aaronayres35, line 28:
```suggestion
from pyface.action.schema.api import (
    GroupSchema,
    MenuSchema,
    MenuBarSchema,
    SchemaAddition,
)
```

We probably don't want a `pyface.tasks` import here

aaronayres35 (approved):
LGTM, just a few comments

pyface/tasks/action/api.py
corranwebster: My thought was it's too soon - we don't want to issue warnings until we've had a chance to fix upstream packages otherwise people will get warnings that they can't do anything about.

pyface/action/schema/action_manager_builder.py
corranwebster: Yes,I noticed this and decided not to mess with it in this PR.

pyface/tasks/action/task_action_manager_builder.py
corranwebster: It should be changed because if a subclass or other user of the class messes with how the `additions` are built then we will get inconsistent results.

pyface/tasks/tests/test_action_manager_builder.py
corranwebster: Yep - missed this.  There are more places that the imports need to be changed (eg. examples, docs).

pyface/tasks/action/api.py
aaronayres35: Makes sense! I've opened #1077

corranwebster:
> Only thing I notice is that we still have a`Instance("pyface.tasks.action.schema_addition.SchemaAddition")` in `tasks_application.py` which could instead be `Instance("pyface.action.schema.schema_addition.SchemaAddition")`

Good catch - my search only looked for imports.

Corrected, and I will merge after this.

PR #1078, opened 2022-01-14
Expose toolkit app, icon and name

corranwebster:
This adds an `application` trait to the `GUI` and uses that to access the application icon and display name to improve appearance/branding of the application (rather than using Python defaults).  This is particularly aimed at applications run from a script using a Python installation, rather than frozen applications


corranwebster:
Needs tests.

PR #1079, opened 2022-01-17
Use observers rather than listeners in TaskActionManagerBuilder

corranwebster:
Does what it says.


rahulporuri (approved):
Looks innocent to me.

PR #1085, opened 2022-01-26
Skip PySide6 on MacOS for now

corranwebster:
Due to some change in the last week or so, the PyPI PySide6 wheels will no longer import on MacOS, Python 3.6.  Skipping for now.


corranwebster:
I think we have a solution for this for now.

PR #1088, opened 2022-01-27
Support Python 3.8 runtimes in etstool

corranwebster:
This is a fairly significant change because the branching if statements about which things to install how in which environment was getting confusing.  This adds a lightweight package class that for a particular runtime and platform tells you how to install the package (edm vs pip and the appropriate requirement/arguments).

This permits cleaner expression of what we actually need.

If this is good, we might copy to other etstool.py scripts.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -180,10 +180,6 @@ dependencies = [
     }),
     Package("numpy"),
     Package("pygments"),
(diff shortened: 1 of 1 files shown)


corranwebster:
Closing in favour of #1203

PR #1089, opened 2022-01-27
Use colors from widget palette rather than hard-coded

corranwebster:
This means that alternate styles/dark mode, etc. should work better for these widgets.

This also changes the default for the code editor to not highlight the current line, as that color should probably have some thought put into it if it is used.  This is a technically non-backward compatible change.

### DIFF
--- a/pyface/ui/qt4/code_editor/code_widget.py
+++ b/pyface/ui/qt4/code_editor/code_widget.py
@@ -31,7 +31,7 @@ class CodeWidget(QtGui.QPlainTextEdit):
     focus_lost = QtCore.Signal()
 
     def __init__(
-        self, parent, should_highlight_current_line=True, font=None, lexer=None
+        self, parent, should_highlight_current_line=False, font=None, lexer=None
     ):
         super().__init__(parent)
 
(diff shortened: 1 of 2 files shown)


pyface/ui/qt4/code_editor/code_widget.py
```diff
@@ -31,7 +31,7 @@ class CodeWidget(QtGui.QPlainTextEdit):
     focus_lost = QtCore.Signal()
 
     def __init__(
-        self, parent, should_highlight_current_line=True, font=None, lexer=None
+        self, parent, should_highlight_current_line=False, font=None, lexer=None
```
flongford, line 34: Just FYI - the code line highlight no longer shows up when running the `examples/python_editor/python_editor_application.py` app. I presume this is since the Qt5 `AdvancedCodeWidget` class uses the default value of `should_highlight_current_line` when instantiating the `AdvancedCodeWidget.code` attribute:

https://github.com/enthought/pyface/blob/8ad15a8a64f095c48c017575bc97772547b321e5/pyface/ui/qt4/code_editor/code_widget.py#L563

Is this intentional / expected?

flongford: Ah, I guess from 
> This also changes the default for the code editor to not highlight the current line, as that color should probably have some thought put into it if it is used.
It would be expected?

corranwebster: Those whole set of classes need a refactor, because a lot of stuff doesn't surface up to the `PythonEditor` widget.

Some of this is captured as an item here: https://github.com/enthought/pyface/issues/551

I'll add a more specific issue.

corranwebster: Added #1094

flongford (approved):
LGTM!

PR #1093, opened 2022-01-28
Use a weakref to the window that the EventFilter is attached to

### LINKED ISSUE #1081: Failure to close dialog causing segfaults
On an internal Enthought project we are seeing segfaults on test runs for a dialog when the dialog did not close properly.

Crash was in attribute look up of the window event filter, here: https://github.com/enthought/pyface/blob/a44a5592f6df50101cd7449c4f0bfe4f41a7d7f5/pyface/ui/qt4/window.py#L203

Will attempt to create a minimal replicating example for this issue.

corranwebster:
Fixes #1081.

### DIFF
--- a/pyface/ui/qt4/window.py
+++ b/pyface/ui/qt4/window.py
@@ -9,6 +9,7 @@
 #
 # Thanks for using Enthought open source!
 
+import weakref
 
 from pyface.qt import QtCore, QtGui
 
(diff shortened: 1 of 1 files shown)


mdickinson (approved):
LGTM, but it would be great to have a test for this.

corranwebster:
Right now a blunt test is that the test suite segfaults in PySide6 without this fix.  I'll think through if there is a way to test this independently without either getting too hung up on implementation details and given the constraints of trying to test around object finalization.

corranwebster:
I've added an issue for testing, as I think it is a larger problem than just this PR: https://github.com/enthought/pyface/issues/1097

PR #1102, opened 2022-02-01
A handful more changelog updates

corranwebster:
These should be the last before the release, unless we find some show-stopper.


aaronayres35 (approved):
LGTM

PR #1105, opened 2022-02-02
Update documentation for Pyface 7.4.0

corranwebster:
This is a straightforward creation of a new 7.4 subdirectory together with replacing the top-level content with the 7.4.0 content.


rahulporuri:
Something is definitely broken in the documentation. The "Standard Dialogs" page was a new addition in this release and clicking on the "Standard Dialogs" page in the index raises a 404. Note that the same thing happens with "Images in Pyface".

Another puzzling thing in the docs is "Next topic" in the "Widgets" page is "Standard Dialogs" whereas in the index page, after "Widgets" we have "Pyface Applications", "Pyface Trait Types" and "Images in Pyface". Interestingly, "Pyface Applications" and "Pyface Trait Types" pages look as expected when accessed directly from the index page.

It looks like the "Testing" page is also new - and is also raising a 404.

corranwebster:
Thanks - I'll look into it

rahulporuri (approved):
:shipit:

PR #1116, opened 2022-02-24
Update documentation for 7.4.1 release

corranwebster:
This overwrites the 7.4.0 documentation, which I think is the correct behaviour.

For most files, this simply updates the date.  The change log and files which reference it are also updated.  There should not be significant changes beyond that.


mdickinson (approved):
LGTM; changes look as expected, and works in manual testing.

PR #1134, opened 2022-06-11
Use the "sphinx-copybutton" extension in documentation

rahulporuri:
This PR does the following: 
- Use the `sphinx_copybutton` extension when building the Sphinx documentation
- Create a `docs` extra for the `pyface` package and add `sphinx-copybutton` as a `docs` dependency along with `enthought-sphinx-theme` and `sphinx`
- Install `sphinx-copybutton` from PyPI when building docs using the `etstool.py` utility.

Ref https://sphinx-copybutton.readthedocs.io/en/latest/

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -410,6 +410,8 @@ def docs(edm, runtime, toolkit, environment):
     commands = [
         "{edm} install -y -e {environment} " + packages,
         "{edm} run -e {environment} -- pip install -r doc-src-requirements.txt --no-dependencies",  # noqa: E501
+        # Temporarily install sphinx-copybutton from PyPI
+        "{edm} run -e {environment} -- pip install sphinx-copybutton"
     ]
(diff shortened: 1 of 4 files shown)


etstool.py
```diff
@@ -410,6 +410,8 @@ def docs(edm, runtime, toolkit, environment):
     commands = [
         "{edm} install -y -e {environment} " + packages,
         "{edm} run -e {environment} -- pip install -r doc-src-requirements.txt --no-dependencies",  # noqa: E501
+        # Temporarily install sphinx-copybutton from PyPI
+        "{edm} run -e {environment} -- pip install sphinx-copybutton"
```
corranwebster, line 414: Perhaps add `sphinx-copybutton` to `doc-src-requirements.txt`?

rahulporuri: done.

PR #1137, opened 2022-06-21
Map menu coordinates to global before showing

### LINKED ISSUE #1124: Double check implementation of pyface widget context menu position
In the course of work on another project, we implemented a context menu for a `QtGui.QWidget` derived class in a way that was very similar to what's done here for the Pyface `Widget` class for Qt: https://github.com/enthought/pyface/blob/55f9dae509904db4871cc2d6c27952efc6db22e0/pyface/ui/qt4/widget.py#L136-L145   

However, in order for the context menu open position to match the mouse right click, we needed to map the position passed to the handler via `... = widget.mapToGlobal(pos)`.  

Without that mapping, the position was not always correct. Note that this is especially apparent on multiple monitor setups when an app is not open on the primary screen.  If (expanded?) on the primary screen, the issue may not be obvious.

This issue is to test, and fix if necessary, the implementation in Pyface (which does not appear to include the necessary local-global mapping).

corranwebster:
Fixes #1124.

### DIFF
--- a/pyface/ui/qt4/widget.py
+++ b/pyface/ui/qt4/widget.py
@@ -140,6 +140,7 @@ class Widget(MWidget, HasTraits):
     def _handle_control_context_menu(self, pos):
         """ Signal handler for displaying context menu. """
         if self.control is not None and self.context_menu is not None:
+            pos = self.control.mapToGlobal(pos)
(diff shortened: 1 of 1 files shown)


mdickinson (approved):
LGTM

Issue #539, opened 2020-06-11
ApplicationWindow isn't cleaning up StatusBarManager


ievacerny:
Qt `ApplicationWindow` tests are leaking `QStatusBar`:
<details>
<summary> ApplicationWindow tests output </summary>

```
QObjects at start of test run: {'QApplication': 2}.
[Wed Jun 10 22:41:07 2020] ( 1/16) test_activate (pyface.tests.test_application_window.TestApplicationWindow) ... ok
[Wed Jun 10 22:41:07 2020] ( 2/16) test_close (pyface.tests.test_application_window.TestApplicationWindow) ... ok
[Wed Jun 10 22:41:08 2020] ( 3/16) test_icon (pyface.tests.test_application_window.TestApplicationWindow) ... ok
[Wed Jun 10 22:41:08 2020] ( 4/16) test_menubar (pyface.tests.test_application_window.TestApplicationWindow) ... ok
test_menubar (pyface.tests.test_application_window.TestApplicationWindow) changed QObject counts. added: {'QAction': 5}.
[Wed Jun 10 22:41:08 2020] ( 5/16) test_open_close (pyface.tests.test_application_window.TestApplicationWindow) ... ok
[Wed Jun 10 22:41:08 2020] ( 6/16) test_position (pyface.tests.test_application_window.TestApplicationWindow) ... ok
[Wed Jun 10 22:41:09 2020] ( 7/16) test_reposition (pyface.tests.test_application_window.TestApplicationWindow) ... ok
[Wed Jun 10 22:41:09 2020] ( 8/16) test_resize (pyface.tests.test_application_window.TestApplicationWindow) ... ok
[Wed Jun 10 22:41:09 2020] ( 9/16) test_retitle (pyface.tests.test_application_window.TestApplicationWindow) ... ok
[Wed Jun 10 22:41:09 2020] (10/16) test_show (pyface.tests.test_application_window.TestApplicationWindow) ... ok
[Wed Jun 10 22:41:09 2020] (11/16) test_size (pyface.tests.test_application_window.TestApplicationWindow) ... ok
[Wed Jun 10 22:41:10 2020] (12/16) test_statusbar (pyface.tests.test_application_window.TestApplicationWindow) ... ok
test_statusbar (pyface.tests.test_application_window.TestApplicationWindow) changed QObject counts. added: {'QStatusBar': 1}.
[...]

PR #1141, opened 2022-07-01
Fix for #539

pyhannes:


corranwebster (approved):
Thanks for the contribution!  This looks good.

PR #1144, opened 2022-07-07
Add splash_screen_duration to GUIApplication class.

jonathanrocher:
Adding a new attribute `splash_screen_duration` to control the duration for which the splash screen is shown upon application start.

### DIFF
--- a/pyface/gui_application.py
+++ b/pyface/gui_application.py
@@ -21,11 +21,13 @@ this is most likely to be a subclass of
 
 
 import logging
+from time import sleep
 
 from traits.api import (
     Bool,
     Callable,
     Instance,
+    Int,
     List,
     ReadOnly,
     Tuple,
@@ -63,6 +65,9 @@ class GUIApplication(Application):
 
     #: The splash screen for the application. No splash screen by default
     splash_screen = Instance(ISplashScreen)
+
+    #: How long to display the splash screen, in seconds. Flashed by default.
+    splash_screen_duration = Int
 
     #: The about dialog for the application.
     about_dialog = Instance(IDialog)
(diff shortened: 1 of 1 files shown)


pyface/gui_application.py
```diff
@@ -167,6 +172,8 @@ def start(self):
             # create the GUI so that the splash screen comes up first thing
             if self.gui is Undefined:
                 self.gui = GUI(splash_screen=self.splash_screen)
+                if self.splash_screen_duration > 0:
+                    sleep(self.splash_screen_duration)
```
corranwebster, line 176: This halts all processing for a number of seconds, and in particular the event loop will not be running so the app will be locked-up, which is generally not a good thing.  At a minimum, this should occasionally be processing events.

pyface/gui_application.py
```diff
@@ -64,6 +66,9 @@ class GUIApplication(Application):
     #: The splash screen for the application. No splash screen by default
     splash_screen = Instance(ISplashScreen)
 
+    #: How long to display the splash screen, in seconds. Flashed by default.
+    splash_screen_duration = Int
```
corranwebster, line 70: This should be a float.

corranwebster (commented):
If I'm guessing what you want, is for the splash screen to persist for at least a minimum amount of time so it is visible to the user.

Probably the right way of doing this is to instead modify `start_event_loop` to do something like
```
    def start_event_loop(self):
        if self._splash_screen is not None:
            self.invoke_after(
                self.splash_screen_duration * 1000,  # milliseconds
                self.self._splash_screen.close,
            )
    ...
```
which avoids locking up the app, and also allows the user to dismiss by clicking on it.

Issue #1117, opened 2022-02-24
Workbench state not persisting in Mayavi


corranwebster:
Reports from the Mayavi team that state of perspectives is not being stored properly:
```
If you run mayavi2, for the first time, the default perspective is OK but if you close it and restart, or change it, or even create a new perspective, you get a terribly messed up perspective with the editor area being very tiny. Am happy to share a screenshot but this is almost trivial to reproduce. Just run mayavi2 with a qt backend twice on the terminal.
```

corranwebster:
I can confirm the behaviour.

corranwebster:
I am seeing `QBytes` objects in the memento state - possible issue there?

corranwebster:
I think the core issue here may be that Mayavi doesn't implement an `EditorManager`, and so there is no code restoring the content or the editor tabs, and so they get squashed as they don't ask for any room.

PR #1149, opened 2022-07-11
Fixes from investigating #1117

corranwebster:
These are some drive-by fixes from issues found when investigating #1117.  None of these actually fix the issue, however.

The changes to the state-saving code don't impact the file format, as no changes are needed for the reader (it can happily ingest both `QByteArray` and `bytes` objects).


mdickinson (approved):
LGTM

PR #1154, opened 2022-07-18
Update documentation for Pyface 7.4.2

corranwebster:
Does what it says.  As with previous bugfix releases this overwrites the documentation for the 7.4.1 release.

This includes the Sphinx copy button functionality, so this is a little more than a straight replacement of the old files with the new.


mdickinson (approved):
LGTM. We're still running into https://github.com/enthought/enthought-sphinx-theme/issues/19, but that doesn't appear to affect the HTML rendering in any detectable way. The copy button works as expected on my system (macOS / Chrome).

PR #1161, opened 2022-10-10
BUG: Safe hasattr for PyQt5

larsoner:
Over in https://github.com/enthought/mayavi/pull/1180 I am working on VTK 9.2.2 fixes, and I am getting a segfault in CIs during GC:

https://github.com/enthought/mayavi/actions/runs/3222141217/jobs/5270891488

Locally I don't get a segfault but rather a slightly nicer error message:

```
Exceptions caught in Qt event loop:
________________________________________________________________________________
  File "/home/larsoner/python/pyface/pyface/ui/qt4/action/action_item.py", line 379, in _qt4_on_destroyed
    if hasattr(self.control, "_tool_instance"):
RuntimeError: wrapped C/C++ object of type QAction has been deleted
```

This PR fixes this by catching this `RuntimeError` during `hasattr`. Locally at least the Mayavi tests pass with PyQt5 with this change.

### DIFF
--- a/pyface/ui/qt4/action/action_item.py
+++ b/pyface/ui/qt4/action/action_item.py
@@ -376,7 +376,12 @@ class _Tool(HasTraits):
         """
         if self.control is not None:
             # Remove the cycle since we're no longer needed.
-            if hasattr(self.control, "_tool_instance"):
+            try:
+                has_instance = hasattr(self.control, "_tool_instance")
+            # PyQt5 "wrapped C/C++ object ... has been deleted"
+            except RuntimeError:
+                has_instance = False
+            if has_instance:
                 del self.control._tool_instance
         self.control = None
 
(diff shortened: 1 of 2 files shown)


corranwebster:
Thanks - this looks like a reasonable fix.

pyface/ui/qt4/action/action_item.py
```diff
@@ -376,7 +376,12 @@ def _qt4_on_destroyed(self, control=None):
         """
         if self.control is not None:
             # Remove the cycle since we're no longer needed.
-            if hasattr(self.control, "_tool_instance"):
+            try:
+                has_instance = hasattr(self.control, "_tool_instance")
+            # PyQt5 "wrapped C/C++ object ... has been deleted"
```
corranwebster, line 381:
```suggestion
            # fixes mayavi #1180: PyQt5 "wrapped C/C++ object ... has been deleted"
```

larsoner:
There was a timeout on one run after 360 minutes (!) that's hopefully unrelated, so rather than pushing an empty limit I added a `timeout-minutes: 20` to the CI run. Looks like all others were < 5 minutes so this seemed like a reasonable buffer against slow downloads and such.

corranwebster:
Yes, it is almost certainly orthogonal - I haven't seen that sort of error in the CI runs, but I will keep an eye out for it in the future.

I think this should be good to merge.

corranwebster (approved):

PR #1165, opened 2022-10-23
`QtGui.QSizePolicy.Policy.*`

ntntnlstdnt:
The Fixed, Preferred, Expanding attributes now belong to `QtGui.QSizePolicy`


ntntnlstdnt:
There more files that  yield similar errors. Namely places where attributes from `QtCore` and `QtGui` modules are used.

corranwebster:
Thanks for the contribution and apologies for the delay in responding!

I might need to do a manual check to see if this causes a problem with PyQt6 in addition to the other supported Qt wrappers.

corranwebster:
I think that the current `main` branch needs to be merged in to this PR branch before tests will pass.

PR #1166, opened 2022-10-23
Attribute changes

ntntnlstdnt:
1. `Accepted` and `Rejected` are now attributes of `QtGui.Dialog`
2. Similarly, Ok, Cancel, Yes, No are now attributes of `QtGui.QMessageBox`


corranwebster:
Similarly here: I think that the current `main` branch needs to be merged in to this PR branch before tests will pass.

PR #1186, opened 2022-12-09
CI changes for Ubuntu 22.04

corranwebster:
This PR fixes CI for Ubuntu 22.04 so that it is back into a working state.  To this end it:

- updates the Qt debian packages
- removes the use of `GabrielBB/xvfb-action@v1` (`xvfb` is in the default 22.04 image)
- removes linux/wx tests as we don't have a compatible wxpython wheel/egg for ubuntu 22.04/wxPython 4.0


corranwebster:
Thanks!

PR #1189, opened 2022-12-19
Release 7.4.4

corranwebster:
Trying something a little different here - combining the changelog, release flag and version bump into one PR, combined with tagging the release flag commit.

We will see how this works


corranwebster:
Note for the future: better to push the version bump later so that the actual release gets CI run on it.

Issue #614, opened 2020-07-20
Get wx version of GuiTestAssistant code


corranwebster:
This PR needs to be completed https://github.com/enthought/pyface/pull/266.

PR #1192, opened 2023-01-10
Remove 'parent' argument from wx StatusBarManager

greschd:
Another small fix:

Remove the 'parent' argument from 'StatusBarManager.destroy' in the wx backend. The argument is unused, and doesn't match the interface definition.

If you prefer keeping the argument (to not break applications which may pass it explicitly?), I'd propose adding a default `=None`, and emitting a deprecation warning.

### DIFF
--- a/pyface/ui/wx/action/status_bar_manager.py
+++ b/pyface/ui/wx/action/status_bar_manager.py
@@ -50,7 +50,7 @@ class StatusBarManager(HasTraits):
 
         return self.status_bar
 
-    def destroy(self, parent):
+    def destroy(self):
         """ Removes a status bar. """
 
         if self.status_bar is not None:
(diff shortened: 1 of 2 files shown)


corranwebster (changes requested):
Once again, thanks for the PR.  This probably needs a corresponding change to the call here: https://github.com/enthought/pyface/blob/acccc5ea314f006d8734a51bfb0074c8906beefa/pyface/ui/wx/application_window.py#L225

corranwebster:
> I wonder if we should add some kind of test?

There are tests for the create/destroy aspects of the `StatusBarManager` API that would have detected the problem here: https://github.com/enthought/pyface/blob/9aeb3bd56bb6bec52b244132a928416e40d4b878/pyface/tests/test_application_window.py#L222-L252

Unfortunately those tests need the `GuiTestAssistant` class which we haven't yet integrated for WxPython (see #614 and #266 - the primary obstacle is not writing the class, it's getting everything running correctly once you have it).

So the long and short of it is, no, no need to write a test, and this looks good to go.

corranwebster (approved):

greschd:
Thanks for the review, and the context on testing!

PR #1226, opened 2023-03-24
Don't try to dispatch immediately in Qt Gui `_FutureCall`

corranwebster:
This has gone back and forth a few times, but right now unhandled exceptions in the `event` handler are causing test failures in TraitsUI on 6.4.3 and this fixes them.


mdickinson (approved):
LGTM; if this fixes the issues, I'm all for it.

PR #1227, opened 2023-03-27
Remove general uses of "qt4" in favour of "qt".  Update docs.

corranwebster:
This changes names of some private methods which should be fine. In some cases public methods/functions are changed in which case we use keep backwards-compatible aliases around.


corranwebster:
Not a clean PR (pulling in things from other work I'd been doing)

PR #1229, opened 2023-03-29
Expose more ui traits in Pyface API

### LINKED ISSUE #1071: Add `PyfaceColor`, `PyfaceFont` traits to pyface.api
Possibly also want `Color` and `Font` classes added there as well.  Not sure we want that quite yet, as there isn't much that external users can do with these objects.

corranwebster:
This adds a few trait types from pyface.ui_traits which were not in the public API.  It also makes the Font and Color classes visible, for consistency.

Fixes #1071

### DIFF
--- a/pyface/api.py
+++ b/pyface/api.py
@@ -87,23 +87,27 @@ UI Traits
 ---------
 
 - :attr:`~.Alignment`
-- :class:`~.Border`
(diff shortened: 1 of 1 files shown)


mdickinson (approved):
LGTM

PR #1240, opened 2023-04-06
Import optional imports from relative module rather than toolkit

corranwebster:
This fixes an issue seen in TraitsUI bleeding-edge tests.


mdickinson (approved):
LGTM

PR #1260, opened 2025-02-09
Fix application toolbars

bpteague:
This PR fixes https://github.com/enthought/pyface/issues/1251, which keeps toolbars from being created due to a copy-and-paste error.


johannesloibl (approved):

mdickinson (approved):
LGTM

